PR #39, opened 2014-01-31
[WiP] - rely on GitStatus.notifyCommit to trigger a build

ndeloof:
github post commit hook [payload](https://help.github.com/articles/post-receive-hooks) contains enough information to trigger a build using git plugin GitStatus.notifyCommit

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
+++ b/src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
@@ -9,6 +9,7 @@ import hudson.model.Hudson.MasterComputer;
 import hudson.model.Item;
 import hudson.model.AbstractProject;
 import hudson.model.Project;
+import hudson.plugins.git.GitStatus;
 import hudson.triggers.Trigger;
 import hudson.triggers.TriggerDescriptor;
 import hudson.util.SequentialExecutionQueue;
@@ -30,9 +31,11 @@ import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
+import jenkins.model.Jenkins;
 import net.sf.json.JSONObject;
 
 import org.apache.commons.jelly.XMLOutput;
+import org.eclipse.jgit.transport.URIish;
 import org.kohsuke.github.GHException;
 import org.kohsuke.github.GHRepository;
 import org.kohsuke.stapler.DataBoundConstructor;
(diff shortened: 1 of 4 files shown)


src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -60,58 +63,14 @@ public void onPost() {
...
-                    if (job.scheduleBuild(cause)) {
-                        LOGGER.info("SCM changes detected in "+ job.getName()+". Triggering "+name);
-                    } else {
-                        LOGGER.info("SCM changes detected in "+ job.getName()+". Job is already in the queue");
-                    }
-                }
-            }
-        });
+    public void onPost(String triggeredByUser, URIish repository, String sha1, String ref) {
+        String branch = ref.substring("refs/heads/".length());
```
KostyaSha: As i understand this can prevent from tracking any `refs/*` and not `refs/head` (that is just default of "--heads" and branches). Why do you need hardcode `refs/heads` at all?

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -60,58 +63,14 @@ public void onPost() {
...
-                        throw e;
-                    } finally {
-                        listener.close();
-                    }
-                } catch (IOException e) {
-                    LOGGER.log(Level.SEVERE,"Failed to record SCM polling",e);
-                }
-                return false;
-            }
+        throw new UnsupportedOperationException("deprecated");
```
oleg-nenashev: IMO should be logged to Jenkins log

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -60,58 +63,14 @@ public void onPost() {
      * Called when a POST is made.
      */
     public void onPost(String triggeredByUser) {
```
oleg-nenashev: Mark as deprecated + appropriate Javadoc would be useful

KostyaSha:
Probably blocked now because of https://issues.jenkins-ci.org/browse/JENKINS-28506 and some other PRs.

smazurov:
Would this solve https://issues.jenkins-ci.org/browse/JENKINS-12545?

KostyaSha:
@smazurov i already answered in this issue

Yes it was blocked with different things.

jglick:
This change would make sense to me. I was surprised to find that the plugin did not already use SCM polling in the normal way.

KostyaSha:
@jglick in plan after credentials, if you want rework it or see any issues, please inform.

lanwen:
Relying to notifyCommit is wrong way to resolve polling issues. Or this plugin bacame to just a proxy from hooks to SCMTrigger. (Which breaks current behaviour)

More information: https://issues.jenkins-ci.org/browse/JENKINS-27041

PR #44, opened 2014-07-21
Filter git SCM branch in github push

aserrallerios:
[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/44)

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/GitHubBranch.java
+++ b/src/main/java/com/cloudbees/jenkins/GitHubBranch.java
@@ -1,0 +1,62 @@
+package com.cloudbees.jenkins;
+
+import hudson.plugins.git.BranchSpec;
+
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * Identifies a GitHub branch and hides the SCM branch from its clients.
+ *
+ * @author aserralle
+ *
+ */
+public class GitHubBranch {
+
+    private static final String GITHUB_BRANCH_PREFIX = "refs/heads";
+    private static final int GITHUB_BRANCH_PREFIX_LENGTH = 10;
+
+    private final BranchSpec branch;
+
+    private GitHubBranch(final BranchSpec branch) {
+        this.branch = branch;
+    }
+
+    public boolean matches(GitHubRepositoryName repository, String ref) {
+        // ref (github) -> refs/heads/master
+        // branch (git SCM) -> REPO/remote/branch
+        // matching the meaningful part of the github branch name with
+        // the configured SCM branch expression
+        if (ref != null && !ref.equals("")) {
+            String branchExpression = "*";
+            if (repository != null && repository.repositoryName != null) {
+                branchExpression = repository.repositoryName;
+            }
+            if (ref.startsWith(GITHUB_BRANCH_PREFIX)) {
+                branchExpression += ref.substring(10);
+            } else {
+                branchExpression += "/" + ref;
+            }
+            LOGGER.log(Level.FINE, "Does SCM branch " + branch.getName()
+                    + " match GitHub branch " + branchExpression + "?");
+            return branch.matches(branchExpression);
+        }
+        return false;
+    }
+
+    public static GitHubBranch create(BranchSpec branch) {
+        if (isValidGitHubBranch(branch)) {
+            return new GitHubBranch(branch);
+        } else {
+            return null;
+        }
+    }
+
+    // TODO: implement logic to validate git SCM branches.
+    private static boolean isValidGitHubBranch(BranchSpec branch) {
+        return true;
+    }
+
+    private static final Logger LOGGER = Logger.getLogger(GitHubBranch.class
+            .getName());
+}
--- a/src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
+++ b/src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
@@ -7,9 +7,11 @@ import hudson.ExtensionPoint;
 import hudson.model.AbstractProject;
 import hudson.model.EnvironmentContributor;
 import hudson.model.TaskListener;
+import hudson.plugins.git.BranchSpec;
 import hudson.plugins.git.GitSCM;
 import hudson.scm.SCM;
 import jenkins.model.Jenkins;
+
 import org.eclipse.jgit.transport.RemoteConfig;
 import org.eclipse.jgit.transport.URIish;
 import org.jenkinsci.plugins.multiplescms.MultiSCM;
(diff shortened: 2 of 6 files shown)


aserrallerios:
Rebased, and decoupled some concepts. Added some tests for the GitHub - SMC branch matching.

I tried to add some tests for the `GitHubWebHook` code but I was unable to mock everything up.

jstriebel:
+1 really missing this feature at the moment!

freeyoung:
+1 for this feature

ema:
+1 please

tavisto:
+1 please

src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
```diff
@@ -7,9 +7,11 @@
 import hudson.model.AbstractProject;
 import hudson.model.EnvironmentContributor;
 import hudson.model.TaskListener;
+import hudson.plugins.git.BranchSpec;
 import hudson.plugins.git.GitSCM;
 import hudson.scm.SCM;
 import jenkins.model.Jenkins;
+
```
KostyaSha: exclude new line

src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
```diff
@@ -32,6 +34,12 @@
      */
     public abstract void parseAssociatedNames(AbstractProject<?,?> job, Collection<GitHubRepositoryName> result);
 
+    /**
+     * Looks at the definition of {@link AbstractProject} and list up its branches, then puts them into
+     * the collection.
+     */
+    public abstract void parseAssociatedBranches(AbstractProject<?,?> job, Collection<GitHubBranch> result);
```
KostyaSha: Will this break backward compatibility?

yani-:
+1 please

src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
oleg-nenashev: Yes, it will. A default implementation is mandatory

KostyaSha: ` public abstract class GitHubRepositoryNameContributor implements ExtensionPoint `
Breaking compatibility for ExtensionPoint is bad idea.

src/main/java/com/cloudbees/jenkins/GitHubBranch.java
```diff
@@ -0,0 +1,62 @@
...
+        }
+    }
+
+    // TODO: implement logic to validate git SCM branches.
+    private static boolean isValidGitHubBranch(BranchSpec branch) {
+        return true;
+    }
+
+    private static final Logger LOGGER = Logger.getLogger(GitHubBranch.class
+            .getName());
```
KostyaSha: Are you coding from mobile? Why so short width?

npaufler:
Am I understanding this PR correctly in that it will filter the branch out of the github webhook push request, so that only the branch that has received a push will be poked? We have a serious issue right now whereby we effectively DoS ourselves due to the number of branches we have and a webhook push causes -every- branch to get considered and poked.

If so, any idea when this will get merged in? 

I did try building it locally (installed oracle JDK7, maven, and did a mvn hpi:hpi per the readme) but the resulting plugin .hpi file did not seem to work - Jenkins did not recognize it correctly. All github references were removed from the config options, and Jenkins complained about incorrect data in the config files.

I think it's more likely that I made a mistake in building it than a problem with the PR, though. Happy to try it if I can get a pointer on how to properly build it.

oleg-nenashev:
@npaufler 
The change is valid and useful, but the implementation needs an improvement: binary backward compatibility, usage of classes from *github-api* library. Somebody just needs to invest some time and to rework the pull request.

src/main/java/com/cloudbees/jenkins/GitHubBranch.java
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Identifies a GitHub branch and hides the SCM branch from its clients.
+ *
+ * @author aserralle
+ *
+ */
+public class GitHubBranch {
+
+    private static final String GITHUB_BRANCH_PREFIX = "refs/heads";
```
KostyaSha: Why it hardcoded? What if i want trigger refs/tags, refs/notes?

KostyaSha:
Is https://issues.jenkins-ci.org/browse/JENKINS-24291 related?

KostyaSha:
@npaufler 
> We have a serious issue right now whereby we effectively DoS ourselves due to the number of branches we have and a webhook push causes -every- branch to get considered and poked.

What do you mean under "-every- branch to get considered and poked"?
I have some ideas what can be improved for performance but firstly i want understand other use cases

In general this PR looks not well formed, @aserrallerios proposed PR without referring to any jira issue or any description.

kohsuke:
It's quite possible that I'm not understanding this, but if you are interested in scheduling builds only for specific branches, the right thing to do is to set the branch spec in the Git SCM configuration.

All this plugin is doing is to schedule a polling, and it's up to the Git plugin to determine if the change pushed is worth the build.

There are lots of criteria when it comes to what is considered an interesting change and what is not. We can't add every one of them in this plugin. We've learned that lesson in the Subversion plugin and the Git v1 plugin.

KostyaSha:
@kohsuke probably penalty is in heavy weight polling check that is called in SCM internals. Of course github plugin can't determine that some job configuration meets build criteria (for example excluded messages and other behaviours), but it can do light filtering, for example by filtering not matched branch specifier configuration. (just a guess)

KostyaSha:
Btw, @aserrallerios @npaufler could you test this PR https://github.com/jenkinsci/github-plugin/pull/39 ? It changes the way how plugin triggers builds it also deal with branches. I think #39 will be merged in any case because it switches to better API usage.

llasram:
I'm interested in the feature this PR provides, but it seems fairly complex and inflexible vs the approach of providing values from the push event payload as build parameters, as discussed in [JENKINS-24291](https://issues.jenkins-ci.org/browse/JENKINS-24291).

I hacked together enough a prototype to verify that exposing such parameters is both doable and enables the results that I'm trying to achieve.  If I cleaned the implementation and opened a PR, would such a change be accepted?

KostyaSha:
@llasram i doubt. Atm trigger do the simple thing - it calls GitSCM poll that then checks whether build should be scheduled. Gitscm contains branch specifier and other features that "filter" == choses branch to build.

KostyaSha:
Closing as initial author didn't respond to any questions and code is not mergeable anymore for a long time.

talcloudshare:
Is there any intention on making this work for upcoming versions? Seems like a really useful functionality

hachi:
Can we please get a working version of this? There are several tickets that have been open since 2014 that apply to this concept.

Passing data from the webhook as environment variables would be minimally useful.
Filtering branches or commits automatically by a checkbox would be the ultimate.

Either way, having a new project that builds all branches because it has never been built before is pretty useless. I just hit this at work and had a 20minute build job that fired off for some 30 branches because of the number of engineers working... and a brand new jenkins installation.

KostyaSha:
YOU CAN'T DO IT BECAUSE PUSH TRIGGER IS ONLY KICK POLLING WITH SEPARATE LOGIC. 
That written in documentation. If you want achieve what you want, you need use other triggers.

PR #45, opened 2014-09-15
[WiP] - JENKINS-24702

ndeloof:
convert legacy credentials stored as plain text (including Oauth token :-\) to use credentials-plugin.
Maybe AccessTokenCredential is generic enough for this to be defined by some oauth-credentials plugin, not sure how portable this is and if this would make sense.

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/github/AccessTokenCredential.java
+++ b/src/main/java/com/cloudbees/jenkins/github/AccessTokenCredential.java
@@ -90,5 +90,20 @@ public class AccessTokenCredential extends BaseStandardCredentials implements St
             }
             return candidates;
         }
+
+
+        public FormValidation doValidate(@QueryParameter String apiUrl, @QueryParameter String username, @QueryParameter String oauthAccessToken) throws IOException {
+            GitHub gitHub;
+            if (Util.fixEmpty(apiUrl) != null) {
+                gitHub = GitHub.connectToEnterprise(apiUrl,oauthAccessToken);
+            } else {
+                gitHub = GitHub.connect(username,oauthAccessToken);
+            }
+
+            if (gitHub.isCredentialValid())
+                return FormValidation.ok("Verified");
+            else
+                return FormValidation.error("Failed to validate the account");
+        }
     }
 }
--- a/src/main/java/com/cloudbees/jenkins/github/AccessTokenCredential.java
+++ b/src/main/java/com/cloudbees/jenkins/github/AccessTokenCredential.java
@@ -1,0 +1,94 @@
+package com.cloudbees.jenkins.github;
+
+import com.cloudbees.plugins.credentials.CredentialsDescriptor;
+import com.cloudbees.plugins.credentials.CredentialsScope;
+import com.cloudbees.plugins.credentials.common.StandardUsernameCredentials;
+import com.cloudbees.plugins.credentials.impl.BaseStandardCredentials;
+import hudson.Extension;
+import hudson.Util;
+import hudson.model.AutoCompletionCandidates;
+import hudson.util.FormValidation;
+import hudson.util.Secret;
+import jenkins.model.Jenkins;
+import org.kohsuke.github.GHAuthorization;
+import org.kohsuke.github.GitHub;
+import org.kohsuke.stapler.DataBoundConstructor;
+import org.kohsuke.stapler.QueryParameter;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * @author <a href="mailto:nicolas.deloof@gmail.com">Nicolas De Loof</a>
+ */
+public class AccessTokenCredential extends BaseStandardCredentials implements StandardUsernameCredentials {
+
+    private final String githubServer;
+
+    private final String username;
+
+    private final Secret token;
+
+    @DataBoundConstructor
+    public AccessTokenCredential(CredentialsScope scope,
+                                 String id, String description,
+                                 String githubServer, String username, String token) {
+        super(scope, id, description);
+        this.githubServer = githubServer;
+        this.username = Util.fixNull(username);
+        this.token = Secret.fromString(token);
+    }
+
+    public String getUsername() {
+        return username;
+    }
+
+    public Secret getToken() {
+        return token;
+    }
+
+    @Extension
+    public static class DescriptorImpl extends CredentialsDescriptor {
+
+        @Override
+        public String getDisplayName() {
+            return "Github Access Token";
+        }
+
+
+        public FormValidation doCreateToken(@QueryParameter("githubServer") String githubServer,
+                                            @QueryParameter("username") String username,
+                                            @QueryParameter("password") String password,
+                                            @QueryParameter("scope") String scopes) {
+
+
+            String[] sc = scopes.split("\\s*,[,\\s]*"); // GHAuthorization.REPO_STATUS, GHAuthorization.REPO
+
+            try {
+                GitHub gh = GitHub.connectToEnterprise(githubServer, username, password);
+                GHAuthorization token = gh.createToken(Arrays.asList(sc),
+                        "Jenkins GitHub Plugin", Jenkins.getInstance().getRootUrl());
+                return FormValidation.ok("Access token created: " + token.getToken());
+            } catch (IOException ex) {
+                return FormValidation.error("GitHub API token couldn't be created: " + ex.getMessage());
+            }
+        }
+
+        // TODO build from GHAuthorization.*, but require https://github.com/kohsuke/github-api/pull/128
+        final List<String> scopes = Arrays.asList("user", "user:email", "user:follow",
+                "public_repo", "repo", "repo_deployment", "repo:status", "delete_repo",
+                "notifications", "gist",
+                "read:repo_hook", "write:repo_hook", "admin:repo_hook",
+                "read:org", "write:org", "admin:org",
+                "read:public_key", "write:public_key", "admin :public_key");
+
+        public AutoCompletionCandidates doAutoCompleteScope(@QueryParameter String value) {
+            AutoCompletionCandidates candidates = new AutoCompletionCandidates();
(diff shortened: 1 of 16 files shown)


src/main/java/com/cloudbees/jenkins/github/AccessTokenCredential.java
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * @author <a href="mailto:nicolas.deloof@gmail.com">Nicolas De Loof</a>
+ */
+public class AccessTokenCredential extends BaseStandardCredentials implements StandardUsernameCredentials {
+
+    private final String githubServer;
+
+    private final String username;
+
+    private final Secret token;
```
KostyaSha: Does credentials plugin has some base class for tokens, so github-plugin would extend it only with it's unique internals?
What is really needed for storing tokens? Isn't enough for plugin to store server url separately and add token (but with credentials plugin) to configuration?

KostyaSha:
Status?

ndeloof:
I'm stuck here. I'd like the github server URL NOT to be part of the token credential type, but then can't offer a "generate" UX

KostyaSha:
@ndeloof should we call @stephenc for help? Credentials code looks complex and as i see almost all plugins not implementing credentials because.
Is there some design description of this credentials plugin? I don't understand why every plugin should implement so basic Credential types.

stephenc:
On Saturday, April 11, 2015, Kanstantsin Shautsou <notifications@github.com>
wrote:

> @ndeloof <https://github.com/ndeloof> should we call @stephenc
> <https://github.com/stephenc> for help? Credentials code looks complex
> and as i see almost all plugins not implementing credentials because.
> Is there some design description of this credentials plugin? I don't
> understand why every plugin should implement so basic Credential types.
>

My rule of thumb. If the backing credential information could conceivably
be synced with another consumer of credentials, then use a common type.

The first test case is username/password. People like ldap backed sso. So
do not create a credential type to hold username&password because that
causes at best interface envy and at worst class envy. There is converter
support to help roll back if you shipped such a mistake

The second case is ssh key because you likely will use the same key against
multiple services.

I'm on my phone so cannot see what is going on here but I am strongly
against every plugin and their mother implementing credential types:

* confuses users as to which type they need to create
* creates type envy causing inter-locking plugin dependencies
* really just a hack around using domain specs/reqs

—
> Reply to this email directly or view it on GitHub
> <https://github.com/jenkinsci/github-plugin/pull/45#issuecomment-91880053>
> .
>


-- 
Sent from my phone

KostyaSha:
@stephenc thanks for your comment, i will be glad to read some examplse/documentation of right credentials usage/implementation (or maybe i missed existed). 
1) In case of github plugin credentials: a) username+password, b) token, c) (probably) token + secure field(used for decrypting incoming hooks data) - can be reused by other GH plugins (for example i created new github-pullrequest-plugin and will be glad to fix/implement in right place credentials). 
2) Also i not really understand how URL should be matched/related/bundled in credentials. From one side GitHub related passwords are just GH related, but on other hand they can used for git https checkout communication.

stephenc:
J

On Saturday, April 11, 2015, Kanstantsin Shautsou <notifications@github.com>
wrote:

> @stephenc <https://github.com/stephenc> thanks for your comment, i will
> be glad to read some examplse/documentation of right credentials
> usage/implementation (or maybe i missed existed).
>
>
> Sadly I am kept too busy on enterprise features to write up the docs I
have wanted to. Most of this I have covered with some people implementing
OAuth credentials. I cannot recall whether that was on/off list or over IRC

I think I need to document it out as a number of people are getting
confused all at the same time.

>
> 1) In case of github plugin credentials: a) username+password,
>
> This should be a standard user-pass. No special type. If somebody has
github enterprise their username/password might be synced to corp ldap, so
we want them to only have one credential to update every 90 days, not send
them running to update 50 different credentials before jenkins pushes three
failed logins and locks their account out

>  b) token,
>
> So here the case is how is the token generated. This is presumably
generated by github or github enterprise. It would this make sense to store
the "issuer" as a field and provide a button that initiates a web-flow
against the issuer to generate the token. In that case "issuer" would need
to be a global config so we can give the user a drop down if issuers to
trigger the web flow against.

This token might well implement some other interfaces or you may use the
auth-tokens-api to convert it for use.

The point is that the web flow needs are specific to the class of issuer,
so you would have a gh/GHE specific credential type for this

>  c) (probably) token + secure field(used for decrypting incoming hooks
> data) - can be reused by other GH plugins (for example i created new
> github-pullrequest-plugin and will be glad to fix/implement in right place
> credentials).
>
>
> A similar analysis would apply here.

>
> 2) Also i not really understand how URL should be matched/related/bundled
> in credentials. From one side GitHub related passwords are just GH related,
> but on other hand they can used for git https checkout communication.
>
When you are asking the user to select credentials you have a target that
the credentials are for, you use that to generate the domain requirements.
Typically you would use the git url, but you can add additional
requirements (of custom types) as you see fit.

The user creates a credential domain and attached domain specifications to
that domain. They then put the credentials for that domain into the domain
and presto those credentials show up.

Example:

Create a domain with the spec: *.acme.com
Put user/pass from ldap.acme.com into domain.
When asking for credentials to check out from git.acme.com that user pass
will be available.






> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jenkinsci/github-plugin/pull/45#issuecomment-91882871>
> .
>


-- 
Sent from my phone

src/main/java/com/cloudbees/jenkins/github/AccessTokenCredential.java
stephenc: @ndeloof the token is really a password with restricted permissions. The user gets to select how much permissions to give the token. To my way of thinking a better way would be to have domain specifications and domain requirements that match the permissions that can be granted to a github token. That way users can put the tokens into the domains that match the permissions they granted.

A username/password would have full permissions, while a username/token would be in a different domain. As such I am unclear that the token needs to be a separate type from the username/password. Things that would argue for a separate type:

* The auth with a token being different than the auth with a password, e.g. when using a password you have to pass `Authenication: Basic ...` but with a token you have to pass `Authentication: Token ...`
* You are providing a special API to generate the token (e.g. via a web flow)... if the resulting token behaves just like a password in terms of authentication then you would implement the `StandardUsernamePasswordsCredentials` interface but use a button or such to control generating/validation/updating the token and you would not display a "password" field.
* Perhaps some other exotic use case I cannot think of right now.

Absence the above, I would favour not creating a special type for this.

In any case I think you need to look at creating some `GitHubPermissionDomainRequirement` and `GitHubPermissionDomainSpecification` style class(es) so that users can put tokens into domains that match what the tokens are good for.

src/main/java/com/cloudbees/jenkins/github/AccessTokenCredential.java
```diff
@@ -0,0 +1,109 @@
...
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * @author <a href="mailto:nicolas.deloof@gmail.com">Nicolas De Loof</a>
+ */
+public class AccessTokenCredential extends BaseStandardCredentials implements StandardUsernameCredentials {
+
+    private final String githubServer;
```
stephenc: If you are providing a specialized UI to provide for generating/revalidating/regenerating the token, then yes the server that you operate against (i.e. "the origin of authentication truth") is part of the credential. This does not mean that the credential cannot be used against other services, e.g. they may be pass-through services or they may be services that delegate auth to a github enterprise instance.

src/main/java/com/cloudbees/jenkins/github/AccessTokenCredential.java
ndeloof: @stphenc my intent here was to offer option to create a token from jenkins UI, using a classic username/password authentication to github token generation API. The token then for sure can be considered being some username/password item, but the related UI is a distinct one.

stephenc: So you want to provide a different UI flow, but my understanding is that
the resulting token is used just like a password. So you probably want to
implement `StandardUsernamePasswordCredentials` *because* the token is used
to authenticate *as if it were a password*.

Because you are providing an alternative UI flow it makes sense to be a
separate credential type, but the actual stored credentials (to my
understanding, correct me if I am wrong) is a username and a *"password"*
(much like the app specific passwords that GMail lets you generate with 2FA
enabled... The app you give that "password" to does not know it is a token
and not your actual password)

On Sunday, April 12, 2015, Nicolas De loof <notifications@github.com> wrote:

> In src/main/java/com/cloudbees/jenkins/github/AccessTokenCredential.java
> <https://github.com/jenkinsci/github-plugin/pull/45#discussion_r28201391>:
>
> > +import org.kohsuke.stapler.QueryParameter;
> > +
> > +import java.io.IOException;
> > +import java.util.Arrays;
> > +import java.util.List;
> > +
> > +/**
> > + * @author <a href="mailto:nicolas.deloof@gmail.com <javascript:_e(%7B%7D,'cvml','nicolas.deloof@gmail.com');>">Nicolas De Loof</a>
> > + */
> > +public class AccessTokenCredential extends BaseStandardCredentials implements StandardUsernameCredentials {
> > +
> > +    private final String githubServer;
> > +
> > +    private final String username;
> > +
> > +    private final Secret token;
>
> @stphenc my intent here was to offer option to create a token from jenkins
> UI, using a classic username/password authentication to github token
> generation API. The token then for sure can be considered being some
> username/password item, but the related UI is a distinct one.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jenkinsci/github-plugin/pull/45/files#r28201391>.
>


-- 
Sent from my phone

stephenc:
[I decided to finally write up some of my thinking around the Credentials API](http://developer-blog.cloudbees.com/2015/04/the-credentials-api-in-jenkins.html)

Hopefully this helps thinking around this pull request

KostyaSha:
Closing via #65

PR #47, opened 2014-11-04
Use the combined status API

esteve:
This branch adds the "jenkins" context to statuses so other tools can update the status of a pull request in separate contexts.

More info:
https://developer.github.com/v3/repos/statuses/#get-the-combined-status-for-a-specific-ref

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/GitHubCommitNotifier.java
+++ b/src/main/java/com/cloudbees/jenkins/GitHubCommitNotifier.java
@@ -43,33 +43,22 @@ import org.jenkinsci.plugins.github.util.BuildDataHelper;
 public class GitHubCommitNotifier extends Notifier {
 
     private final String resultOnFailure;
-    private final String statusReporterName;
     private static final Result[] SUPPORTED_RESULTS = {FAILURE, UNSTABLE, SUCCESS};
-    private static final String DEFAULT_STATUS_REPORTER_NAME = "jenkins";
-
+    
     @DataBoundConstructor
-    public GitHubCommitNotifier(String resultOnFailure, String statusReporterName) {
-        this.resultOnFailure = resultOnFailure;
-        this.statusReporterName = statusReporterName;
-    }
-
     public GitHubCommitNotifier(String resultOnFailure) {
-        this(resultOnFailure, DEFAULT_STATUS_REPORTER_NAME);
+        this.resultOnFailure = resultOnFailure;
     }
-
+    
     @Deprecated
     public GitHubCommitNotifier() {
-        this(getDefaultResultOnFailure().toString(), DEFAULT_STATUS_REPORTER_NAME);
+        this(getDefaultResultOnFailure().toString());
     }
 
     public @Nonnull String getResultOnFailure() {
         return resultOnFailure != null ? resultOnFailure : getDefaultResultOnFailure().toString();
     }
-
-    public @Nonnull String getStatusReporterName() {
-        return statusReporterName != null ? statusReporterName : DEFAULT_STATUS_REPORTER_NAME;
-    }
-
+     
     public static @Nonnull Result getDefaultResultOnFailure() {
         return SUPPORTED_RESULTS[0];
     }
@@ -108,9 +97,9 @@ public class GitHubCommitNotifier extends Notifier {
         }
         return true;
     }
-
-    private void updateCommitStatus(@Nonnull AbstractBuild<?, ?> build, @Nonnull BuildListener listener) throws InterruptedException, IOException {
-        final String sha1 = ObjectId.toString(BuildDataHelper.getCommitSHA1(build));
+        
+    private void updateCommitStatus(@Nonnull AbstractBuild<?, ?> build, @Nonnull BuildListener listener) throws InterruptedException, IOException {       
+        final String sha1 = ObjectId.toString(BuildDataHelper.getCommitSHA1(build));  
         for (GitHubRepositoryName name : GitHubRepositoryNameContributor.parseAssociatedNames(build.getProject())) {
             for (GHRepository repository : name.resolve()) {
                 GHCommitState state;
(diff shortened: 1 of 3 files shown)


dirk-thomas:
@esteve Can you please update the PR to use the job name as the reporter (instead of manually entering it).

nicklamer:
@dirk-thomas Correct me if I am wrong, but reporter name is optional parameter. When you'll leave it blank it will use DEFAULT_STATUS_REPORTER_NAME which is "jenkins.
https://github.com/jenkinsci/github-plugin/pull/47/files#diff-752cdecf02b12bb22c1fe3285f4d99a5R69

I agree that in most cases it would be more specific to use job name instead of "jenkins" as default value.
For example, in our project we build few jenkins job per commit and would like to attach all of them to commit as different context (reportername).

On the other hand I cannot imagine situation, when we would like to overwrite commit status from multiple jenkins jobs per commit (situation with "jenkins" default value).


@esteve Great stuff I thought about the same improvement in plugin. Hope it will be merged and released quickly :)

KostyaSha:
>On the other hand I cannot imagine situation, when we would like to overwrite commit status from >multiple jenkins jobs per commit (situation with "jenkins" default value).

Different test jobs (short/long time) for the same commit

nicklamer:
>>On the other hand I cannot imagine situation, when we would like to overwrite commit status from >>multiple jenkins jobs per commit (situation with "jenkins" default value).
>Different test jobs (short/long time) for the same commit

Could you be more specific? Do you mean that you would like to have only one status for two jobs?
Is about parallel run the same job on different machines (fast/slow)?
Anyway, It will be still possbile when you override parameter with "jenkins" and allow your jobs to override commit status.

(You could edit your previous response, then I will remove my)

KostyaSha:
Any new UI entry/option must make sense. For this case imho it's enough just to use job name (probably getFullName() to exclude folder collisions, because folder can be copied), so just set commit status with context = job name. Fix/improvement is just one line change.

src/main/java/com/cloudbees/jenkins/GitHubCommitNotifier.java
```diff
@@ -129,7 +129,7 @@ private void updateCommitStatus(@Nonnull AbstractBuild<?, ?> build, @Nonnull Bui
                 }
 
                 listener.getLogger().println(Messages.GitHubCommitNotifier_SettingCommitStatus(repository.getUrl() + "/commit/" + sha1));
-                repository.createCommitStatus(ObjectId.toString(sha1), state, build.getAbsoluteUrl(), msg);
+                repository.createCommitStatus(ObjectId.toString(sha1), state, build.getAbsoluteUrl(), msg, "jenkins");
```
KostyaSha: @oleg-nenashev do you really want to have another one configuration in UI? Correct job name will have correct commit status.

esteve:
I just reverted the Jelly changes and used the build's full name instead.

KostyaSha:
Also try test it with folders plugin.

esteve:
@KostyaSha I'm not using that plugin, could you test that this change works with the folders plugin? Thanks.

KostyaSha:
Just install it and put job in folder. I have no time for testing your PR.

dirk-thomas:
The new API was introduced in the `github-api` plugin in version 1.55 (https://github.com/kohsuke/github-api/commits/github-api-1.55),  in the following commits to be specific:
* https://github.com/kohsuke/github-api/commit/16a6623095898f440d4ee3a6f414886117a25e3a
* https://github.com/kohsuke/github-api/commit/e3d6e08b6a1b5077eca658d4ea6cd0a3d98b1a98

Since @esteve has removed the UI changes and the patch in this PR is now very minimal (basically upgrading the version dependency and passing the job name) please consider merging this and releasing a new version of the plugin.

KostyaSha:
@dirk-thomas only after it will be verified to work with folders plugin.

dirk-thomas:
Earlier I was under the impression that this API is being used by the "GitHub Pull Request Builder" plugin. It turned out that that plugin is now calling `createCommitStatus()` directly without relying in the API changed in this PR. Since I don't have a use case to test this PR with anymore I won't look into any testing.

janpapenbrock:
Would love to see this merged.

Use case: Right now two concurrent test jobs would both send first Pending then Success/Failure state on the commit - and the last one would stick to the commit.

Contexts are a great way to solve this.

We rely on jenkings-github directly to update the status, not Github Pull Request Builder in between.

If I can contribute to testing, I'm happy to do so.

KostyaSha:
@janpapenbrock please test current patch with folders plugin.

janpapenbrock:
@KostyaSha - how could I do that? What is the folders plugin? Do you mean this plugin https://wiki.jenkins-ci.org/display/JENKINS/CloudBees+Folders+Plugin ?

KostyaSha:
@janpapenbrock yes

KostyaSha:
@janpapenbrock install this hpi https://jenkins.ci.cloudbees.com/job/plugins/job/github-plugin/80/com.coravy.hudson.plugins.github$github/ and folder plugin, then create two subfolders and job in it, then run build and check how status will look on GH.

janpapenbrock:
Alright, I compiled the snapshot and tested with Folders plugin. Results here https://github.com/janpapenbrock/github-plugin-test

Appears there is a default context created which never finishes. Its build number '#7' actually links to test1's build '#7'.

KostyaSha:
@janpapenbrock that was i fair, project.getFullName() should be used instead of buildFullName(). Wait few minutes, i will provide fix. Could you join IRC #freenode channel?

janpapenbrock:
@esteve thanks for your initial effort!

esteve:
@janpapenbrock no problem, thank you for testing this PR. I'm happy to see this feature getting merged in one way or another.

PR #51, opened 2015-03-20
Fix issue with jobs using the CloudBees folders plugin being deleted

ryangardner:
This should fix https://issues.jenkins-ci.org/browse/JENKINS-25127 - I tested it on my installation and it no longer deleting the hooks for jobs that are in folders


ryangardner:
In case the link above wasn't enough to trigger any bot linking issues, this addresses [JENKINS-25127]

leandro-lucarella-sociomantic:
I hope this is fixed soon, is really complicating my setup :S

Thanks for the PR!

KostyaSha:
Looks ok

PR #52, opened 2015-03-27
Some enhancements

KostyaSha:


src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -141,27 +142,31 @@ public File getLogFile() {
...
-                                if(createJenkinsHook(repo, getDescriptor().getHookUrl())) {
-                                    continue OUTER;
-                                }
-                            } catch (Throwable e) {
-                                LOGGER.log(Level.WARNING, "Failed to add GitHub webhook for "+name, e);
+            registerHooks();
+        }
+    }
+
+    public void registerHooks() {
```
oleg-nenashev: Do you really need to keep it public? If yes, Javadoc is required

KostyaSha: Commit description says "Move to separate method to have ability re-register hooks" from groovy for example.

oleg-nenashev: Yes, My bad. I finally bothered to read the PR

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -320,6 +326,25 @@ public FormValidation doCheckHookUrl(@QueryParameter String value) {
 
         }
 
+        public FormValidation doReRegister() {
+            if (!manageHook) {
+                return FormValidation.error("Works only when jenkins manages hooks");
```
oleg-nenashev: **J**enkins

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -320,6 +326,25 @@ public FormValidation doCheckHookUrl(@QueryParameter String value) {
...
+            }
+
+            int triggered = 0;
+            for (AbstractProject<?,?> job : Jenkins.getInstance().getAllItems(AbstractProject.class)) {
```
oleg-nenashev: Check for null?
Somebody can call the method before the full startup

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -320,6 +326,25 @@ public FormValidation doCheckHookUrl(@QueryParameter String value) {
...
+                GitHubPushTrigger trigger = job.getTrigger(GitHubPushTrigger.class);
+                if (trigger!=null) {
+                    LOGGER.log(Level.FINE, "Calling registerHooks() for {0}", job.getFullName());
+                    trigger.registerHooks();
+                    triggered++;
+                }
+            }
+
+            LOGGER.log(Level.INFO, "Called registerHooks() for {0} jobs", triggered);
```
oleg-nenashev: I would recommend to have something human-readable. Jenkins users do not know what <code>registerHooks()</code> means

KostyaSha: This is log, it also has different hashes for objects, weird class names and etc.

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -228,6 +233,7 @@ public void writeLogTo(XMLOutput out) throws IOException {
 
     @Extension
     public static class DescriptorImpl extends TriggerDescriptor {
+        private static final Logger LOGGER = Logger.getLogger(DescriptorImpl.class.getName());
```
oleg-nenashev: I would recommend a logger for <code>GitHubPushTrigger.class</code>. Such approach would be much more useful for logs filtering

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
KostyaSha: Like in other code.

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
KostyaSha: I need see log from Descriptor and not from multiple triggers. Descriptor is singleton and i don't need to track from what Describable instance this log was printed.

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -141,27 +142,35 @@ public File getLogFile() {
...
-                            } catch (Throwable e) {
-                                LOGGER.log(Level.WARNING, "Failed to add GitHub webhook for "+name, e);
+            registerHooks();
+        }
+    }
+
+    /**
+     * Tries to register hook for current associated job.
+     * Useful for using from groovy scripts.
+     */
```
oleg-nenashev: @since TODO

KostyaSha:
rebased, force-pushed

KostyaSha:
User confirmed that it works

PR #53, opened 2015-04-16
Allow registering as a webhook instead of the GitHub Jenkins Service

s0undt3ch:
Refs https://issues.jenkins-ci.org/browse/JENKINS-27842

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
+++ b/src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
@@ -39,6 +39,7 @@ import net.sf.json.JSONObject;
 import org.apache.commons.codec.binary.Base64;
 import org.apache.commons.jelly.XMLOutput;
 import org.jenkinsci.main.modules.instance_identity.InstanceIdentity;
+import org.kohsuke.github.GHEvent;
 import org.kohsuke.github.GHException;
 import org.kohsuke.github.GHRepository;
 import org.kohsuke.stapler.DataBoundConstructor;
@@ -176,7 +177,13 @@ public class GitHubPushTrigger extends Trigger<AbstractProject<?,?>> implements 
 
     private boolean createJenkinsHook(GHRepository repo, URL url) {
         try {
-            repo.createHook("jenkins", Collections.singletonMap("jenkins_hook_url", url.toExternalForm()), null, true);
+            if (getDescriptor().isUseWebHooks()) {
+                Collection<GHEvent> events = new ArrayList();
+                events.add(GHEvent.PULL_REQUEST);
+                repo.createWebHook(new URL(url.toExternalForm()), events);
+            } else {
+                repo.createHook("jenkins", Collections.singletonMap("jenkins_hook_url", url.toExternalForm()), null, true);
+            }
             return true;
         } catch (IOException e) {
             throw new GHException("Failed to update jenkins hooks", e);
@@ -242,6 +249,7 @@ public class GitHubPushTrigger extends Trigger<AbstractProject<?,?>> implements 
         private transient final SequentialExecutionQueue queue = new SequentialExecutionQueue(MasterComputer.threadPoolForRemoting);
 
         private boolean manageHook;
+        private boolean useWebHooks;
         private String hookUrl;
         private volatile List<Credential> credentials = new ArrayList<Credential>();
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -176,7 +177,13 @@ public void run() {
 
     private boolean createJenkinsHook(GHRepository repo, URL url) {
         try {
-            repo.createHook("jenkins", Collections.singletonMap("jenkins_hook_url", url.toExternalForm()), null, true);
+            if (getDescriptor().isUseWebHooks()) {
+                Collection<GHEvent> events = new ArrayList();
+                events.add(GHEvent.PULL_REQUEST);
```
KostyaSha: This plugin not about PRs

s0undt3ch: Sorry, failed on the event(I was also working on a separate plugin). I'll fix that.

s0undt3ch:
Fixed the event and since this is not using the Jenkins github service, we need to react accordingly to github's ping events.

KostyaSha:
If you don't want to use automanaged hook, then you can specify "manage hook manually" in global configuration.

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -176,7 +176,12 @@ public void run() {
 
     private boolean createJenkinsHook(GHRepository repo, URL url) {
         try {
-            repo.createHook("jenkins", Collections.singletonMap("jenkins_hook_url", url.toExternalForm()), null, true);
+            if (getDescriptor().isUseWebHooks()) {
+                // No need to pass events since the default event is PUSH
+                repo.createWebHook(new URL(url.toExternalForm()));
+            } else {
+                repo.createHook("jenkins", Collections.singletonMap("jenkins_hook_url", url.toExternalForm()), null, true);
```
KostyaSha: So what is the difference between this two hooks then?

KostyaSha:
I will try understand your use case again, but it looks like you are trying to integrate gh plugin with other requirements in a wrong way. I'm working on searching way of providing some APIs for gh related plugin for working with gh hooks, maybe it will help you. I will let you know when i found some solution.

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
s0undt3ch: The main and most important difference is that you're allowed to register more than one Jenkins instance on the same repository, something that using the "jenkins" hook does not allow.

s0undt3ch:
Also, I do want to use the auto-managed hook feature. I don't want to register the hooks by hand.

s0undt3ch:
The use case is 3 Jenkins instances that need to be triggered from the same repository, and, since only switching the way hooks are registered allows the same user experience, I felt that adding a _flag_ to switch the register behavior is better than creating a duplicate plugin for which the only difference is the way hooks are registered.

KostyaSha:
IMO just third choice "don't manage hooks" should be right

KostyaSha:
Problem that service is limited only to one jenkins is GH issue. I will try get answer from their support.

s0undt3ch:
Don't manage hooks requires you to add hooks by hand on the GitHub UI. Not optimal, and this ain't that drastic of a change. But I'll wait for your findings.

KostyaSha:
That means that solution provided by github is bad and need to be extended/resolved.

s0undt3ch:
Yes, but does the github hook apps allow multiple registrations from the base? Or is this just an issue with the Jenkins github hook app?

s0undt3ch:
Either way, this solves the immediate problem.

KostyaSha:
According to documentations services are single instances. I sent email to github support to get some information how they see their feature for end-users. If it deprecated feature, then it makes sense to not using services and describe github hooks configuration in documentation.

KostyaSha:
Support said they doesn't plan extending services and they doesn't accept service templates anymore. It makes sense to migrate plugin to using plain hooks. I'm thinking how it better can be done.

s0undt3ch:
OK. Its probably best to add some documentation to the plugin about the move and eventually ask GitHub to remove the service a few versions down the road to allow for users to migrate to a newer plugin version.

KostyaSha:
You still can use "service" from gh UI, it just a hardcode of "jenkins" + "push". Hook managment is internal, if you choosed "manage manually" then no changes required. Only automanage, but automanaged hooks can be easily upgraded :)

samrocketman:
If you register 3 separate applications (in GitHub) and give each Jenkins separate application credentials then your three separate instances will register hooks for each.  You can't make use of the Jenkins service integration with multiple instances if you're only using one application.  Generate new app codes for each.

s0undt3ch:
I am using one application per Jenkins. But a repository can only have a single Jenkins service integration at a time, independent of the applications... Last time I checked...

samrocketman:
Ah then might want to check again.  I accidentally registered two applications using different API tokens and the service duplicated registration.  I had regenerated the token because I forgot it.

samrocketman:
Though to be clear I still think it's useful to provide registering as webhooks automatically.  I only piped up with a workaround for you in the mean time.

KostyaSha:
Please try 1.12.0

PR #54, opened 2015-04-27
Add ability to specify GitHub events.

marcus-chan:
Adds functionality to support other GitHub event types other than push. Extension of https://github.com/jenkinsci/github-plugin/pull/53.

[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/54)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -9,7 +9,7 @@
   <groupId>com.coravy.hudson.plugins.github</groupId>
   <artifactId>github</artifactId>
   <packaging>hpi</packaging>
-  <version>1.11.3</version>
+  <version>1.11.4-SNAPSHOT</version>
   <name>GitHub plugin</name>
   <url>http://wiki.jenkins-ci.org/display/JENKINS/Github+Plugin</url>
 
@@ -82,7 +82,7 @@
     <connection>scm:git:git://github.com/jenkinsci/github-plugin.git</connection>
     <developerConnection>scm:git:git@github.com:jenkinsci/github-plugin.git</developerConnection>
     <url>https://github.com/jenkinsci/github-plugin</url>
-    <tag>github-1.11.3</tag>
+    <tag>HEAD</tag>
   </scm>
 
   <repositories>
--- a/pom.xml
+++ b/pom.xml
@@ -36,7 +36,7 @@
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>github-api</artifactId>
-      <version>1.59</version>
+      <version>1.67</version>
     </dependency>
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -9,7 +9,7 @@
   <groupId>com.coravy.hudson.plugins.github</groupId>
   <artifactId>github</artifactId>
   <packaging>hpi</packaging>
-  <version>1.11.2</version>
+  <version>1.11.3-SNAPSHOT</version>
   <name>GitHub plugin</name>
   <url>http://wiki.jenkins-ci.org/display/JENKINS/Github+Plugin</url>
 
@@ -82,7 +82,7 @@
     <connection>scm:git:git://github.com/jenkinsci/github-plugin.git</connection>
     <developerConnection>scm:git:git@github.com:jenkinsci/github-plugin.git</developerConnection>
     <url>https://github.com/jenkinsci/github-plugin</url>
-    <tag>github-1.11.2</tag>
+    <tag>HEAD</tag>
   </scm>
 
   <repositories>
--- a/pom.xml
+++ b/pom.xml
@@ -9,7 +9,7 @@
   <groupId>com.coravy.hudson.plugins.github</groupId>
   <artifactId>github</artifactId>
   <packaging>hpi</packaging>
-  <version>1.11</version>
+  <version>1.12-SNAPSHOT</version>
   <name>GitHub plugin</name>
   <url>http://wiki.jenkins-ci.org/display/JENKINS/Github+Plugin</url>
 
@@ -82,7 +82,7 @@
     <connection>scm:git:git://github.com/jenkinsci/github-plugin.git</connection>
     <developerConnection>scm:git:git@github.com:jenkinsci/github-plugin.git</developerConnection>
     <url>https://github.com/jenkinsci/github-plugin</url>
-    <tag>github-1.11</tag>
+    <tag>HEAD</tag>
   </scm>
 
   <repositories>
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>1.532</version>
+    <version>1.610</version>
   </parent>
 
   <groupId>com.coravy.hudson.plugins.github</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -9,7 +9,7 @@
   <groupId>com.coravy.hudson.plugins.github</groupId>
   <artifactId>github</artifactId>
   <packaging>hpi</packaging>
-  <version>1.11.2-SNAPSHOT</version>
+  <version>1.11.2</version>
   <name>GitHub plugin</name>
   <url>http://wiki.jenkins-ci.org/display/JENKINS/Github+Plugin</url>
 
@@ -82,7 +82,7 @@
     <connection>scm:git:git://github.com/jenkinsci/github-plugin.git</connection>
     <developerConnection>scm:git:git@github.com:jenkinsci/github-plugin.git</developerConnection>
     <url>https://github.com/jenkinsci/github-plugin</url>
-    <tag>HEAD</tag>
+    <tag>github-1.11.2</tag>
   </scm>
 
   <repositories>
--- a/pom.xml
+++ b/pom.xml
@@ -9,7 +9,7 @@
   <groupId>com.coravy.hudson.plugins.github</groupId>
   <artifactId>github</artifactId>
   <packaging>hpi</packaging>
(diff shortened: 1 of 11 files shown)


KostyaSha:
:-1: coding style changes, can't follow real changes. Plugin is not about any other event types rather then PUSH.

jtillman:
:+1: PUSH event seems real limited. In my setup, we just build during Tagging/Branch Creation. This make sense for changes that go out in batch rather than individual items.

KostyaSha:
This change has problems and useless complexity in every line of change, have no even enough time to comment them all. 

@jtillman please open and described your issue in jira.

jtillman:
@KostyaSha I think my issues are already describe in an existing ticket: https://issues.jenkins-ci.org/browse/JENKINS-27842

KostyaSha:
@jtillman about "webhooks" vs "services" i already answered in your PR. If push ignores tags, then it other issue. Don't mix many issues in one.

This PR wants add extra configurations and end user complexity without describing any real case. Allowing register many events in plugin that doesn't support this events is bad behaviour. In the same way gh plugin can do coffee.

jtillman:
Sorry but I couldn't fully understand your last explanation. What I think that the ticket references is the ability to subscribe different jobs to different repositories and events? Briefly looking at the UI change in this PR, it seems it just adds an advanced configuration which doesn't effect the existing use case. I'd be willing to help clean it up if its just stylistic issues. I'd like to understand better your concern around having this more configurable.

KostyaSha:
Events registered by this plugin incoming into this plugin, registering useless event that doesn't handled by this plugin is redundant configuration without any real case. Listener is the right direction but UI configuration in this plugin absolutely useless. 
Having variable for providing other variable optionally is also not good variant of UI configuration.
GitHubCause looks like a hack for his own single use case.  
I split current plugin issues in 2 issues https://issues.jenkins-ci.org/browse/JENKINS-28138 and https://issues.jenkins-ci.org/browse/JENKINS-28139 
And firstly i think should be resolved #39 , then migration to webhooks and finally extension.

pom.xml
```diff
@@ -3,7 +3,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>1.532</version>
+    <version>1.610</version>
```
KostyaSha: Reason?

marcus-chan: Jenkins has a bug https://issues.jenkins-ci.org/browse/JENKINS-18537 that has problems with JDK8.

adepue:
:+1:  Being able to wire to more than PUSH events would be nice for me as well.

pom.xml
KostyaSha: Is it compile time issue of your java8 or runtime of all jenkins?

rogerhu: It's a runtime issue for Jenkins when using this plugin and running mvi hpi:run.  The moment you startup Jenkins with this plugin, you should be able to see the issue.

rogerhu:
+1 -- allowing more granular GitHub event types makes sense and enabling this feature helps automate the process of installing webhooks across multiple repo's that only need to listen for certain events.

pom.xml
oleg-nenashev: Makes sense to decrease the dependency to 1.609.1 when it gets released

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -53,8 +56,22 @@
  * @author Kohsuke Kawaguchi
  */
 public class GitHubPushTrigger extends Trigger<AbstractProject<?,?>> implements GitHubTrigger {
+
+    private boolean overrideEventTypes;
+    private EnumSet<GHEvent> eventTypes;
+
     @DataBoundConstructor
-    public GitHubPushTrigger() {
```
oleg-nenashev: Binary compatibility loss. The original constructor should be retained somehow

src/main/java/com/cloudbees/jenkins/GitHubCause.java
```diff
@@ -0,0 +1,41 @@
...
+import java.io.IOException;
+
+/**
+ * UI object that says a build is started by GitHub post-commit hook.
+ *
+ * @author Marcus Chan
+ */
+public class GitHubCause extends SCMTriggerCause {
+
+    private String eventType;
```
oleg-nenashev: I would explicitly mark fields and methods as <code>@CheckForNull</code>. There's inconsistency in the code: you default the value to empty string, but check against null in <code>getShortDescription</code>

src/main/java/com/cloudbees/jenkins/GitHubCause.java
```diff
@@ -0,0 +1,41 @@
...
+
+    private String name;
```
oleg-nenashev: Makes sense to rename to "author"

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -174,9 +200,37 @@ public void run() {
         });
     }
 
+    private void updateWebHook(GHRepository repo, URL url) {
+        try {
+            String urlExternalForm = url.toExternalForm();
+            GHHook hook = null;
+            for (GHHook h : repo.getHooks()) {
+                if (h.getName().equals("web") && h.getConfig().get("url").equals(urlExternalForm)) {
+                    hook = h;
```
oleg-nenashev: and break?

src/main/java/com/cloudbees/jenkins/GitHubWebHook.java
```diff
@@ -163,16 +164,23 @@ public void doIndex(StaplerRequest req, StaplerResponse rsp) {
...
+                    GHEvent.valueOf(eventType.toUpperCase());
+                    String payload = req.getParameter("payload");
+                    if (payload == null) {
+                        throw new IllegalArgumentException("Not intended to be browsed interactively (must specify payload parameter). " +
+                                "Make sure payload version is 'application/vnd.github+form'.");
+                    }
+
+                    processGitHubPayload(eventType, payload, GitHubPushTrigger.class);
+                } catch (IllegalArgumentException e) {
+                    throw new IllegalArgumentException("Github Webhook event of type " + eventType + " is not supported.");
```
oleg-nenashev: Also catches the internal throw with a loss of data

src/main/java/com/cloudbees/jenkins/GitHubWebHook.java
```diff
@@ -219,8 +232,13 @@ public void processGitHubPayload(String payload, Class<? extends Trigger<?>> tri
             } finally {
                 SecurityContextHolder.getContext().setAuthentication(old);
             }
-            for (Listener listener: Jenkins.getInstance().getExtensionList(Listener.class)) {
-                listener.onPushRepositoryChanged(pusherName, changedRepository);
+
+            for (Listener listener : Jenkins.getInstance().getExtensionList(Listener.class)) {
+                listener.onRepositoryChanged(eventType, name, changedRepository);
+                if ("push".equals(eventType)) {
```
oleg-nenashev: should be checked before launching the for cycle

src/main/java/com/cloudbees/jenkins/GitHubWebHook.java
```diff
@@ -254,7 +272,10 @@ public static GitHubWebHook get() {
          * @param changedRepository the changed repository.
          * @since 1.8
          */
+        @Deprecated
         public abstract void onPushRepositoryChanged(String pusherName, GitHubRepositoryName changedRepository);
+
+        public abstract void onRepositoryChanged(String eventType, String name, GitHubRepositoryName changedRepository);
```
oleg-nenashev: Javadoc is important for an extension point

oleg-nenashev:
The PR cannot be merged in the current state. 
I suppose #57 and #60 supersede this PR

rogerhu:
So are you OK with exposing event types in general that can be used for GitHub webhooks?

I've briefly looked at the new changes but wondering if you have any thoughts about the best way to integrate some of this stuff?

KostyaSha:
@rogerhu you will need create separate plugin that will register extension points and consume anything you want. gh-plugin itself needs only pushes, pr plugin needs issue+pr, your plugin whatever event type you want.

rogerhu:
Can the github-plugin be configured like the GitHub to be able to control what events to trigger?

The reason we like the GitHub plugin to do this it can manage all the webhooks related to Jenkins and GitHub integration.  We only need to trigger when there are builds/releases issuse and not for every push.

image

KostyaSha:
Interesting idea, currently i'm working on new pullrequest plugin and it supports more triggering options. Probably we can try to do it more generic and call as github-trigger-plugin. 
@lanwen what do you think?

rogerhu:
That would be great if you could add support... :)  That was the intent of this PR I believe... any branch that you currently working on?

KostyaSha:
lanwen is working on credentials now, later we will think how better implement other triggers.

lanwen:
yep, the next issue will be common gh trigger

rogerhu:
Happy to help test...just point me to a branch when you're ready!

KostyaSha:
@rogerhu we will notify you and request for testing. Atm we need prepare API for reusing them. For example credentials support.

src/main/java/com/cloudbees/jenkins/GitHubCause.java
```diff
@@ -0,0 +1,41 @@
...
+        super(pollingLog);
+        this.eventType = eventType;
+        this.name = name;
+    }
+
+    @Override
+    public String getShortDescription() {
+        String event = eventType != null ? eventType : "";
+        String author = name != null ? name : "";
+        return "Started by GitHub " + event + " by " + author;
```
shubhampa: Why constants, you can create constants at common constants file. use them here

PR #58, opened 2015-06-23
Ignorable pusher

aderigs:
This defines a regular expression to which the name or email of an ignorable pusher is to be matched. Such a push will be ignored. Normally a preceding push of Jenkins him self could be prevented that way.

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/GitHubTrigger.java
+++ b/src/main/java/com/cloudbees/jenkins/GitHubTrigger.java
@@ -50,4 +50,15 @@ public interface GitHubTrigger {
             }
         }
     }
+
+    String pushBy();
+
+    /**
+     * Gets the regular expression to which the name or email of an ignorable pusher is to be
+     * matched.
+     *
+     * @return the regular expression to which the name or email of an ignorable pusher is to be
+     *         matched.
+     */
+    String getIgnorablePusher();
 }
--- a/pom.xml
+++ b/pom.xml
@@ -47,7 +47,7 @@
     <dependency>
       <groupId>org.jmock</groupId>
       <artifactId>jmock-junit4</artifactId>
-      <version>2.5.1</version>
+      <version>2.8.1</version>
       <scope>test</scope>
     </dependency>
     <dependency>
--- a/src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
+++ b/src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
@@ -4,10 +4,10 @@ import hudson.Extension;
 import hudson.Util;
 import hudson.console.AnnotatedLargeText;
 import hudson.model.Action;
-import hudson.model.Hudson;
-import hudson.model.Hudson.MasterComputer;
 import hudson.model.Item;
 import hudson.model.AbstractProject;
+import hudson.model.Hudson;
+import hudson.model.Hudson.MasterComputer;
 import hudson.model.Project;
 import hudson.triggers.Trigger;
 import hudson.triggers.TriggerDescriptor;
@@ -33,6 +33,8 @@ import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
+import javax.inject.Inject;
+
 import jenkins.model.Jenkins;
 import net.sf.json.JSONObject;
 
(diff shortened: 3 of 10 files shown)


KostyaSha:
This should work and be configured in GIT scm configuration.

aderigs:
How would you do this? The only possibility I have found would be something like "Polling ignores commits from certain users". But this would involve committers but not pushers.

felfert:
@KostyaSha:

Wrong. You are confusing this with the existing featue of ignoring a commit-author. The pusher is not necessarily the committer!

src/main/java/com/cloudbees/jenkins/GitHubTrigger.java
```diff
@@ -50,4 +50,15 @@ public void parseAssociatedNames(AbstractProject<?, ?> job, Collection<GitHubRep
...
+    String pushBy();
+
+    /**
+     * Gets the regular expression to which the name or email of an ignorable pusher is to be
+     * matched.
+     *
+     * @return the regular expression to which the name or email of an ignorable pusher is to be
+     *         matched.
+     */
+    String getIgnorablePusher();
```
KostyaSha: this breaks backward compatibility

ouaibsky:
Hi
I think this PR solve this issue: https://issues.jenkins-ci.org/browse/JENKINS-24208
I missed this PR and I create another one https://github.com/jenkinsci/github-plugin/pull/83  (probably more basic then this one)

@KostyaSha 
It can not be done in GIT scm configuration, have look at jira comment.

Anyway if you can validate this PR or mine, it's critical for our jenkins/github dev. pipeline

Regards
Christophe

igreenfield:
When this should be merged?

lanwen:
This will be never merged it current state, because of it provides only one case of filtering push (there is same for branches, committers, titles). But it still open because of its a good idea and not bad implementation. But we beed some more common implementation of filtering feature. 

You can use hpi from the jenkins build to use this custom implementation (thanks @aderigs for rebasing)

igreenfield:
What about using the Run Condition Plugin to support all kinds of filtering?

KostyaSha:
You can solve this issue by using intermediate `decider` job that can do additional verification and trigger real builds.
Pushers need to be implemented in more generic security API. API will be clear when you implement all possible gh triggers. I started this work in https://github.com/jenkinsci/github-integration-plugin .

@lanwen on other hand this 'hook -> gitscm poll' trigger wouldn't fit in any good APIs so probably such feature may be added to this trigger. But everything should be localised somewhere in code unique to this trigger.

KostyaSha:
Keeping in mind that gh utilises git polling log i can suggest try set exclusions using git configuration features (i.e. user exclusions). It wouldn't use gh names, but should work fine if people setting right names in git commits.

igreenfield:
When I install the plugin with this change I don't see where to configure the new configuration!

aderigs:
Yes. The last version has introduced src/main/resources/com/cloudbees/jenkins/GitHubPushTrigger/config.groovy which hides src/main/resources/com/cloudbees/jenkins/GitHubPushTrigger/config.jelly. I think the clean way would be to migrate config.jelly into the new config.groovy. But I do not know how to do this.

igreenfield:
the groovy file should be:


```groovy
package com.cisco.jenkins.GitHubPushTrigger

import com.cisco.jenkins.GitHubPushTrigger

def f = namespace(lib.FormTagLib);

tr {
    td(colspan: 4) {
        f.entry(title: _("Ignorable Pusher"), field: "ignorablePusher") {
            f.textbox()
        }
    }
}

script(src: "${rootURL}${h.getResourcePath()}/plugin/github/js/warning.js")
script {
    text("""
InlineWarning.setup({ 
    id: 'gh-hooks-warn',
    url: ${descriptor.getCheckMethod('hookRegistered').toCheckUrl()}, 
    input: 'input[name="${GitHubPushTrigger.class.getName().replace(".", "-")}"]'
}).start();
""")
}
```

KostyaSha:
Please rebase and not merge

ouaibsky:
@KostyaSha, it's why I did

> git pull --rebase origin master

But don't know why, sounds not working, maybe missing some settings ?

KostyaSha:
That message to @aderigs who owns this PR

kdawgwilk:
Does this support ignoring specific commit messages as well?

aderigs:
No. It ignores the pusher by name or email. If you want to ignore by specific commit messages you could use the git-plugin.

kdawgwilk:
I have not been able to find a solution for this using the github folder plugin, which manages webhooks, and I believe uses this plugin.  It automatically sets up the job and doesn't give you anywhere to configure the SCM that was setup. I just need a way to ignore webhook triggers if a specific commit message exists.  I was hoping this would solve the issue :(

PR #57, opened 2015-06-16
[JENKINS-28138] Migration to webhooks with ext point

lanwen:
Working to migrate to webhooks instead of services ([JENKINS-28138](https://issues.jenkins-ci.org/browse/JENKINS-28138))

The plan is
- [x] Incapsulate working with hooks in separate class
  This can help to make process testable and simplify classes with other logic (Cleaner, Triggers)
- [x] Migrate to new webhook manager class in cleaner
- [x] Migrate to new manager in push trigger
- [x] Use in `Cleaner` *queue interface* instead of synchronized block. It can simplify add and fetch operations on how to add for cleaning new names
- [x] Clean old service hooks and replace it with webhooks on job reconfig or manual reregister btn
- [x] Add extension point to contribute new events to hook for some jobs
   New events will be merged with old ones from existent hooks 
- ~~[ ] Extend extension point to contribute hook events to  make possible parse events it interested in~~ (will be done in separate PR)
- [x] Write tests

## Next step is #60 

[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/57)

PR #59, opened 2015-07-03
[JENKINS-27136] Workflow support for GitHub Push Triggers

amuniz:
https://issues.jenkins-ci.org/browse/JENKINS-27136

Currently Workflow jobs can not be triggered when a GitHub push hook is received, this PR makes it possible.

In this [document](https://github.com/jenkinsci/workflow-plugin/blob/master/COMPATIBILITY.md#historical-background) there is information about the required changes in plugins to make them compatible with Workflow. This changes have been applied here.

@reviewbybees


pom.xml
```diff
@@ -3,8 +3,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-      <!--JENKINS-18537-->
-    <version>1.554.1</version>
+    <version>1.580.1</version>
```
KostyaSha: reason?

amuniz: The minimun Workflow compatible core version is 1.580.1

s0undt3ch:
I'm **realy** after this ... :+1:

pom.xml
KostyaSha: compatible with what?

KostyaSha:
:-1: until #57 will be merged. Also this change looks like breaking interfaces change.

src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
```diff
@@ -58,15 +62,17 @@ protected EnvVars buildEnv(AbstractProject<?, ?> job) {
...
-            if (scm instanceof GitSCM) {
-                GitSCM git = (GitSCM) scm;
-                for (RemoteConfig rc : git.getRepositories()) {
-                    for (URIish uri : rc.getURIs()) {
-                        String url = env.expand(uri.toString());
-                        GitHubRepositoryName repo = GitHubRepositoryName.create(url);
-                        if (repo != null) {
-                            r.add(repo);
+        protected static void addRepositories(Collection<? extends SCM> scms, EnvVars env, Collection<GitHubRepositoryName> r) {
+            for(SCM scm : scms) {
```
KostyaSha: formating

src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
```diff
@@ -90,20 +96,19 @@ public void parseAssociatedNames(AbstractProject<?, ?> job, Collection<GitHubRep
...
-                EnvVars env = buildEnv(job);
-
-                MultiSCM multiSCM = (MultiSCM) job.getScm();
-                List<SCM> scmList = multiSCM.getConfiguredSCMs();
-                for (SCM scm : scmList) {
-                    addRepositories(scm, env, result);
+        public void parseAssociatedNames(Job<?, ?> job, Collection<GitHubRepositoryName> result) {
+            SCMTriggerItem item = SCMTriggerItems.asSCMTriggerItem(job);
+            EnvVars env = buildEnv(job);
+            for(SCM scm : item.getSCMs()) {
```
KostyaSha: formatting

pom.xml
s0undt3ch: _[You need to be running a sufficiently recent Jenkins release: an LTS in the 1.580.x line or newer](https://github.com/jenkinsci/workflow-plugin#installation)_

amuniz: Any Workflow related feature requires core 1.580.X. Since I'm adding Workflow compatibility, the plugin requires 1.580.X now.

See https://github.com/jenkinsci/workflow-plugin/blob/master/README.md#installation

KostyaSha: @s0undt3ch this is "Installation" part related to workflow itself.

KostyaSha: > Any Workflow related feature requires core 1.580.X.

This doesn't mean that you must update core dependencies in non workflow plugins

amuniz:
@KostyaSha the only way to make this compatible with Workflow jobs is changing interfaces. Not sure how this impact in #57

pom.xml
s0undt3ch: As @amuniz mentioned, in order for this plugin to be supported by the workflow plugin, the jenkins version needs to be updated because the workflow plugin requires a higher jenkins minimal version

KostyaSha: @s0undt3ch https://github.com/jenkinsci/github-plugin/pull/59#discussion_r33868443

lanwen:
please wait for #57 because of hard rebase process instead

pom.xml
daniel-beck: @KostyaSha 

> This doesn't mean that you must update core dependencies in non workflow plugins

`SCMTriggerItem` was added in 1.568. 1.580.x is the next LTS after that.

KostyaSha: And this is right answer

tfennelly:
@amuniz for the methods that were modified, maybe you could override them instead, deprecating the old method signatures?

src/main/java/com/cloudbees/jenkins/Cleaner.java
```diff
@@ -52,10 +56,13 @@ protected void doRun() throws Exception {
         }
 
         // subtract all the live repositories
-        for (AbstractProject<?,?> job : Hudson.getInstance().getAllItems(AbstractProject.class)) {
-            GitHubPushTrigger trigger = job.getTrigger(GitHubPushTrigger.class);
-            if (trigger!=null) {
-                names.removeAll(GitHubRepositoryNameContributor.parseAssociatedNames(job));
+        for (Job<?,?> job : Jenkins.getInstance().getAllItems(Job.class)) {
```
tfennelly: I think you'll need to add a `null` check on `Jenkins.getInstance()`. Otherwise, we'll be getting FindBugs errors I think.

src/main/java/com/cloudbees/jenkins/Cleaner.java
```diff
@@ -52,10 +56,13 @@ protected void doRun() throws Exception {
...
+            ParameterizedJobMixIn.ParameterizedJob pJob = (ParameterizedJobMixIn.ParameterizedJob) job;
+            for (Trigger trigger : pJob.getTriggers().values()) {
+                if(trigger instanceof GitHubPushTrigger) {
+                    names.removeAll(GitHubRepositoryNameContributor.parseAssociatedNames(job));
```
tfennelly: `parseAssociatedNames` is guaranteed to not return null (is `@Nonnull`) ?

src/main/java/com/cloudbees/jenkins/Cleaner.java
```diff
@@ -66,7 +73,7 @@ protected void doRun() throws Exception {
             for (GHRepository repo : r.resolve()) {
                 try {
                     removeHook(repo, Trigger.all().get(DescriptorImpl.class).getHookUrl());
-                    LOGGER.fine("Removed a hook from "+r+"");
+                    LOGGER.info("Removed a hook from "+r+"");
```
tfennelly: Maybe you meant to change this back to `fine` before commiting?

src/main/java/com/cloudbees/jenkins/Cleaner.java
KostyaSha: new jenkins has getJenkinsInstance instead

src/main/java/com/cloudbees/jenkins/Cleaner.java
KostyaSha: Maybe better ask why was it changed?

src/main/java/com/cloudbees/jenkins/Cleaner.java
```diff
@@ -52,10 +56,13 @@ protected void doRun() throws Exception {
...
 
         // subtract all the live repositories
-        for (AbstractProject<?,?> job : Hudson.getInstance().getAllItems(AbstractProject.class)) {
-            GitHubPushTrigger trigger = job.getTrigger(GitHubPushTrigger.class);
-            if (trigger!=null) {
-                names.removeAll(GitHubRepositoryNameContributor.parseAssociatedNames(job));
+        for (Job<?,?> job : Jenkins.getInstance().getAllItems(Job.class)) {
+            ParameterizedJobMixIn.ParameterizedJob pJob = (ParameterizedJobMixIn.ParameterizedJob) job;
+            for (Trigger trigger : pJob.getTriggers().values()) {
+                if(trigger instanceof GitHubPushTrigger) {
```
KostyaSha: formatting

src/main/java/com/cloudbees/jenkins/GitHubTrigger.java
```diff
@@ -44,8 +47,9 @@
     @Extension
     public static class GitHubRepositoryNameContributorImpl extends GitHubRepositoryNameContributor {
         @Override
-        public void parseAssociatedNames(AbstractProject<?, ?> job, Collection<GitHubRepositoryName> result) {
-            for (GitHubTrigger ght : Util.filter(job.getTriggers().values(),GitHubTrigger.class)) {
+        public void parseAssociatedNames(Job<?, ?> job, Collection<GitHubRepositoryName> result) {
+            ParameterizedJobMixIn.ParameterizedJob p = (ParameterizedJobMixIn.ParameterizedJob) job;
```
tfennelly: Maybe a type-check here first?

src/main/java/com/cloudbees/jenkins/GitHubWebHook.java
```diff
@@ -204,13 +208,20 @@ public void processGitHubPayload(String payload, Class<? extends Trigger<?>> tri
             Authentication old = SecurityContextHolder.getContext().getAuthentication();
             SecurityContextHolder.getContext().setAuthentication(ACL.SYSTEM);
             try {
-                for (AbstractProject<?, ?> job : Jenkins.getInstance().getAllItems(AbstractProject.class)) {
-                    GitHubTrigger trigger = (GitHubTrigger) job.getTrigger(triggerClass);
-                    if (trigger != null) {
+                for (Job<?, ?> job : Jenkins.getInstance().getAllItems(Job.class)) {
+                    GitHubTrigger ghTrigger = null;
+                    ParameterizedJobMixIn.ParameterizedJob pJob = (ParameterizedJobMixIn.ParameterizedJob) job;
```
tfennelly: Type-check?

src/main/java/com/cloudbees/jenkins/GitHubWebHook.java
```diff
@@ -204,13 +208,20 @@ public void processGitHubPayload(String payload, Class<? extends Trigger<?>> tri
...
+                    for (Trigger trigger : pJob.getTriggers().values()) {
+                        if (triggerClass.isInstance(trigger)) {
```
tfennelly: How about an `instanceof GitHubTrigger` check here instead? Might be a bit clearer and it looks like that's all you need, no?

src/main/java/com/cloudbees/jenkins/Cleaner.java
oleg-nenashev: Starting from 1.596 only AFAIK

KostyaSha: Then just copy-paste it realisation and put to static function imho.

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -114,7 +118,8 @@ public void run() {
                         LOGGER.log(Level.WARNING, "Failed to parse the polling log",e);
                         cause = new GitHubPushCause(pushBy);
                     }
-                    if (job.scheduleBuild(cause)) {
+                    ParameterizedJobMixIn scheduler = new ParameterizedJobMixIn() {@Override protected Job asJob() {return job;}};
```
KostyaSha: please don't collapse in one line, really difficult read.

src/main/java/com/cloudbees/jenkins/GitHubWebHook.java
KostyaSha: Can't `getAllItems` request this type?

src/main/java/com/cloudbees/jenkins/Cleaner.java
amuniz: Yes, it is. It returns an empty list if no names found.

src/main/java/com/cloudbees/jenkins/Cleaner.java
amuniz: It was changed because the log relative to adding the hook already was at info level, so I think the remove hook log (opposite operation) should be at info level too.

src/main/java/com/cloudbees/jenkins/GitHubWebHook.java
amuniz: Yes, it's what I need, but the right operand (usually ```GitHubTrigger```, but not always) is given as a method parameter, so can not use ```instanceof```.

src/main/java/com/cloudbees/jenkins/Cleaner.java
```diff
@@ -52,10 +56,13 @@ protected void doRun() throws Exception {
         }
 
         // subtract all the live repositories
-        for (AbstractProject<?,?> job : Hudson.getInstance().getAllItems(AbstractProject.class)) {
-            GitHubPushTrigger trigger = job.getTrigger(GitHubPushTrigger.class);
-            if (trigger!=null) {
-                names.removeAll(GitHubRepositoryNameContributor.parseAssociatedNames(job));
+        for (Job<?,?> job : getActiveInstance().getAllItems(Job.class)) {
+            ParameterizedJobMixIn.ParameterizedJob pJob = (ParameterizedJobMixIn.ParameterizedJob) job;
```
jglick: This will be a CCE if it is not in fact a `ParameterizedJob`. :bug:

src/main/java/com/cloudbees/jenkins/Cleaner.java
```diff
@@ -52,10 +56,13 @@ protected void doRun() throws Exception {
         }
 
         // subtract all the live repositories
-        for (AbstractProject<?,?> job : Hudson.getInstance().getAllItems(AbstractProject.class)) {
-            GitHubPushTrigger trigger = job.getTrigger(GitHubPushTrigger.class);
-            if (trigger!=null) {
-                names.removeAll(GitHubRepositoryNameContributor.parseAssociatedNames(job));
+        for (Job<?,?> job : getActiveInstance().getAllItems(Job.class)) {
```
jglick: Alternately, just use `getInstance`, and if null, skip this block.

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -255,7 +264,7 @@ public DescriptorImpl() {
 
         @Override
         public boolean isApplicable(Item item) {
-            return item instanceof AbstractProject;
+            return item instanceof Job;
```
jglick:
```java
&& SCMTriggerItems.asSCMTriggerItem(job) != null && job instanceof ParameterizedJobMixIn.ParameterizedJob
```

src/main/java/com/cloudbees/jenkins/GitHubRepositoryName.java
```diff
@@ -7,8 +7,6 @@
 import org.kohsuke.github.GitHub;
 
 import java.io.IOException;
-import java.net.MalformedURLException;
-import java.net.URL;
```
jglick: Better omitted from this PR since it does not contribute to the change.

src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
```diff
@@ -30,13 +34,13 @@
      * Looks at the definition of {@link AbstractProject} and list up the related github repositories,
      * then puts them into the collection.
      */
-    public abstract void parseAssociatedNames(AbstractProject<?,?> job, Collection<GitHubRepositoryName> result);
+    public abstract void parseAssociatedNames(Job<?,?> job, Collection<GitHubRepositoryName> result);
```
jglick: This will be binary- and source-incompatible for plugins implementing this extension point. I cannot find any, but still…safer to add a new overload, deprecate this one, and use delegation and `Util.isOverridden` as needed to make it compatible.

src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
```diff
@@ -30,13 +34,13 @@
...
 
     public static ExtensionList<GitHubRepositoryNameContributor> all() {
         return Jenkins.getInstance().getExtensionList(GitHubRepositoryNameContributor.class);
     }
 
-    public static Collection<GitHubRepositoryName> parseAssociatedNames(AbstractProject<?, ?> job) {
+    public static Collection<GitHubRepositoryName> parseAssociatedNames(Job<?, ?> job) {
```
jglick: Same here, though easier since you need only introduce the new method and deprecate the existing one.

src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
```diff
@@ -90,20 +96,19 @@ public void parseAssociatedNames(AbstractProject<?, ?> job, Collection<GitHubRep
...
-                for (SCM scm : scmList) {
-                    addRepositories(scm, env, result);
+        public void parseAssociatedNames(Job<?, ?> job, Collection<GitHubRepositoryName> result) {
+            SCMTriggerItem item = SCMTriggerItems.asSCMTriggerItem(job);
+            EnvVars env = buildEnv(job);
+            for (SCM scm : item.getSCMs()) {
+                if (scm instanceof MultiSCM){
+                    MultiSCM multiSCM = (MultiSCM) scm;
+                    List<SCM> scmList = multiSCM.getConfiguredSCMs();
+                    addRepositories(scmList, env, result);
```
jglick: `FromMultiSCM` can probably be deleted entirely (and the dependency on the `multiple-scms` plugin removed!) since [`Project.getSCMs` already handles this case](https://github.com/jenkinsci/jenkins/blob/0016f85a500ca74a90358177c3cdb51ad7c8e5d2/core/src/main/java/hudson/model/Project.java#L120-122).

src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
```diff
@@ -58,15 +62,17 @@ protected EnvVars buildEnv(AbstractProject<?, ?> job) {
...
-            if (scm instanceof GitSCM) {
-                GitSCM git = (GitSCM) scm;
-                for (RemoteConfig rc : git.getRepositories()) {
-                    for (URIish uri : rc.getURIs()) {
-                        String url = env.expand(uri.toString());
-                        GitHubRepositoryName repo = GitHubRepositoryName.create(url);
-                        if (repo != null) {
-                            r.add(repo);
+        protected static void addRepositories(Collection<? extends SCM> scms, EnvVars env, Collection<GitHubRepositoryName> r) {
+            for (SCM scm : scms) {
```
jglick: To minimize the diff, I would recommend moving the `for`-loop into the one caller of this method that will remain after `FromMultiSCM` is deleted.

src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
KostyaSha: @jglick could you describe why workflow requires from plugins using Job class that is not supported by plugin internals?

amuniz:
@jglick All your review comments have been fixed
@KostyaSha the last commit adds backward code compatibility on public interfaces

pom.xml
```diff
@@ -39,27 +38,15 @@
...
-      <version>0.12.1</version>
-    </dependency>
-    <dependency>
-      <groupId>org.jenkins-ci.plugins</groupId>
-      <artifactId>multiple-scms</artifactId>
-      <version>0.2</version>
-      <optional>true</optional>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-api</artifactId>
+      <version>1.6.6</version>
```
jglick: Any particular reason you are adding this dependency now?

src/test/java/com/cloudbees/jenkins/GitHubPushHookReceiverTest.java
```diff
@@ -0,0 +1,103 @@
...
+    public void receivePushHookOnWorkflow() throws Exception {
+        j.jenkins.getInjector().injectMembers(this);
+        WorkflowJob job = j.jenkins.createProject(WorkflowJob.class, "Test Workflow");
+
+        GitHubPushTrigger trigger = new GitHubPushTrigger();
+        trigger.start(job, false);
+        job.addTrigger(trigger);
+        job.setDefinition(new CpsFlowDefinition(
+            "node {\n" +
+            "    git credentialsId: '', url: 'https://github.com/amuniz/github-plugin.git'\n" +
```
jglick: Can be simplified to

```groovy
git 'https://github.com/amuniz/github-plugin.git'
```

src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
```diff
@@ -44,9 +70,17 @@
         return names;
```
jglick: More easily:

```java
return parseAssociatedNames((Job) job);
```

pom.xml
amuniz: ```org.slf4j:slf4j-api:1.6.6``` was in the classpath because ```org.jenkins-ci.plugins:git:2.0``` got it transitively, it was removed on ```org.jenkins-ci.plugins:git:2.3.5```. Since ```slf4j``` is directly used in the plugin code we have to add it explicitly (or stop using it, but this is out of the PR scope).

src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
```diff
@@ -29,13 +33,35 @@
     /**
      * Looks at the definition of {@link AbstractProject} and list up the related github repositories,
      * then puts them into the collection.
+     *
+     * @deprecated Use {@link GitHubRepositoryNameContributor#parseAssociatedNames(Job job, Collection result)}
```
jglick: I think you mean to use the syntax `{@link #parseAssociatedNames(Job, Collection)}`.

src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
```diff
@@ -46,7 +76,7 @@
 
 
     static abstract class AbstractFromSCMImpl extends GitHubRepositoryNameContributor {
```
jglick: BTW since this now has only one subclass, the code could be simplified a bit by collapsing the two classes into one.

src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
```diff
@@ -78,32 +108,13 @@ protected static void addRepositories(SCM scm, EnvVars env, Collection<GitHubRep
...
-
-                MultiSCM multiSCM = (MultiSCM) job.getScm();
-                List<SCM> scmList = multiSCM.getConfiguredSCMs();
-                for (SCM scm : scmList) {
-                    addRepositories(scm, env, result);
+        public void parseAssociatedNames(Job<?, ?> job, Collection<GitHubRepositoryName> result) {
+            SCMTriggerItem item = SCMTriggerItems.asSCMTriggerItem(job);
+            if (item != null) {
+                for (SCM scm : item.getSCMs()) {
+                    addRepositories(scm, buildEnv(job), result);
```
jglick: Could extract the result of `buildEnv(job)` out of the loop. Rarely matters, though.

src/test/java/com/cloudbees/jenkins/GitHubPushHookReceiverTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+        GitHubPushTrigger trigger = new GitHubPushTrigger();
+        trigger.start(job, false);
+        job.addTrigger(trigger);
+        job.setScm(scm);
+
+        String payload = IOUtils.toString(getClass().getResourceAsStream("/com/cloudbees/jenkins/plugins/github/payload2.json"));
+        receiver.processGitHubPayload(payload, GitHubPushTrigger.class);
+        Run build = waitForBuild(1, job);
+        assertTrue("Build was triggered but did not success: " + build.getLog(300), Result.SUCCESS.equals(build.getResult()));
```
jglick: Simpler to use `j.assertBuildStatusSuccess(build)`.

src/test/java/com/cloudbees/jenkins/GitHubPushHookReceiverTest.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    private Run waitForBuild(int n, Job job) throws InterruptedException {
+        System.out.println("Waiting for build #" + n + " to appear and finish");
+        int counter = 0;
+        while (true) {
+            if(counter > 30) {
+                assertTrue("Waiting more than a minute for the build to start", false);
+            }
```
jglick: All `JenkinsRule` tests will time out after a few minutes anyway, so this is not actually necessary.

jglick:
:bee:, nice!

src/main/java/com/cloudbees/jenkins/Cleaner.java
```diff
@@ -52,10 +56,18 @@ protected void doRun() throws Exception {
         }
 
         // subtract all the live repositories
-        for (AbstractProject<?,?> job : Hudson.getInstance().getAllItems(AbstractProject.class)) {
-            GitHubPushTrigger trigger = job.getTrigger(GitHubPushTrigger.class);
-            if (trigger!=null) {
-                names.removeAll(GitHubRepositoryNameContributor.parseAssociatedNames(job));
+        Jenkins jenkins = Jenkins.getInstance();
+        if(jenkins != null) {
```
KostyaSha: again no space after if

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -114,7 +118,12 @@ public void run() {
                         LOGGER.log(Level.WARNING, "Failed to parse the polling log",e);
                         cause = new GitHubPushCause(pushBy);
                     }
-                    if (job.scheduleBuild(cause)) {
+                    ParameterizedJobMixIn scheduler = new ParameterizedJobMixIn() {
```
KostyaSha: why it called scheduler?

amuniz: Because the object is created **only** to schedule a build, so it's used as a **scheduler**. Anyway, I can change it if you think is going to be clearer. Any suggestion?

src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
amuniz: This would be a refactor that will make the merge after #57 even harder, I'd prefer to keep it as is now.

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
KostyaSha: This a Job object and you call schedule() on it. This is not a `scheduler` like you think.

KostyaSha: Also such naming logic is not used anywhere in jenkins code.

pom.xml
KostyaSha: @jtnord seems it related to your idea of shipping slf4j in core instead of log4j?

jtnord: the slf-api (and slf4j-jdk14) are already provided by jenkins core (since a long time ago), so this should probably be marked as `<scope>provided</scope>`

src/main/java/com/cloudbees/jenkins/Cleaner.java
jtnord: regardless is is better form to use parameterized logging - and the concatenation of the final empty string looks bogus!

src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
```diff
@@ -29,14 +33,40 @@
...
+        parseAssociatedNames((Job) job, result);
+    }
+
+    /**
+     * Looks at the definition of {@link Job} and list up the related github repositories,
+     * then puts them into the collection.
      */
-    public abstract void parseAssociatedNames(AbstractProject<?,?> job, Collection<GitHubRepositoryName> result);
+    public /*abstract*/ void parseAssociatedNames(Job<?,?> job, Collection<GitHubRepositoryName> result) {
+        if (Util.isOverridden(GitHubRepositoryNameContributor.class, getClass(),
```
jtnord: would this not have been so much easier to just throw an exception in `parseAssociatedNames((AbstractProject<?,?> job, Collection<GitHubRepositoryName> result))` - no one could have called `super.` on it as it was abstract.
then this is purely just needing to call that method and let subclassed overrride this.

src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
```diff
@@ -29,14 +33,40 @@
...
+     */
+    @Deprecated
+    public void parseAssociatedNames(AbstractProject<?,?> job, Collection<GitHubRepositoryName> result) {
+        parseAssociatedNames((Job) job, result);
+    }
+
+    /**
+     * Looks at the definition of {@link Job} and list up the related github repositories,
+     * then puts them into the collection.
      */
```
jtnord: this javadoc does not match at all what this method does.

src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
```diff
@@ -29,14 +33,40 @@
...
-    public abstract void parseAssociatedNames(AbstractProject<?,?> job, Collection<GitHubRepositoryName> result);
+    public /*abstract*/ void parseAssociatedNames(Job<?,?> job, Collection<GitHubRepositoryName> result) {
+        if (Util.isOverridden(GitHubRepositoryNameContributor.class, getClass(),
+                "parseAssociatedNames", AbstractProject.class, Collection.class) && job instanceof AbstractProject) {
+            parseAssociatedNames((AbstractProject) job, result);
```
jtnord: and what if this is a job?

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
amuniz: I don't think it's a ```scheduler```, I know what it is. I just said I'm using it to schedule a job.
I repeat, tell me your name proposal and I'll change it.

src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
amuniz: I just followed the same pattern used by others, for example here: https://github.com/jenkinsci/rebuild-plugin/blob/e4fdcee240b609fe6bbbcd58071951d6b38c9453/src/main/java/com/sonyericsson/rebuild/RebuildValidator.java#L46

src/test/java/com/cloudbees/jenkins/GitHubPushHookReceiverTest.java
```diff
@@ -0,0 +1,103 @@
...
+import org.junit.Assert;
+import org.junit.Rule;
+import org.junit.Test;
+import org.jvnet.hudson.test.JenkinsRule;
+
+/**
+ * Test case covering the code responsible to receive the push notification
+ * from GitHub.
+ */
+public class GitHubPushHookReceiverTest extends Assert {
```
lanwen: dont extend Assert class - it is junit3 style which was deprecated about a 10 years ago. Use hamcrest matchers instead of asserts

src/test/java/com/cloudbees/jenkins/GitHubPushHookReceiverTest.java
```diff
@@ -0,0 +1,103 @@
...
+        // Trigger the build once to register SCMs
+        WorkflowRun lastRun = j.assertBuildStatusSuccess(job.scheduleBuild2(0));
+        // Testing hack! This will make the polling believe that there was remote changes to build
+        lastRun.getActions(BuildData.class).get(0).buildsByBranchName = new HashMap<String, Build>();
+
+        // Then simulate a GitHub push
+        String payload = IOUtils.toString(getClass().getResourceAsStream("/com/cloudbees/jenkins/plugins/github/payload2.json"));
+        receiver.processGitHubPayload(payload, GitHubPushTrigger.class);
+        Run build = waitForBuild(2, job);
+        assertTrue("Build was triggered but did not success: " + build.getLog(300), Result.SUCCESS.equals(build.getResult()));
```
lanwen: please use assertThat + equalTo (or `j.assertBuildStatusSuccess(build).`) as more refactorable and human readable

src/test/java/com/cloudbees/jenkins/GitHubPushHookReceiverTest.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    private Run waitForBuild(int n, Job job) throws InterruptedException {
+        System.out.println("Waiting for build #" + n + " to appear and finish");
+        int counter = 0;
+        while (true) {
```
lanwen: `while(true)` is bad practice - why not just check counter in while loop and use `fail()` after that?

src/test/java/com/cloudbees/jenkins/GitHubPushHookReceiverTest.java
```diff
@@ -0,0 +1,103 @@
...
+            "    git 'https://github.com/amuniz/github-plugin.git'\n" +
+            "}"));
+
+        // Trigger the build once to register SCMs
+        WorkflowRun lastRun = j.assertBuildStatusSuccess(job.scheduleBuild2(0));
+        // Testing hack! This will make the polling believe that there was remote changes to build
+        lastRun.getActions(BuildData.class).get(0).buildsByBranchName = new HashMap<String, Build>();
+
+        // Then simulate a GitHub push
+        String payload = IOUtils.toString(getClass().getResourceAsStream("/com/cloudbees/jenkins/plugins/github/payload2.json"));
```
lanwen: this logic completely rewritten and simplified in #60. So this test can be simplified too after merging

src/test/java/com/cloudbees/jenkins/GitHubPushHookReceiverTest.java
```diff
@@ -0,0 +1,103 @@
...
+    @Test
+    public void receivePushHookOnWorkflow() throws Exception {
+        j.jenkins.getInjector().injectMembers(this);
+        WorkflowJob job = j.jenkins.createProject(WorkflowJob.class, "Test Workflow");
+
+        GitHubPushTrigger trigger = new GitHubPushTrigger();
+        trigger.start(job, false);
+        job.addTrigger(trigger);
+        job.setDefinition(new CpsFlowDefinition(
+            "node {\n" +
```
lanwen: please save dsl as test resource. This can be helpful to maintain it in future

src/test/java/com/cloudbees/jenkins/GitHubPushHookReceiverTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+    @Rule
+    public JenkinsRule j = new JenkinsRule();
+
+    @Inject
+    GitHubWebHook receiver;
+
+    @Test
+    public void receivePushHookOnFreeStyle() throws IOException, InterruptedException, ExecutionException {
+        j.jenkins.getInjector().injectMembers(this);
```
lanwen: do such things in rule (`ExternalResource`) for example. Don't mix preperation to test and testcase itself

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -114,7 +118,12 @@ public void run() {
...
+                        @Override protected Job asJob() {
```
lanwen: please split `@Override` annotation and method signature to two lines as more preferable formatting way

src/test/java/com/cloudbees/jenkins/GitHubPushHookReceiverTest.java
amuniz: The extend is only used to have a comfortable way to use ```assert*``` methods. Is something wrong in it?

src/test/java/com/cloudbees/jenkins/GitHubPushHookReceiverTest.java
amuniz: Not sure what you mean. Is it about using a ```@Before``` annotation in a separate method?

amuniz:
@lanwen your review comments have been addressed

src/test/java/com/cloudbees/jenkins/GitHubPushHookReceiverTest.java
jtnord: @amuniz import static is your friend.

pom.xml
jtnord: (see also jenkins#1765)

src/test/java/com/cloudbees/jenkins/GitHubPushHookReceiverTest.java
amuniz: @jtnord ok. Done.

src/test/java/com/cloudbees/jenkins/GitHubPushHookReceiverTest.java
lanwen: @amuniz yep, its the same, but more powerful
https://github.com/junit-team/junit/wiki/Rules

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
daniel-beck: `scheduledJob` would fit IMO. It doesn't schedule, it gets scheduled.

KostyaSha: There is a job that can schedule build, why do you need do it through parametrized?

amuniz: @daniel-beck ```scheduledJob``` works for me. I'll change it.
@KostyaSha The ```job``` variable is an instance of ```Job``` which doesn't have any of the ```schedule*``` methods. If you know a better way to schedule a build based on a ```Job``` instance, please tell me, I'll be happy to modify it.

amuniz:
@KostyaSha @lanwen I've merged from master and solved conflicts. An additional test that I added before has been removed since it's already covered by new tests introduced in master. 

Is something else required to have this merged in master?

src/main/java/org/jenkinsci/plugins/github/util/JobInfoHelpers.java
```diff
@@ -31,10 +34,20 @@ private JobInfoHelpers() {
...
      * @return predicate with true on apply if job contains trigger of given class
      */
-    public static Predicate<AbstractProject> withTrigger(final Class<? extends Trigger> clazz) {
-        return new Predicate<AbstractProject>() {
-            public boolean apply(AbstractProject job) {
-                return job.getTrigger(clazz) != null;
+    public static Predicate<Job> withTrigger(final Class<? extends Trigger> clazz) {
+        return new Predicate<Job>() {
+            public boolean apply(Job job) {
+                if (job instanceof ParameterizedJobMixIn.ParameterizedJob) {
```
lanwen: is `ParameterizedJobMixIn.ParameterizedJob` equals old `AbstractProject`?

src/main/java/org/jenkinsci/plugins/github/webhook/subscriber/DefaultPushGHEventSubscriber.java
```diff
@@ -82,8 +87,17 @@ protected void onEvent(GHEvent event, String payload) {
             ACL.impersonate(ACL.SYSTEM, new Runnable() {
                 @Override
                 public void run() {
-                    for (AbstractProject<?, ?> job : Jenkins.getInstance().getAllItems(AbstractProject.class)) {
-                        GitHubTrigger trigger = job.getTrigger(GitHubPushTrigger.class);
+                    for (Job<?, ?> job : Jenkins.getInstance().getAllItems(Job.class)) {
```
lanwen: its became very ugly

lanwen: is it the simplest way to get trigger from job?

lanwen:
Can you add to description of this PR example of use case?

src/main/java/org/jenkinsci/plugins/github/util/JobInfoHelpers.java
amuniz: How equals? Actually ```AbstractProject``` implements ```ParameterizedJob```, is that what you mean?

amuniz: If ```job``` is an instance of ```AbstractProject``` then ```job instanceof ParameterizedJobMixIn.ParameterizedJob``` will be ```true```, so the behavior will be the same than it was before this change.

Perhaps your question was more about this.

src/main/java/org/jenkinsci/plugins/github/webhook/subscriber/DefaultPushGHEventSubscriber.java
amuniz: Yes. I don't know any other way to do it (there is no ```getTrigger(Class)``` in ```Job```).

amuniz:
@lanwen done. But I'm not sure if it was what you wanted. There was an user raising an issue to have this working (GitHub Plugin + Workflow jobs): https://issues.jenkins-ci.org/browse/JENKINS-27136
And I think there are more users waiting for this.

s0undt3ch:
I'm sure the workflow plugin will get more attention as soon as this is in.

KostyaSha:
This PR will be merged when we resolved all planned changes on current core version, then core version will be bumped and PR will be merged.

src/main/java/org/jenkinsci/plugins/github/util/JobInfoHelpers.java
jglick: See [this document](https://github.com/jenkinsci/workflow-plugin/blob/master/COMPATIBILITY.md#historical-background).

amuniz:
@KostyaSha Is there a list of planned changes that I can check? Just to know how long is it and if I can help with.

src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
jglick: Note: #57 and #60 are now merged so this simplification is again an option.

pom.xml
jglick: @amuniz a tip: if you change this line to

```xml
<version>1.580.1</version> <!-- first LTS after SCMTriggerItem -->
```

then this whole comment thread becomes hidden by default. :-)

pom.xml
```diff
@@ -53,23 +52,10 @@
           <version>1.67</version>
       </dependency>
 
-      <dependency>
+      <dependency><!-- minimun required version to work with Workflow 1.4 GitSCM step-->
```
jglick: typo

amuniz:
@KostyaSha Moreover, I'd like to know how "big" that changes are. I hope you understand that keeping this PR mergeable (no conflicts) for a long time while receiving changes to master is a total pain, don't you agree?

src/main/java/com/cloudbees/jenkins/Cleaner.java
```diff
@@ -2,6 +2,7 @@
 
 import hudson.Extension;
 import hudson.model.AbstractProject;
```
jglick: BTW you could clean up this import (if you fixed the obsolete Javadoc link in line 32).

pom.xml
KostyaSha: @jglick :scream_cat:

KostyaSha:
> @KostyaSha Is there a list of planned changes that I can check? Just to know how long is it and if I can help with.

I described it twice in some of PRs/issues. Also spent 2 months requesting @oleg-nenashev help. Now @lanwen continue my work with fixing/extending GH-plugin. Sometimes work is blocked by requesting many months github-api releases.

> @KostyaSha Moreover, I'd like to know how "big" that changes are. I hope you understand that keeping this PR mergeable (no conflicts) for a long time while receiving changes to master is a total pain, don't you agree?

We will wrote you when time come. If you unable to do it we will do it ourself.

pom.xml
KostyaSha: facepalm

lanwen:
@amuniz I mean example of workflow definition or screenshot, or what else... why it needs to do such changes. Some kind of full explanation for dummies.

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -110,7 +116,13 @@ public void run() {
                         LOGGER.log(Level.WARNING, "Failed to parse the polling log",e);
                         cause = new GitHubPushCause(pushBy);
                     }
-                    if (job.scheduleBuild(cause)) {
+                    ParameterizedJobMixIn scheduledJob = new ParameterizedJobMixIn() {
+                        @Override
+                        protected Job asJob() {
+                            return job;
+                        }
+                    };
```
jglick: BTW I am working on a core PR to offer a nicer interface for this idiom.

amuniz:
@jglick You review comments were fixed. Thanks!

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
amuniz: +1 It'd be nice

amuniz:
@KostyaSha 

> I described it twice in some of PRs/issues.

Could you point me to it please? I can not find it anywhere.

> We will wrote you when time come. If you unable to do it we will do it ourself.

Sure, I'll be able. I just want to know how far we are from that moment and try to remove any impediment.

amuniz:
@lanwen I don't think I could explain it with screenshots. There is a lot information in the Workflow Plugin repository. About the interfacing changes, @jglick already pointed [to this documentation link](https://github.com/jenkinsci/workflow-plugin/blob/master/COMPATIBILITY.md#historical-background), which explains the hierarchy and interfaces changes.

src/main/java/org/jenkinsci/plugins/github/webhook/subscriber/DefaultPushGHEventSubscriber.java
jglick: [This idiom](https://github.com/jenkinsci/jenkins/blob/252e1296099e838b8fba52221910442e0a757db4/core/src/main/java/jenkins/triggers/ReverseBuildTrigger.java#L261-263) is shorter, but I agree it would be nice to have a single `getTrigger(Class)` for convenience.

jglick:
:bee:

lanwen:
@amuniz - ok, i've read this. Please add this to pr description too. It clear for me now

amuniz:
@lanwen great! Done.

amuniz:
@reviewbybees done

amuniz:
@lanwen @KostyaSha how far are we from merging this? Can I help somehow to speed up the process?

KostyaSha:
no, no eta

austinnichols101:
@amuniz - please add my name to the list of people waiting for this update.

src/main/java/org/jenkinsci/plugins/github/webhook/subscriber/DefaultPushGHEventSubscriber.java
```diff
@@ -82,8 +87,17 @@ protected void onEvent(GHEvent event, String payload) {
...
+                        GitHubTrigger trigger = null;
+                        if (job instanceof ParameterizedJobMixIn.ParameterizedJob) {
+                            ParameterizedJobMixIn.ParameterizedJob pJob = (ParameterizedJobMixIn.ParameterizedJob) job;
+                            for (Trigger candidate : pJob.getTriggers().values()) {
+                                if (candidate instanceof GitHubTrigger) {
+                                    trigger = (GitHubTrigger) candidate;
```
jglick:
```java
// TODO use standard method in 1.621+
```

src/test/java/org/jenkinsci/plugins/github/webhook/subscriber/WebhookWorkflow.java
```diff
@@ -0,0 +1,65 @@
...
+
+        // Trigger the build once to register SCMs
+        WorkflowRun lastRun = j.assertBuildStatusSuccess(job.scheduleBuild2(0));
+        // Testing hack! This will make the polling believe that there was remote changes to build
+        lastRun.getActions(BuildData.class).get(0).buildsByBranchName = new HashMap<String, Build>();
+
+        // Then simulate a GitHub push
+        new DefaultPushGHEventSubscriber()
+                .onEvent(GHEvent.PUSH, GitHubWebHookFullTest.classpath("payloads/push-wf.json"));
+        Run build = waitForBuild(2, job);
```
jglick: `JenkinsRule.waitUntilNoActivity` would be easier. That is how I normally test triggers. [example](https://github.com/jenkinsci/workflow-plugin/blob/7fd6727c4931fddd0e6722d5bdd4c813fe333008/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/scm/GitSampleRepoRule.java#L46-L50)

amuniz:
Merge conflicts solved. This PR is mergeable again.
@jglick your comments has been addressed.

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
jglick: This also is in 1.621 BTW.

src/main/java/com/cloudbees/jenkins/GitHubTrigger.java
```diff
@@ -44,9 +47,12 @@
     @Extension
     public static class GitHubRepositoryNameContributorImpl extends GitHubRepositoryNameContributor {
         @Override
-        public void parseAssociatedNames(AbstractProject<?, ?> job, Collection<GitHubRepositoryName> result) {
-            for (GitHubTrigger ght : Util.filter(job.getTriggers().values(),GitHubTrigger.class)) {
-                result.addAll(ght.getGitHubRepositories());
+        public void parseAssociatedNames(Job<?, ?> job, Collection<GitHubRepositoryName> result) {
+            if (job instanceof ParameterizedJobMixIn.ParameterizedJob) {
+                ParameterizedJobMixIn.ParameterizedJob p = (ParameterizedJobMixIn.ParameterizedJob) job;
+                for (GitHubTrigger ght : Util.filter(p.getTriggers().values(),GitHubTrigger.class)) {
```
jglick: Another place where the 1.621 method would be nicer.

src/main/java/org/jenkinsci/plugins/github/util/JobInfoHelpers.java
```diff
@@ -31,10 +34,20 @@ private JobInfoHelpers() {
...
+                    ParameterizedJobMixIn.ParameterizedJob pJob = (ParameterizedJobMixIn.ParameterizedJob) job;
+                    for (Trigger trigger : pJob.getTriggers().values()) {
+                        if (trigger instanceof GitHubPushTrigger) {
```
jglick: And another.

src/test/java/org/jenkinsci/plugins/github/webhook/subscriber/WebhookWorkflow.java
```diff
@@ -0,0 +1,47 @@
...
+import com.cloudbees.jenkins.GitHubWebHookFullTest;
+
+public class WebhookWorkflow {
+
+    @Rule
+    public JenkinsRule j = new JenkinsRule();
+
+    @Test
+    public void receivePushHookOnWorkflow() throws Exception {
+        j.jenkins.getInjector().injectMembers(this);
```
jglick: pointless

jglick:
:bee:

jglick:
:bee:

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -182,7 +194,7 @@ public DescriptorImpl getDescriptor() {
      * Action object for {@link Project}. Used to display the polling log.
      */
     public final class GitHubWebHookPollingAction implements Action {
-        public AbstractProject<?, ?> getOwner() {
+        public Job<?, ?> getOwner() {
```
oleg-nenashev: Seems we need the <code>withBridgeMethods</code> magic here as well

oleg-nenashev:
@amuniz 
Did you agree on the binary compatibility impact with @lanwen and @KostyaSha . There're some issues in the PR, but IIRC there was a formal breakage in the previous release. Probably the compatibility loss is OK now.

lanwen:
i'm ok to merge it. But @KostyaSha thinks that bumping core version before all other useful PRs will be closed on current core version may cause this plugin unusable for older jenkins instances.

jglick:
If the plugin maintainer desires commit squashing I would suggest that be done as part of the merge rather than destroying history in the PR itself.

KostyaSha:
@jglick i can suggest you firstly work with recena and amuniz PRs in CB forks and only after review send to maintainers. Never saw in jenkinsci PRs with so many commits and issues before. Back to upstream repos when their PRs wouldn't have "child" issues.
I'm always squashing my commits and i have no time for doing it instead of your full time employees.

@oleg-nenashev could you say what exactly was broken for you in last releases?

amuniz:
@KostyaSha just ask for squash before you are going to merge, I'll do it.

The current amount of revisions and comments in every PR is hopefully going to significantly increase the code base quality in the Jenkins community in general. Moreover, I think is quite good for other developers to have all this review information available publicly, since it is, from my point if view, a nice source of knowledge.

KostyaSha:
@amuniz CB pays you for your work, but it doesn't mean that CB that hired and locked all possible jenkins related devs must lock jenkins activities on their internal plans/needs. This noise is waste of time for FOSS devs, instead of working on new features and other plugins i'm getting 100 comments with newbie errors. Compare to @jglick PRs that doesn't require many discussions. 

I will back to this PR when we resolve other planned features/fixes and will be ok with core bump.

src/main/java/org/jenkinsci/plugins/github/webhook/WebhookManager.java
```diff
@@ -73,7 +76,7 @@ public static WebhookManager forHookUrl(URL endpoint) {
      * @return runnable to create hooks on run
      * @see #createHookSubscribedTo(List)
      */
-    public Runnable registerFor(final AbstractProject<?, ?> project) {
+    public Runnable registerFor(final Job<?, ?> project) {
```
daniel-beck: This change makes the Javadoc comment in line 133 obsolete, and without that, AbstractProject is an unused import as well.

amuniz: @daniel-beck thanks. Fixed.

amuniz:
Merge conflicts resolved.

lanwen:
@amuniz please don't rebase. We have 3 or 4 prs in "must have" queue before we are going to merge this pr. I'll do the rebase by myself when the time will come

amuniz:
@lanwen ok. If you face some issue while merging, ping me.

amuniz:
@lanwen is there any pending PR blocking this to merge yet?

KostyaSha:
@amuniz core update blocks it.

amuniz:
@KostyaSha could you elaborate please? What's wrong with the core update?

KostyaSha:
1.554.1 (2014/04/30)
1.565.3 (2014/10/01)
1.580.1 (2014/10/29)
1) So soon it will be possible to bump new version
2) Not all planned features are done, if we will change something after, you will jump with "backward compatibility is broken" so better wait. I don't remember exact plan, but `notifyCommit` and generic push trigger is one of.

lanwen:
no, think generic trigger is too complex to move workflow in queue. 

And while git-plugin of new version is already depends on core version with workflow think we should try to merge this.

KostyaSha:
ok, as you wish, i tired from this workflow spam

lanwen:
please review #82

lanwen:
Merged in #82

PR #61, opened 2015-07-09
Make the Github property work on all Jobs

emanuelez:
Don’t assume AbstractProject. This should allow it to work with the workflow plugin.


emanuelez:
My bad... I should have notices there's already activity on this front :)

PR #63, opened 2015-07-16
[WiP] Enable FindBugs in the plugin

oleg-nenashev:
- [x] Enable FindBugs and require its runs on the verify stages
- [x] Some code polishing

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -7,7 +7,7 @@
       <!--JENKINS-18537-->
     <version>1.554.1</version>
   </parent>
-
+  
   <groupId>com.coravy.hudson.plugins.github</groupId>
   <artifactId>github</artifactId>
   <packaging>hpi</packaging>
@@ -164,6 +164,9 @@
     <properties>
         <maven-compiler-plugin.version>3.3</maven-compiler-plugin.version>
         <maven-release-plugin.version>2.5.1</maven-release-plugin.version>
+        <findbugs-maven-plugin.version>3.0.1</findbugs-maven-plugin.version>
+        <!--TODO: replace by true when it's clean-->
+        <findbugs.failOnError>false</findbugs.failOnError>
     </properties>
     
   <build>
@@ -178,11 +181,11 @@
         <plugin>
           <groupId>org.codehaus.mojo</groupId>
           <artifactId>findbugs-maven-plugin</artifactId>
-          <version>3.0.1</version>
+          <version>${findbugs-maven-plugin.version}</version>
           <configuration>
             <xmlOutput>true</xmlOutput>
             <findbugsXmlWithMessages>true</findbugsXmlWithMessages>
-            <failOnError>false</failOnError>
+            <failOnError>${findbugs.failOnError}</failOnError>
           </configuration>
           <executions>
             <execution>
--- a/pom.xml
+++ b/pom.xml
@@ -78,6 +78,13 @@
           <version>1.3</version>
           <scope>provided</scope>
       </dependency>
+      
+      <dependency>
+        <groupId>com.google.code.findbugs</groupId>
+        <artifactId>annotations</artifactId>
+        <version>3.0.0</version>
+        <scope>provided</scope>
+      </dependency>
 
       <!--TEST DEPS-->
       <dependency>
@@ -168,7 +175,26 @@
                 <target>1.6</target>
             </configuration>
         </plugin>
-    </plugins>  
+        <plugin>
+          <groupId>org.codehaus.mojo</groupId>
+          <artifactId>findbugs-maven-plugin</artifactId>
+          <version>3.0.1</version>
+          <configuration>
+            <xmlOutput>true</xmlOutput>
+            <findbugsXmlWithMessages>true</findbugsXmlWithMessages>
+            <failOnError>false</failOnError>
+          </configuration>
+          <executions>
+            <execution>
+              <id>run-findbugs</id>
+              <phase>verify</phase> 
+              <goals>
+                <goal>check</goal> 
+              </goals>
+            </execution>
+          </executions>
+        </plugin>
+    </plugins>    
   </build>
     
 </project>  
--- a/src/main/java/com/cloudbees/jenkins/GitHubPushCause.java
+++ b/src/main/java/com/cloudbees/jenkins/GitHubPushCause.java
@@ -3,6 +3,7 @@ package com.cloudbees.jenkins;
 import hudson.triggers.SCMTrigger.SCMTriggerCause;
 import java.io.File;
 import java.io.IOException;
+import java.util.Objects;
 
 /**
  * UI object that says a build is started by GitHub post-commit hook.
@@ -33,5 +34,22 @@ public class GitHubPushCause extends SCMTriggerCause {
     public String getShortDescription() {
         String pusher = pushedBy != null ? pushedBy : "";
         return "Started by GitHub push by " + pusher;
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hashCode(this.pushedBy);
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (!getClass().isInstance(obj)) {
+            return false;
+        }
+        final GitHubPushCause other = (GitHubPushCause) obj;
+        if ((this.pushedBy == null) ? (other.pushedBy != null) : !this.pushedBy.equals(other.pushedBy)) {
+            return false;
+        }
+        return true;
     }
 }
--- a/src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
+++ b/src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
@@ -2,6 +2,7 @@ package com.cloudbees.jenkins;
 
 import com.google.common.base.Charsets;
 import com.google.common.base.Function;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.Extension;
 import hudson.Util;
 import hudson.console.AnnotatedLargeText;
@@ -29,6 +30,7 @@ import javax.inject.Inject;
 import java.io.File;
 import java.io.IOException;
 import java.io.PrintStream;
+import java.io.UnsupportedEncodingException;
 import java.net.HttpURLConnection;
 import java.net.MalformedURLException;
 import java.net.URL;
@@ -299,13 +301,17 @@ public class GitHubPushTrigger extends Trigger<AbstractProject<?, ?>> implements
                     return FormValidation.warning("It doesn't look like " + value + " is talking to any Jenkins. Are you running your own app?");
                 }
                 RSAPublicKey key = identity.getPublic();
-                String expected = new String(Base64.encodeBase64(key.getEncoded()));
+                String expected = new String(Base64.encodeBase64(key.getEncoded()), "UTF-8" );
                 if (!expected.equals(v)) {
                     // if it responds but with a different ID, that's more likely wrong than correct
                     return FormValidation.error(value+" is connecting to different Jenkins instances");
                 }
 
                 return FormValidation.ok();
+            } catch (UnsupportedEncodingException ex) {
+                // Should not happen
+                // TODO: get rid of it after migrating to JDK7
+                return FormValidation.error("Unsupported exception {0}", ex.getLocalizedMessage());
             } catch (IOException e) {
                 return FormValidation.error(e,"Failed to test a connection to "+value);
             }
@@ -336,6 +342,8 @@ public class GitHubPushTrigger extends Trigger<AbstractProject<?, ?>> implements
     /**
      * Set to false to prevent the user from overriding the hook URL.
      */
+    @SuppressFBWarnings(value = "MS_SHOULD_BE_FINAL", 
+            justification = "Common Jenkins config pattern, which may be changed in the runtime")
     public static boolean ALLOW_HOOKURL_OVERRIDE = !Boolean.getBoolean(GitHubPushTrigger.class.getName() + ".disableOverride");
 
     private static final Logger LOGGER = Logger.getLogger(GitHubPushTrigger.class.getName());
--- a/src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
+++ b/src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
@@ -1,5 +1,6 @@
 package com.cloudbees.jenkins;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.EnvVars;
 import hudson.Extension;
 import hudson.ExtensionList;
@@ -18,6 +19,8 @@ import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.logging.Level;
+import org.slf4j.Logger;
 
 /**
  * Extension point that associates {@link GitHubRepositoryName}s to a project.
@@ -26,6 +29,9 @@ import java.util.Set;
  * @since 1.7
  */
 public abstract class GitHubRepositoryNameContributor implements ExtensionPoint {
+    
+    public static final Logger LOGGER = org.slf4j.LoggerFactory.getLogger(GitHubRepositoryNameContributor.class);
+    
     /**
      * Looks at the definition of {@link AbstractProject} and list up the related github repositories,
      * then puts them into the collection.
@@ -51,8 +57,8 @@ public abstract class GitHubRepositoryNameContributor implements ExtensionPoint 
             for (EnvironmentContributor contributor : EnvironmentContributor.all()) {
                 try {
                     contributor.buildEnvironmentFor(job, env, TaskListener.NULL);
-                } catch (Exception e) {
-                    // ignore
+                } catch (Throwable e) {
+                    LOGGER.warn("Unhandled exception. Skipping environment contributor " + contributor, e);
                 }
             }
             return env;
@@ -92,8 +98,13 @@ public abstract class GitHubRepositoryNameContributor implements ExtensionPoint 
     @Extension(optional=true)
     @SuppressWarnings("unused")
     public static class FromMultiSCM extends AbstractFromSCMImpl {
-        // make this class fail to load if MultiSCM is not present
-        public FromMultiSCM() { MultiSCM.class.toString(); }
+        
+        @SuppressFBWarnings(value = "RV_RETURN_VALUE_IGNORED_NO_SIDE_EFFECT", 
+                justification = "As designed, error spot-check")
+        public FromMultiSCM() {
+            // make this class fail to load if MultiSCM is not present
+            MultiSCM.class.toString(); 
+        }
 
         @Override
         public void parseAssociatedNames(AbstractProject<?, ?> job, Collection<GitHubRepositoryName> result) {
--- a/src/main/java/org/jenkinsci/plugins/github/extension/GHEventsSubscriber.java
+++ b/src/main/java/org/jenkinsci/plugins/github/extension/GHEventsSubscriber.java
@@ -15,7 +15,7 @@ import java.util.Set;
 
 import static java.util.Collections.emptySet;
 import static org.apache.commons.lang3.ObjectUtils.defaultIfNull;
-import org.apache.commons.lang3.Validate;
+import static org.apache.commons.lang3.Validate.notNull;
 
 /**
  * Extension point to subscribe events from GH, which plugin interested in.
@@ -77,8 +77,7 @@ public abstract class GHEventsSubscriber implements ExtensionPoint {
         return new Function<GHEventsSubscriber, Set<GHEvent>>() {
             @Override
             public Set<GHEvent> apply(GHEventsSubscriber subscriber) {
-                Validate.notNull(subscriber);
-                return defaultIfNull(subscriber.events(), Collections.<GHEvent>emptySet());
+                return defaultIfNull(notNull(subscriber).events(), Collections.<GHEvent>emptySet());
             }
         };
     }
@@ -94,8 +93,7 @@ public abstract class GHEventsSubscriber implements ExtensionPoint {
         return new Predicate<GHEventsSubscriber>() {
             @Override
             public boolean apply(GHEventsSubscriber subscriber) {
-                Validate.notNull(subscriber);
-                return subscriber.isApplicable(project);
+                return notNull(subscriber).isApplicable(project);
             }
         };
     }
@@ -111,8 +109,7 @@ public abstract class GHEventsSubscriber implements ExtensionPoint {
         return new Predicate<GHEventsSubscriber>() {
             @Override
             public boolean apply(GHEventsSubscriber subscriber) {
-                Validate.notNull(subscriber);
-                return defaultIfNull(subscriber.events(), emptySet()).contains(event);
+                return defaultIfNull(notNull(subscriber).events(), emptySet()).contains(event);
             }
         };
     }
@@ -129,9 +126,10 @@ public abstract class GHEventsSubscriber implements ExtensionPoint {
         return new Function<GHEventsSubscriber, Void>() {
             @Override
             public Void apply(GHEventsSubscriber subscriber) {
-                Validate.notNull(subscriber);
                 try {
-                    subscriber.onEvent(event, payload);
+                    notNull(subscriber).onEvent(event, payload);
+                } catch (Error error) {
+                    throw error;
                 } catch (Throwable t) {
                     LOGGER.error("Subscriber {} failed to process {} hook, skipping...",
                            subscriber.getClass().getName(), event, t);
--- a/src/main/java/org/jenkinsci/plugins/github/extension/GHEventsSubscriber.java
+++ b/src/main/java/org/jenkinsci/plugins/github/extension/GHEventsSubscriber.java
@@ -136,6 +136,7 @@ public abstract class GHEventsSubscriber implements ExtensionPoint {
                     LOGGER.error("Subscriber {} failed to process {} hook, skipping...",
                            subscriber.getClass().getName(), event, t);
                 }
+                return null;
             }
         };
     }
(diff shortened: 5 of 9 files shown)


oleg-nenashev:
@KostyaSha @lanwen @reviewbybees
Ready for the review

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -299,13 +301,17 @@ public FormValidation doCheckHookUrl(@QueryParameter String value) {
...
                 if (!expected.equals(v)) {
                     // if it responds but with a different ID, that's more likely wrong than correct
                     return FormValidation.error(value+" is connecting to different Jenkins instances");
                 }
 
                 return FormValidation.ok();
+            } catch (UnsupportedEncodingException ex) {
+                // Should not happen
+                // TODO: get rid of it after migrating to JDK7
+                return FormValidation.error("Unsupported exception "+ex.getLocalizedMessage());
```
KostyaSha: space around +

src/main/java/org/jenkinsci/plugins/github/webhook/RequirePostWithGHHookPayload.java
```diff
@@ -84,7 +85,11 @@ public void generateResponse(StaplerRequest req, StaplerResponse rsp, Object nod
                             throws IOException, ServletException {
                         RSAPublicKey key = new InstanceIdentity().getPublic();
                         rsp.setStatus(HttpServletResponse.SC_OK);
-                        rsp.setHeader(GitHubWebHook.X_INSTANCE_IDENTITY, new String(encodeBase64(key.getEncoded())));
+                        try {
+                            rsp.setHeader(GitHubWebHook.X_INSTANCE_IDENTITY, new String(encodeBase64(key.getEncoded()), "UTF-8"));
+                        } catch(UnsupportedEncodingException ex) {
```
KostyaSha: space

oleg-nenashev: where?

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
lanwen: FormValidation supports string formatting:
```java
FormValidation error(String format, Object... args) {
        return error(String.format(format,args));
    }
```

src/main/java/org/jenkinsci/plugins/github/webhook/RequirePostWithGHHookPayload.java
KostyaSha: is missing after catch

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -299,13 +301,17 @@ public FormValidation doCheckHookUrl(@QueryParameter String value) {
                     return FormValidation.warning("It doesn't look like " + value + " is talking to any Jenkins. Are you running your own app?");
                 }
                 RSAPublicKey key = identity.getPublic();
-                String expected = new String(Base64.encodeBase64(key.getEncoded()));
+                String expected = new String(Base64.encodeBase64(key.getEncoded()), "UTF-8" );
```
lanwen: use `com.google.common.base.Charsets.UTF_8` - it don't throw exc

src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
```diff
@@ -26,6 +29,9 @@
  * @since 1.7
  */
 public abstract class GitHubRepositoryNameContributor implements ExtensionPoint {
+    
+    public static final Logger LOGGER = Logger.getLogger(GitHubRepositoryNameContributor.class.getName());
```
lanwen: use slf4j logger in new places

src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
```diff
@@ -51,8 +57,8 @@ protected EnvVars buildEnv(AbstractProject<?, ?> job) {
             for (EnvironmentContributor contributor : EnvironmentContributor.all()) {
                 try {
                     contributor.buildEnvironmentFor(job, env, TaskListener.NULL);
-                } catch (Exception e) {
-                    // ignore
+                } catch (Throwable e) {
+                    LOGGER.log(Level.WARNING, "Unhandled exception. Skipping environment contributor" + contributor, e);
```
lanwen: please use templating for logging - it simplifies reading

src/main/java/org/jenkinsci/plugins/github/extension/GHEventsSubscriber.java
```diff
@@ -73,6 +73,9 @@ protected void onEvent(GHEvent event, String payload) {
         return new Function<GHEventsSubscriber, Set<GHEvent>>() {
             @Override
             public Set<GHEvent> apply(GHEventsSubscriber subscriber) {
+                if (subscriber == null) {
```
lanwen: this never should happen. Use Validate.notNull() from lang3 - it should throw exc and fixed if that happens

src/main/java/org/jenkinsci/plugins/github/webhook/WebhookManager.java
```diff
@@ -291,6 +310,9 @@ public GHHook apply(GHRepository repo) {
     protected Predicate<GHHook> deleteWebhook() {
         return new Predicate<GHHook>() {
             public boolean apply(GHHook hook) {
+                if (hook ==null) {
```
lanwen: i can get null here instead of hook if only myself call this method with null. And this is impossible, because of hooks deleted only after filtering

src/main/java/com/cloudbees/jenkins/GitHubPushCause.java
```diff
@@ -34,4 +34,26 @@ public String getShortDescription() {
         String pusher = pushedBy != null ? pushedBy : "";
         return "Started by GitHub push by " + pusher;
     }
+
+    @Override
+    public int hashCode() {
+        int hash = 5;
+        hash = 79 * hash + (this.pushedBy != null ? this.pushedBy.hashCode() : 0);
```
KostyaSha: what IDE generated this?

src/main/java/org/jenkinsci/plugins/github/webhook/WebhookManager.java
```diff
@@ -242,7 +254,11 @@ public boolean apply(GHHook hook) {
         return new Function<GHHook, Iterable<GHEvent>>() {
             @Override
             public Iterable<GHEvent> apply(GHHook input) {
-                return input.getEvents();
```
lanwen: you add this check already in the GHEventSubscriber

src/main/java/org/jenkinsci/plugins/github/webhook/WebhookManager.java
```diff
@@ -24,6 +24,7 @@
 import static com.google.common.base.Predicates.notNull;
 import static com.google.common.base.Predicates.or;
 import static java.lang.String.format;
+import java.util.Collections;
```
KostyaSha: how you inserted it between static block?)

src/main/java/org/jenkinsci/plugins/github/webhook/WebhookManager.java
```diff
@@ -183,7 +187,9 @@ public GHHook apply(GitHubRepositoryName name) {
         return new Predicate<GHHook>() {
             @Override
             public boolean apply(GHHook input) {
-                LOGGER.debug(format("%s {} (events: {})", format), input.getUrl(), input.getEvents());
+                if (input != null) {
```
lanwen: this method is private. No any can call it with null. And it cannot be null because it gets hook from previous step

src/main/java/org/jenkinsci/plugins/github/webhook/RequirePostWithGHHookPayload.java
lanwen: just use Charsets.UTF_8 and you shouldn't catch this exc

src/main/java/com/cloudbees/jenkins/GitHubPushCause.java
oleg-nenashev: Netbeans

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
oleg-nenashev: Formally we still need to keep the code compilable on JDK6. See my comment with TODO below

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
oleg-nenashev: I'll add spaces. Formatting is an overkill for a single string addition to the end

lanwen:
I think all null checks is redundant. Its like a shoot in the leg. It never should happen.
You can wrap it with Validate.notNull() from lang3 for public methods to protect from groovy scripts. But it shouldn't handle nulls quietly

src/main/java/org/jenkinsci/plugins/github/extension/GHEventsSubscriber.java
oleg-nenashev: It may happen - see Guava Javadocs. API explicitly allows to pass nulls. If it is not supported, NPE should be thrown

src/main/java/org/jenkinsci/plugins/github/webhook/RequirePostWithGHHookPayload.java
oleg-nenashev: Not in JDK6, see my comment above

src/main/java/org/jenkinsci/plugins/github/webhook/WebhookManager.java
oleg-nenashev: IDE automatically inserted it

src/main/java/com/cloudbees/jenkins/GitHubPushCause.java
```diff
@@ -34,4 +34,26 @@ public String getShortDescription() {
         String pusher = pushedBy != null ? pushedBy : "";
         return "Started by GitHub push by " + pusher;
     }
+
+    @Override
+    public int hashCode() {
```
lanwen: you can use `return Objects.hashCode(pushedBy)`

src/main/java/com/cloudbees/jenkins/GitHubPushCause.java
```diff
@@ -34,4 +34,26 @@ public String getShortDescription() {
...
+
+    @Override
+    public boolean equals(Object obj) {
+        if (obj == null) {
+            return false;
+        }
+        if (getClass() != obj.getClass()) {
+            return false;
+        }
+        final GitHubPushCause other = (GitHubPushCause) obj;
```
lanwen: `getClass().isInstance(obj)`
instead of
```
if (obj == null) {
+            return false;
+        }
+        if (getClass() != obj.getClass()) {
+            return false;
+        }

```

oleg-nenashev:
> I think all null checks is redundant. Its like a shoot in the leg. It never should happen.
You can wrap it with Validate.notNull() from lang3 for public methods to protect from groovy scripts. But it shouldn't handle nulls quietly. All other null checks should be ignored

OK, I'll follow this approach

src/main/java/com/cloudbees/jenkins/GitHubPushCause.java
oleg-nenashev: done

oleg-nenashev:
@KostyaSha @lanwen 
Fixed your comments in the code

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
lanwen: it compiled successfully with such const. I've already use it in some places without changing target source and target lang level

src/main/java/org/jenkinsci/plugins/github/extension/GHEventsSubscriber.java
```diff
@@ -92,6 +94,7 @@ protected void onEvent(GHEvent event, String payload) {
         return new Predicate<GHEventsSubscriber>() {
             @Override
             public boolean apply(GHEventsSubscriber subscriber) {
+                Validate.notNull(subscriber);
                 return subscriber.isApplicable(project);
```
lanwen: You can write it directly `Validate.notNull(subscriber).isApplicable(project);`

src/main/java/org/jenkinsci/plugins/github/extension/GHEventsSubscriber.java
```diff
@@ -125,13 +129,13 @@ public boolean apply(GHEventsSubscriber subscriber) {
...
+                Validate.notNull(subscriber);
                 try {
                     subscriber.onEvent(event, payload);
                 } catch (Throwable t) {
-                    LOGGER.error("Subscriber {} failed to process {} hook, skipping...", 
-                            subscriber.getClass().getName(), event, t);
+                    LOGGER.error("Subscriber {} failed to process {} hook, skipping...",
+                           subscriber.getClass().getName(), event, t);
                 }
-                return null;
```
lanwen: why?

src/main/java/org/jenkinsci/plugins/github/webhook/WebhookManager.java
```diff
@@ -137,7 +138,8 @@ public void unregisterFor(GitHubRepositoryName name, List<GitHubRepositoryName>
         return new Function<GitHubRepositoryName, GHHook>() {
             @Override
             public GHHook apply(GitHubRepositoryName name) {
-                try {
+                Validate.notNull(name);
```
lanwen: please use static imports

src/main/java/org/jenkinsci/plugins/github/webhook/WebhookManager.java
```diff
@@ -183,7 +185,8 @@ public GHHook apply(GitHubRepositoryName name) {
         return new Predicate<GHHook>() {
             @Override
             public boolean apply(GHHook input) {
-                LOGGER.debug(format("%s {} (events: {})", format), input.getUrl(), input.getEvents());
+                final GHHook _input = Validate.notNull(input);
```
lanwen: please don't use underline in var names. What for you reassign this to final var? notNull returns input without any change. And nothing can change input to make it final. Don't fix all FB warnings.

src/main/java/com/cloudbees/jenkins/GitHubPushCause.java
```diff
@@ -34,4 +35,21 @@ public String getShortDescription() {
...
     }
+
+    @Override
+    public int hashCode() {
+        return Objects.hashCode(this.pushedBy);
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (!getClass().isInstance(obj)) {
```
jtnord: `GitHubPushCause` is not final and this method is not reflective and could incorrectly match any subclass of this type who adds new fields and overrideds `equals()`
so `a.equals(b) != b.equals(a)`

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -336,6 +342,8 @@ public static boolean allowsHookUrlOverride() {
     /**
      * Set to false to prevent the user from overriding the hook URL.
      */
+    @SuppressFBWarnings(value = "MS_SHOULD_BE_FINAL", 
+            justification = "Common Jenkins config pattern, which may be changed in the runtime")
     public static boolean ALLOW_HOOKURL_OVERRIDE = !Boolean.getBoolean(GitHubPushTrigger.class.getName() + ".disableOverride");
```
jtnord: groovy scripts can change private, and so this field could be marked as private.

src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
```diff
@@ -26,6 +29,9 @@
  * @since 1.7
  */
 public abstract class GitHubRepositoryNameContributor implements ExtensionPoint {
+    
+    public static final Logger LOGGER = org.slf4j.LoggerFactory.getLogger(GitHubRepositoryNameContributor.class);
```
jtnord: public??

src/main/java/org/jenkinsci/plugins/github/extension/GHEventsSubscriber.java
```diff
@@ -76,7 +77,7 @@ protected void onEvent(GHEvent event, String payload) {
         return new Function<GHEventsSubscriber, Set<GHEvent>>() {
             @Override
             public Set<GHEvent> apply(GHEventsSubscriber subscriber) {
-                return defaultIfNull(subscriber.events(), Collections.<GHEvent>emptySet());
+                return defaultIfNull(notNull(subscriber).events(), Collections.<GHEvent>emptySet());
```
jtnord: what am I missing here?
[`notNull(Object)`](https://commons.apache.org/proper/commons-lang/javadocs/api-3.1/org/apache/commons/lang3/Validate.html#notNull(T)) will throw a `NullPointerException` if the Object is `null` - so no bug has been fixed here - just hidden under the carpet.

src/main/java/org/jenkinsci/plugins/github/extension/GHEventsSubscriber.java
```diff
@@ -92,7 +93,7 @@ protected void onEvent(GHEvent event, String payload) {
         return new Predicate<GHEventsSubscriber>() {
             @Override
             public boolean apply(GHEventsSubscriber subscriber) {
-                return subscriber.isApplicable(project);
+                return notNull(subscriber).isApplicable(project);
```
jtnord: hidden `NullPointerException`

src/main/java/org/jenkinsci/plugins/github/webhook/RequirePostWithGHHookPayload.java
```diff
@@ -84,7 +85,11 @@ public void generateResponse(StaplerRequest req, StaplerResponse rsp, Object nod
                             throws IOException, ServletException {
                         RSAPublicKey key = new InstanceIdentity().getPublic();
                         rsp.setStatus(HttpServletResponse.SC_OK);
-                        rsp.setHeader(GitHubWebHook.X_INSTANCE_IDENTITY, new String(encodeBase64(key.getEncoded())));
+                        try {
+                            rsp.setHeader(GitHubWebHook.X_INSTANCE_IDENTITY, new String(encodeBase64(key.getEncoded()), "UTF-8"));
+                        } catch (UnsupportedEncodingException ex) {
+                            throw new IOException("Default UTF-8 encoding is not available", ex);
```
jtnord: this is not an IOException - it is a fatal JVM issue as UTF-8 is mandated by the JLS.

src/main/java/org/jenkinsci/plugins/github/extension/GHEventsSubscriber.java
oleg-nenashev: We discussed it with @lanwen and @KostyaSha . Such behavior actually follows the documentation of Guava, which suggests to throw NPEs if null values is not supported. I've originally proposed an NPE-safe implementation for all methods, but got a push-back from the reviewers

src/main/java/org/jenkinsci/plugins/github/webhook/WebhookManager.java
oleg-nenashev: > What for you reassign this to final var? notNull returns input without any change

_input becomes NonNull

src/main/java/org/jenkinsci/plugins/github/extension/GHEventsSubscriber.java
```diff
@@ -126,10 +127,12 @@ public boolean apply(GHEventsSubscriber subscriber) {
...
                 try {
-                    subscriber.onEvent(event, payload);
+                    notNull(subscriber).onEvent(event, payload);
+                } catch (Error error) {
+                    throw error;
                 } catch (Throwable t) {
-                    LOGGER.error("Subscriber {} failed to process {} hook, skipping...", 
-                            subscriber.getClass().getName(), event, t);
+                    LOGGER.error("Subscriber {} failed to process {} hook, skipping...",
+                           subscriber.getClass().getName(), event, t);
```
jtnord: This can cause yet another `NullPointerException` as subscriber is possibly `null` if `t` is a `NullPointerException` and in this case nothing will be logged :-(

src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
```diff
@@ -18,6 +19,8 @@
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.logging.Level;
+import org.slf4j.Logger;
```
jtnord: you are doing something very wrong if you have both slf4j and java util logging in the same class....

src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
```diff
@@ -51,8 +57,8 @@ protected EnvVars buildEnv(AbstractProject<?, ?> job) {
             for (EnvironmentContributor contributor : EnvironmentContributor.all()) {
                 try {
                     contributor.buildEnvironmentFor(job, env, TaskListener.NULL);
-                } catch (Exception e) {
-                    // ignore
+                } catch (Throwable e) {
+                    LOGGER.warn("Unhandled exception. Skipping environment contributor " + contributor, e);
```
jtnord: this is slf4j so used full parameterized logging as it doesn;t have the same limitaion as JUL with parameterized and throwables.

However I think you actually mean this to be JUL.

src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
```diff
@@ -92,8 +98,13 @@ public void parseAssociatedNames(AbstractProject<?, ?> job, Collection<GitHubRep
     @Extension(optional=true)
     @SuppressWarnings("unused")
     public static class FromMultiSCM extends AbstractFromSCMImpl {
-        // make this class fail to load if MultiSCM is not present
-        public FromMultiSCM() { MultiSCM.class.toString(); }
+        
+        @SuppressFBWarnings(value = "RV_RETURN_VALUE_IGNORED_NO_SIDE_EFFECT", 
+                justification = "As designed, error spot-check")
+        public FromMultiSCM() {
+            // make this class fail to load if MultiSCM is not present
```
jtnord: Bogus comment the class has already loaded here - but not your issue :-)

src/main/java/org/jenkinsci/plugins/github/extension/GHEventsSubscriber.java
jtnord: can you then use the `notNull(Object, String)` so we may have some clue as to what is null in the stack trace - otherwise just supress is actually more explicit that you intent this to barf and not be fixed.

jtnord: reading through this I can't help but think it would be cleaner to have some intermediate classes
NullSafePredicate, NullSafeFunction etc that perform a check then call a different method `public void nsApply(@NotNull Object)`

src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
oleg-nenashev: <code>java.util.logging.Level</code> seems to be an unused import

lanwen: new classes uses slf4j

src/main/java/org/jenkinsci/plugins/github/extension/GHEventsSubscriber.java
lanwen: +1

oleg-nenashev:
It needs another update

pom.xml
```diff
@@ -168,7 +178,26 @@
...
             </configuration>
         </plugin>
-    </plugins>  
+        <plugin>
+          <groupId>org.codehaus.mojo</groupId>
+          <artifactId>findbugs-maven-plugin</artifactId>
+          <version>${findbugs-maven-plugin.version}</version>
+          <configuration>
+            <xmlOutput>true</xmlOutput>
+            <findbugsXmlWithMessages>true</findbugsXmlWithMessages>
```
KostyaSha: you know what to do with this fields :D

src/main/java/org/jenkinsci/plugins/github/webhook/WebhookManager.java
```diff
@@ -165,7 +167,7 @@ public GHHook apply(GitHubRepositoryName name) {
 
                     return createWebhook(endpoint, merged).apply(repo);
                 } catch (Throwable t) {
-                    LOGGER.warn("Failed to add GitHub webhook for {}", name, t);
+                    LOGGER.warn("Failed to add GitHub webhook for {} due to {}", name, t);
```
lanwen: it will be logged without explicitly adding

KostyaSha:
Not actual now, closing

jglick:
Sorry, why was this closed?

jglick:
Because #70 made it have merge conflicts? Then it would just need to be merged with `master` to reflect what is not yet there: to run FindBugs automatically. Or am I missing something?

lanwen:
@jglick we have plan to enable sonarQube analize in a few days. So it will be no need to enable findbug in pom. Also, most of findbug checks are false-negative and should be ignored. Thats why this PR is no more actual

KostyaSha:
Also @oleg-nenashev has no time to continue and part of blind FB fixes was wrong

oleg-nenashev:
Yes, my bad. I have been waiting for other PRs to be merged and finally forgotten to fix the left stuff. I'll resubmit the PR with valuable changes

PR #64, opened 2015-07-19
handle exceptions of extensions

lanwen:
[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/64)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/github/extension/GHEventsSubscriber.java
+++ b/src/main/java/org/jenkinsci/plugins/github/extension/GHEventsSubscriber.java
@@ -7,6 +7,8 @@ import hudson.ExtensionPoint;
 import hudson.model.AbstractProject;
 import jenkins.model.Jenkins;
 import org.kohsuke.github.GHEvent;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/github/extension/GHEventsSubscriber.java
```diff
@@ -122,7 +125,12 @@ public boolean apply(GHEventsSubscriber subscriber) {
         return new Function<GHEventsSubscriber, Void>() {
             @Override
             public Void apply(GHEventsSubscriber subscriber) {
-                subscriber.onEvent(event, payload);
+                try {
+                    subscriber.onEvent(event, payload);
+                } catch (Throwable t) {
+                    LOGGER.error("Subscriber {} failed to process {} hook, skipping...", 
+                            subscriber.getClass().getName(), event, t);
```
KostyaSha: Does subscriber has descriptor and it's name?

lanwen: No, only subscribable has descriptor.

KostyaSha:
:+1:

oleg-nenashev:
:+1:

PR #66, opened 2015-07-29
fix "Unknown content type null" in logs and since for 1.12.0 api

lanwen:
+ some small changes (see commits) to help implementing future prs

[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/66)

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/GitHubRepositoryName.java
+++ b/src/main/java/com/cloudbees/jenkins/GitHubRepositoryName.java
@@ -6,6 +6,7 @@ import org.kohsuke.github.GHCommitPointer;
 import org.kohsuke.github.GHRepository;
 import org.kohsuke.github.GitHub;
 
+import javax.annotation.CheckForNull;
 import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.URL;
@@ -118,9 +119,8 @@ public class GitHubRepositoryName {
      *
      * This is useful if the caller only relies on the read access to the repository and doesn't need to
      * walk possible candidates.
-     *
-     * Result can be null
      */
+    @CheckForNull
     public GHRepository resolveOne() {
         for (GHRepository r : resolve())
             return r;
--- a/src/main/java/com/cloudbees/jenkins/GitHubRepositoryName.java
+++ b/src/main/java/com/cloudbees/jenkins/GitHubRepositoryName.java
@@ -118,6 +118,8 @@ public class GitHubRepositoryName {
      *
      * This is useful if the caller only relies on the read access to the repository and doesn't need to
      * walk possible candidates.
+     *
+     * Result can be null
      */
     public GHRepository resolveOne() {
         for (GHRepository r : resolve())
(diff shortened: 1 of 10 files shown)


src/main/java/com/cloudbees/jenkins/GitHubRepositoryName.java
```diff
@@ -118,6 +118,8 @@ protected GHRepository adapt(GitHub item) {
      *
      * This is useful if the caller only relies on the read access to the repository and doesn't need to
      * walk possible candidates.
+     *
+     * Result can be null
      */
     public GHRepository resolveOne() {
```
KostyaSha: annotate with CheckForNull

src/main/java/org/jenkinsci/plugins/github/util/misc/NullSafeFunction.java
```diff
@@ -0,0 +1,29 @@
...
+import com.google.common.base.Function;
+import com.google.common.base.Preconditions;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+/**
+ * This abstract class calls {@link #applyNullSafe(Object)} only after success validation of inner object for null
+ * 
+ * {@inheritDoc}
```
KostyaSha: shouldn't parameters be inherited by default without `inheritDoc`?

src/main/java/org/jenkinsci/plugins/github/util/misc/NullSafePredicate.java
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * {@inheritDoc}
+ *
+ * @author lanwen (Merkushev Kirill)
+ */
+
+public abstract class NullSafePredicate<T> implements Predicate<T> {
+
+    /**
+     * {@inheritDoc}
```
KostyaSha: afaik they will be inherited by default, just remove

src/main/java/org/jenkinsci/plugins/github/webhook/GHEventPayload.java
```diff
@@ -80,7 +87,8 @@ public String apply(StaplerRequest request) {
         }
 
         /**
-         * used for application/json content-type 
+         * used for application/json content-type
```
KostyaSha: pure description...

lanwen: on next line - `@return function to extract payload from body` think its enough

src/main/java/com/cloudbees/jenkins/GitHubRepositoryName.java
```diff
@@ -118,7 +119,10 @@ protected GHRepository adapt(GitHub item) {
      *
      * This is useful if the caller only relies on the read access to the repository and doesn't need to
      * walk possible candidates.
+     *
+     * Result can be null
```
KostyaSha: remove this line, checkfornull is enough or if you want java doc then change to `@return desc`

PR #70, opened 2015-08-17
unexpected npe fixes for findbugs

lanwen:
contains rewrited parts from #63 (but not whole)

plz review @oleg-nenashev

think other findbugs issues can be ignored

[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/70)

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/GitHubPushCause.java
+++ b/src/main/java/com/cloudbees/jenkins/GitHubPushCause.java
@@ -1,8 +1,12 @@
 package com.cloudbees.jenkins;
 
 import hudson.triggers.SCMTrigger.SCMTriggerCause;
+
 import java.io.File;
 import java.io.IOException;
+
+import static java.lang.String.format;
+import static org.apache.commons.lang3.StringUtils.trimToEmpty;
 
 /**
  * UI object that says a build is started by GitHub post-commit hook.
@@ -31,7 +35,7 @@ public class GitHubPushCause extends SCMTriggerCause {
 
     @Override
     public String getShortDescription() {
-        String pusher = pushedBy != null ? pushedBy : "";
-        return "Started by GitHub push by " + pusher;
+        return format("Started by GitHub push by %s", trimToEmpty(pushedBy));
     }
 }
+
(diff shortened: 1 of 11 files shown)


src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
```diff
@@ -55,8 +55,8 @@ protected EnvVars buildEnv(AbstractProject<?, ?> job) {
             for (EnvironmentContributor contributor : EnvironmentContributor.all()) {
                 try {
                     contributor.buildEnvironmentFor(job, env, TaskListener.NULL);
-                } catch (Exception e) {
-                    LOGGER.debug("{} failed to build environment ({})", contributor.getClass(), e.getMessage());
+                } catch (Throwable t) {
```
oleg-nenashev: After the discussions in other PRs we agreed that this approach is not good, because it suppresses various kinds of system <code>Error</code>s, which are supposed to be system failures

oleg-nenashev:
I'm mostly OK with the fix, but actually it makes the further merge even harder.
Seems I'll have to respin the pull request from scratch

src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
lanwen: so catch only `Exception` ?

lanwen:
@oleg-nenashev All codebase has changed. This PR fix 99% of current fb issues so your pr can be just skipped as i have plan to connect sonarQube

oleg-nenashev:
Yes, seems so. BTW it makes sense to apply the POM changes. I'll do it latet

src/main/java/com/cloudbees/jenkins/GitHubPushCause.java
```diff
@@ -31,7 +37,16 @@ public GitHubPushCause(File pollingLog, String pusher) throws IOException {
...
+        return format("Started by GitHub push by %s", trimToEmpty(pushedBy));
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        return reflectionEquals(this, o);
+    }
+
+    @Override
+    public int hashCode() {
```
KostyaSha: wouldn't this influence on queue?

lanwen:
We (with @KostyaSha) decided not to change equals and hashcode in cause object as of it may brake something in job sheduling

PR #72, opened 2015-08-19
Enable FindBugs in the local build flows.

oleg-nenashev:
Just a POM patch, I decided not to integrate fixes now and just leave it to you.

@KostyaSha @lanwen

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -51,6 +51,9 @@
     <properties>
         <maven-compiler-plugin.version>3.3</maven-compiler-plugin.version>
         <maven-release-plugin.version>2.5.1</maven-release-plugin.version>
+        <findbugs-maven-plugin.version>3.0.1</findbugs-maven-plugin.version>
+        <!--TODO: enforce after cleaning up the issues-->
+        <findbugs.failOnError>false</findbugs.failOnError>
     </properties>
 
     <repositories>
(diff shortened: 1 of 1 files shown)


lanwen:
ok

pom.xml
```diff
@@ -208,6 +211,25 @@
...
                 </configuration>
             </plugin>
+            
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>findbugs-maven-plugin</artifactId>
+                <version>${findbugs-maven-plugin.version}</version>
+                <configuration>
+                    <xmlOutput>true</xmlOutput>
+                    <failOnError>${findbugs.failOnError}</failOnError>
```
KostyaSha: Why do you need redefine variable if you already set in property?

oleg-nenashev: I just keep it explicit, so it improves the POM file readability

lanwen: i've think it is to have ability to redefine with help of `-Dfindbugs.failOnError` ... if it works without this property - it should be inlined, as it doesn't encrease readability

oleg-nenashev: -Dfindbugs.failOnError can be redefined from the external run with such approach even if you don't specify this line explicitly

If you disagree with this string, I'll remove it

lanwen: for what we may need to use it in external build?

pom.xml
```diff
@@ -208,6 +211,25 @@
...
+                </configuration>
+                <executions>
+                    <execution>
+                        <id>run-findbugs</id>
```
KostyaSha: And why do you need define it if it
```
findbugs:check
Full name:

org.codehaus.mojo:findbugs-maven-plugin:3.0.2:check

Description:

(no description)
Attributes:
Binds by default to the lifecycle phase: verify.
```

pom.xml
lanwen: @oleg-nenashev (it may be used as separate goal if we wanna external use)

KostyaSha: How he didn't wrap it into profile yet...

PR #79, opened 2015-08-31
[FIXES JENKINS-30223] return back com.cloudbees.jenkins.Credential

lanwen:
https://issues.jenkins-ci.org/browse/JENKINS-30223

as of it makes fail to boot jenkins after installation of plugin which depends on this class.
Also remove migration to another package for this class and use it directly

linked: https://github.com/jenkinsci/elasticbox-plugin/pull/3

[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/79)


src/main/java/com/cloudbees/jenkins/Credential.java
```diff
@@ -0,0 +1,68 @@
...
+    @SuppressWarnings("visibilitymodifier")
+    public final transient String oauthAccessToken;
+
+    @DataBoundConstructor
+    public Credential(String username, String apiUrl, String oauthAccessToken) {
+        this.username = username;
+        this.apiUrl = apiUrl;
+        this.oauthAccessToken = oauthAccessToken;
+    }
+
```
KostyaSha: extra line

lanwen: fixed

PR #83, opened 2015-09-07
[JENKINS-24208] Allow to bann a pusher to avoid triggering loop when your job is push…

ouaibsky:
…ing to github
banned user can be set at plugin configuration (jenkins admin)
Can be probably enhanced by 
- accepting a list of users 
- move per job configuration

But good enough as a first shot.

It's my first time in jenkins plateform and groovy, let me know what can be improved

---

This change is [Review on Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/83)


src/main/java/org/jenkinsci/plugins/github/config/GitHubPluginConfig.java
```diff
@@ -101,6 +102,14 @@ public void setHookUrl(URL hookUrl) {
         }
     }
 
+    public String getBannedCommitter() {
```
KostyaSha: This shouldn't exist in global config

KostyaSha:
:-1: as it wouldn't be a final solution. After it you will need of course ability to ignore multiple pushers, then you will want regexp, later multiple regexpes...

KostyaSha:
Btw, duplicates #58

KostyaSha:
Could you open issue in jira? I think we will provide some solution later.

src/main/java/org/jenkinsci/plugins/github/config/GitHubPluginConfig.java
ouaibsky: OK but how can I use the configuration property into others classes ?
For example in DefaultPushGHEventSubscriber.
As described into http://jenkins-ci.org/pull-request-greeting I add change only for need ot this feature

I didn't see any injection mechanism in this plugin, I just copy things I saw in the plugin.

ouaibsky:
I put jenkins jira ID into PR title
Do I need more ?

#58 looks like a duplicate,  This is limitation of using JIRA / github, which one is the main repository fot ticket.
https://issues.jenkins-ci.org/browse/JENKINS-24208?jql=text%20~%20%22ignore%20users%22%20and%20component%20%3D%20github-plugin

lanwen:
Think plugin should not keep configuration for this feature. It already implemented in git-plugin, so need to find way to reuse this

ouaibsky:
Maybe leveraging on git-plugin is better but after a quick analysis sounds much more complex (need to master git and git plugin).

Main Benefits of filtering into this plugin is to discard very quickly the triggering of this job. don't need to do scm poll (mean fetch/checkout)

It's fast, simple and works, So it's a nice workaround awaiting someone has a more high level needs or want to deep dive into git-plugin sources.

Christophe

KostyaSha:
@lanwen you are wrong. Pusher name/organisation is specifically a github info, so it valid for push trigger.  When we switch to notifyCommit, we wouldn't use git-plugin polling features. 

@ouaibsky we have more critical things in todo, so this will be implemented later.

KostyaSha:
Btw, i did the same in https://github.com/jenkinsci/github-pullrequest-plugin but in more crazy manner with restrictions API.

ouaibsky:
@KostyaSha I understand you have more critical things todo.
It's exactly why Pull Request exists :-). Code is there, You just need time to review :)

KostyaSha:
@ouaibsky https://github.com/jenkinsci/github-plugin/pull/83#issuecomment-138341620

ouaibsky:
@KostyaSha Do you know what mean agile concept ? it's starting with basic story that provide added value for real use case. 

Right know we can consider there is a bug into github-plugin because if your jenkins job is pushing something on github, you enter into a loop.

I think many users are facing same issue and just adding the ability to ignore one user can help peoples to solve this pb.

Let's developed something more complex when you'll get a need and don't do plan on the comet right now.

I understand that this fix is not the way you would like to do it, but it fixes a bug and you don't have time right now to do something better. 
So let's add this fix and wait for your time (or someone else)

It think is fair enough.
Christophe

KostyaSha:
This is not a bug, you cycled your job to repo where this job commits. In the same way you can easily add a hack in job that will exclude main build part if it see that nothing new exists to break this cycle or chain 2 jobs. I used "job-decider" with conditioning to standard build job and didn't say that it somebodies bug.

I'm not closing this PR, but it wouldn't be merged in such state.

ouaibsky:
Ok. sounds we are not on same page and I have less time than you to dedicate to this issue.

The pb is very easy to reproduce
* Trigger a job with this plugin and start a commanbd like mvn jgitflow:release-start release-finish and have a look (or other command that push sometinh in github)
If you can provide something that is working without putting a plumb in place for such basic need, it will help.

If you still believe that having 2 open PR on the same issue and one jira ticket means there is no pb
and solution is not good enough for you. that's it.

I'll manage with my fork or #58 
C.

lanwen:
In any case merging temp solution is bad practice because of we should save backward compatibility forever. I'll think how to implement this in good way.

For now you can use custom build.

We plan to release workflow first, then create more common triggers (not only push), then reuse notifyCommit. That's the release queue.

KostyaSha:
@ouaibsky we plan implement more triggers and seems it will makes sense migrate security API from github-pullrequest-plugin as such securing will make sense for other push triggers. It not so easy because different triggers will require additional security restriction features. 
In general what you want is in TODO.

emoshaya-cognito:
This is a really crucial feature we need too.. Any ideas when this will be merged?

acejam:
My team could also significantly benefit from this. This is a critical setting in any true CI/CD workflow. Please let me know what's needed to help escalate this along, and I'll try to do the work.

eBeyond:
As the github plugin is used in jenkins pipeline it's no way to use a decider or any other things. It's even not possible to configure git-plugin for ignoring builds (as it also not works with github-webhooks). So you run into a loop if the job is pushing into the same repo.
As far as I can see it's also not a job of the git-plugin to evaluate the webhook of github to decide if a build is run. It's up to the github-plugin, as the main reason of this plugin, to decide to trigger a build or not.

slavakl:
Hello!
Is there any progress on this? We use GitHub Organization Folder Plugin and can't neither configure git-scm nor use decider to prevent a loop. We have to check committer at the beginning of pipeline, which works fine, however it create "empty" build runs that totally messes up statistics.

lanwen:
Hello, im working on proper implementation of the brand new SCMEventApi with Github as a source. Can't provide any estimates, but it should solve filtering problem.

ctoestreich:
@slavakl @lanwen We are in the same exact boat.  Using the GitHub Organization Folder Plugin is great... except we tag and push and get build loops.  @slavakl How are you checking for user on commit, i have tried several methods and have not come up with a reliable method.

PR #84, opened 2015-09-17
Let GithubProjectProperty be applied to any Job

jglick:
No reason for this to be limited to `AbstractProject`.

Note that `GithubLinkAction` will not currently appear on a Workflow job, pending [JENKINS-29880](https://issues.jenkins-ci.org/browse/JENKINS-29880).

@reviewbybees


KostyaSha:
:P fix checkstyle

src/main/java/com/coravy/hudson/plugins/github/GithubProjectProperty.java
```diff
@@ -31,8 +29,12 @@
     private String projectUrl;
 
     @DataBoundConstructor
-    public GithubProjectProperty(String projectUrl) {
-        this.projectUrl = new GithubUrl(projectUrl).baseUrl();
+    public GithubProjectProperty(String url) {
+        this.projectUrl = new GithubUrl(url).baseUrl();
+    }
+
+    public String getUrl() {
```
oleg-nenashev: Why didn't you like the field name?

src/main/java/com/coravy/hudson/plugins/github/GithubProjectProperty.java
```diff
@@ -58,10 +60,6 @@ public DescriptorImpl() {
             load();
         }
 
-        public boolean isApplicable(Class<? extends Job> jobType) {
```
oleg-nenashev: A downside is that the property will start appearing in other job types (External job monitor, Inheritance or whatever). BTW it should be safe enough since the property only produces an action

oleg-nenashev:
:ant: for the field name (at least I don't understand why <code>url</code> is better than <code>projectUrl</code>)

recena:
:bee:

KostyaSha:
:-1: for databinding change

recena:
@KostyaSha What is the problem here?

KostyaSha:
@recena because GH PP should contain url with project, so projectUrl is the right name. Ideally it should store GithubProjectUrl but it other story.

jglick:
Please read the message in my last commit which explains the problem and the solution, without which the test will fail.

KostyaSha:
What test?

lanwen:
Maybe just make ghurl class to be used everywhere as bean and add new constructor?

jglick:
> What test?

That added in 9b2131f3b199f3f6132ae8062fe03c14991889cc.

An alternate fix would be to simply rename the existing `GithubUrl getProjectUrl()` method to something else, and introduce `String getProjectUrl()` that merely echoes back the constructor parameter. I could trivially do that rename inside this plugin, of course, but then [this](https://github.com/jenkinsci/github-pullrequest-plugin/blob/a9606f0eda4d6a030bf9731c2ab22dac4b58ddd9/src/main/java/org/jenkinsci/plugins/github/pullrequest/GitHubPRRepositoryFactory.java#L48) and [this](https://github.com/jenkinsci/github-pullrequest-plugin/blob/8208905e1a928fecbefc7d0a2e4c8cc4c29179fb/src/main/java/org/jenkinsci/plugins/github/pullrequest/GitHubPRTrigger.java#L161) and [this](https://github.com/jenkinsci/elasticbox-plugin/blob/86df8fba5fef7cff7d1c4cbc1b54d1a1889f22fe/src/main/java/com/elasticbox/jenkins/triggers/github/PullRequestBuildHandler.java#L90) would be broken. So the current patch instead preserves binary compatibility at the cost of changing the field name used in databinding.

jglick:
> make ghurl class to be used everywhere as bean and add new constructor

Will not work, at least not without major changes, since Stapler is not going to understand this custom type in databinding.

lanwen:
ok for me :+1:

src/main/java/com/coravy/hudson/plugins/github/GithubProjectProperty.java
KostyaSha: Add javadoc

tfennelly: Where is this method used?

tfennelly:
The `getUrl` should have some docs on it. Any reason why a more descriptive name wouldn't work e.g. `getProjectUrlAsString`. I'd error on the side of it being too verbose.

src/main/java/com/coravy/hudson/plugins/github/GithubProjectProperty.java
```diff
@@ -21,9 +20,8 @@
  * As of now this is only the URL to the github project.
  *
  * @author Stefan Saasen <stefan@coravy.com>
- * @todo Should we store the GithubUrl instead of the String?
```
KostyaSha: Why you removed this line?

src/test/java/com/coravy/hudson/plugins/github/GithubProjectPropertyTest.java
```diff
@@ -0,0 +1,46 @@
...
+import org.jenkinsci.plugins.workflow.structs.DescribableHelper;
+import org.junit.Test;
+import static org.junit.Assert.*;
+import org.junit.Rule;
+import org.jvnet.hudson.test.JenkinsRule;
+
+public class GithubProjectPropertyTest {
+
+    @Rule
+    public JenkinsRule r = new JenkinsRule();
```
KostyaSha: jenkinsRule is called `j` in all tests

src/main/java/com/coravy/hudson/plugins/github/GithubProjectProperty.java
KostyaSha: in field 'url' dunno what is the problem for jessy calling from UI 'getProjectUrl().base()` directly or `toString()`

KostyaSha:
> Note that GithubLinkAction will not currently appear on a Workflow job, pending JENKINS-29880.

No sense in merging this until workflow will support it

KostyaSha:
> No reason for this to be limited to AbstractProject.

Obviously `ProjectProperty` was created for `AbstractProject`?

src/main/java/com/coravy/hudson/plugins/github/GithubProjectProperty.java
```diff
@@ -58,10 +60,6 @@ public DescriptorImpl() {
             load();
         }
 
-        public boolean isApplicable(Class<? extends Job> jobType) {
-            return AbstractProject.class.isAssignableFrom(jobType);
```
lanwen: maybe point this to same scope as trigger? As we have plans to extend property with credentials, or maybe additional GH information. (In this case it should not break anything but will be strange in corner cases)

src/test/java/com/coravy/hudson/plugins/github/GithubProjectPropertyTest.java
lanwen: think both `r` or `j` is not good enough. Should be `rule` (or `jRule`) i think.

KostyaSha: everybody knows that `j` is jenkins!

jglick:
> No sense in merging this until workflow will support it

Well, that is only one of (at least) two known queriers for this property.

src/main/java/com/coravy/hudson/plugins/github/GithubProjectProperty.java
jglick: Because the proposal is not a good one. Better to keep the XStream form as simple as possible, and where possible (not here, alas) aligned with the databinding form.

src/main/java/com/coravy/hudson/plugins/github/GithubProjectProperty.java
jglick: Because the databinding control as well as non-UI tools like `DescribableHelper` expect to call the bean getter directly.

src/main/java/com/coravy/hudson/plugins/github/GithubProjectProperty.java
jglick: I see no reason to artificially restrict it.

src/test/java/com/coravy/hudson/plugins/github/GithubProjectPropertyTest.java
jglick: `jenkins` and `jRule` are used elsewhere in the plugin; is there a single preference?

KostyaSha: never saw `jRule` in core/plugins. I prefer `j` as i see it everywhere.

src/main/java/com/coravy/hudson/plugins/github/GithubProjectProperty.java
```diff
@@ -31,8 +30,18 @@
...
+        this.projectUrl = new GithubUrl(url).baseUrl();
+    }
+
+    /**
+     * Same as {@link #getProjectUrl}, but with a property name and type
+     * which match those used in the {@link #GithubProjectProperty} constructor.
+     * Should have been called {@code getProjectUrl} and that method called something else
+     * (such as {@code getNormalizedProjectUrl}), but that cannot be done compatibly now.
+     */
+    public String getUrl() {
```
KostyaSha: Just call it `getProjectUrlStr()` that is also common pattern in many plugins and core.

KostyaSha: I mean add javadoc that describes that this method returns full url to project. Because GH related code can deal with ghUrl as a plain api url, or url to project/repository. Simple 'url' withour javadoc doesn't provide good description.

src/test/java/com/coravy/hudson/plugins/github/GithubProjectPropertyTest.java
```diff
@@ -0,0 +1,46 @@
+/*
+ * Copyright 2015 CloudBees, Inc.
```
KostyaSha: btw, @oleg-nenashev introduced chaos into this plugin, root license is MIT, but pom defines apache. Could you remove license header for now? Or it required by your company rules?

oleg-nenashev: Uhh, sorry. Seems I didn't follow-up to https://github.com/jenkinsci/github-plugin/commit/97f1c9dd76f6af91da4dfb6f343fda6de8a2727f

oleg-nenashev:
I don't see how filtering by <code>ParameterizedJobMixIn.ParameterizedJob</code> or any other condition helps to address the very abstract @lanwen's use-case, but it's up to the plugin maintainers.

:bee: for others

jglick:
The filtering ought to hide this property from `ExternalJob`s and so on, as you mentioned. It should be harmless on such special `Job`s, but probably not useful, given that the primary plugin feature is the `GitHubPushTrigger`, which is limited to `ParameterizedJob`s for good reason.

KostyaSha:
> the primary plugin feature is the GitHubPushTrigger

This wouldn't be anymore right soon. When we will resolve our plan instead of interrupting on new PRs. After this i suppose it will be clear to what types trigger and PP can be applied.

jglick:
@reviewbybees done

lanwen:
merged manually in 3990a367f84791f37dc978cbc5cc92e3f7b3a339

KostyaSha:
@lanwen JENKINS-29880 is not resolved

lanwen:
@KostyaSha JENKINS-29880 is not blocker for this code

PR #82, opened 2015-09-03
[FIXES JENKINS-27136] Workflow plugin support

lanwen:
Superseed of #59 (squashed and rebased)

[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/82)

PR #85, opened 2015-09-22
Fixed Javadoc link

jglick:
Fixes links orphaned by #82. @reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/github/extension/GHEventsSubscriber.java
+++ b/src/main/java/org/jenkinsci/plugins/github/extension/GHEventsSubscriber.java
@@ -4,7 +4,6 @@ import com.google.common.base.Function;
 import com.google.common.base.Predicate;
 import hudson.ExtensionList;
 import hudson.ExtensionPoint;
-import hudson.model.AbstractProject;
 import hudson.model.Job;
 import jenkins.model.Jenkins;
 import org.jenkinsci.plugins.github.util.misc.NullSafeFunction;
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/github/extension/GHEventsSubscriber.java
```diff
@@ -23,7 +22,7 @@
 
 /**
  * Extension point to subscribe events from GH, which plugin interested in.
- * This point should return true in {@link #isApplicable(AbstractProject)}
+ * This point should return true in {@link #isApplicable}
```
jtnord: I beleive this is an invalid link would refer to the non-existant field `isApplicable` unless something has changed?

jglick: `isApplicable` is a method, not a field.

jglick: Specifically `javadoc:javadoc` with JDK7u80 produces

```html
 This point should return true in <a href="../../../../../org/jenkinsci/plugins/github/extension/GHEventsSubscriber.html#isApplicable(hudson.model.Job)"><code>isApplicable(hudson.model.Job&lt;?, ?&gt;)</code></a>
```

as expected.

jtnord: That was my point - I thought you needed to add `(argumentType)` for links to methods. 
I was mistaken.

jtnord:
:bee:

oleg-nenashev:
:bee: . BTW a little :ant: for a risk of Javadoc ambiguous method links if somebody introduces a new isApplicable() method

jglick:
Yes, it is a risk. On the other hand there is a risk of a link breaking if someone changes the type of a parameter, as happened here. :-)

lanwen:
thx!

PR #88, opened 2015-09-23
using okHttp connector with cache and proxy for github

lanwen:
this can speed up fetching data form GH and reduce rate limits consuming

[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/88)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/github/config/GitHubPluginConfig.java
+++ b/src/main/java/org/jenkinsci/plugins/github/config/GitHubPluginConfig.java
@@ -3,7 +3,6 @@ package org.jenkinsci.plugins.github.config;
 import com.cloudbees.jenkins.GitHubWebHook;
 import com.google.common.base.Predicate;
 import com.google.common.base.Predicates;
-
 import hudson.Extension;
 import hudson.XmlFile;
 import hudson.model.Descriptor;
@@ -25,7 +24,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.inject.Inject;
-
 import java.io.IOException;
 import java.net.HttpURLConnection;
 import java.net.MalformedURLException;
@@ -55,6 +53,13 @@ public class GitHubPluginConfig extends GlobalConfiguration {
     public static final String GITHUB_PLUGIN_CONFIGURATION_ID = "github-plugin-configuration";
 
     /**
+     * Default value in MB for client cache size
+     *
+     * @see #getClientCacheSize()
+     */
+    private static final int DEFAULT_CLIENT_CACHE_SIZE_MB = 20;
+
+    /**
      * Helps to avoid null in {@link GitHubPlugin#configuration()}
      */
     public static final GitHubPluginConfig EMPTY_CONFIG =
@@ -62,6 +67,13 @@ public class GitHubPluginConfig extends GlobalConfiguration {
 
     private List<GitHubServerConfig> configs = new ArrayList<GitHubServerConfig>();
     private URL hookUrl;
+
+    /**
+     * @see #getClientCacheSize()
+     * @see #setClientCacheSize(int)
+     */
+    private int clientCacheSize = DEFAULT_CLIENT_CACHE_SIZE_MB;
+
     private transient boolean overrideHookUrl;
 
     /**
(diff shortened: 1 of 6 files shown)


oleg-nenashev:
How does cache impact the request results reliability? In the code there is no invalidation hooks for the cache, so I suspect the sensitive information (group membership, login status responses, etc.) may be cached and lead to mistakes in the case of the improper usage. Does it also allow to brute-force the oauth token w/o going to real GitHub checks?

lanwen:
GH + okHttp do all work for results reliability (https://developer.github.com/v3/#conditional-requests). 

How can token be brute-forced? Think stealing token from credentials is much easer.

KostyaSha:
@oleg-nenashev please learn github documentation ;)

src/main/java/org/jenkinsci/plugins/github/internal/GitHubLoginFunction.java
```diff
@@ -0,0 +1,151 @@
...
+ * @author lanwen (Merkushev Kirill)
+ * @see GitHubServerConfig#loginToGithub()
+ */
+@Restricted(NoExternalUse.class)
+public class GitHubLoginFunction extends NullSafeFunction<GitHubServerConfig, GitHub> {
+
+    /**
+     * Capacity of cache for GitHub client in MB.
+     *
+     * Can be overridden as system property -Dorg.jenkinsci.plugins.github.GitHubPlugin.cache.size.mb=1 on jenkins jvm
```
KostyaSha: No, don't use properties, they sucks. User can't find what and where configure.

KostyaSha:
It also misses reconnection algorithm that i so tuned and tested :(

oleg-nenashev:
@KostyaSha please study community behavior BKMs :P

src/main/java/org/jenkinsci/plugins/github/internal/GitHubLoginFunction.java
lanwen: think user should not configure it until very big troubles

KostyaSha:
@oleg-nenashev do you remember rtfm times? ;)

src/main/java/org/jenkinsci/plugins/github/internal/GitHubLoginFunction.java
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    /**
+     * Uses proxy if configured on pluginManager/advanced page
+     *
+     * @param apiUrl GitHub's url to build proxy to
+     *
+     * @return proxy to use it in connector
+     */
+    private Proxy getProxy(String apiUrl) {
```
KostyaSha: `@Nonnull` i suppose

lanwen: think in private methods it can be ignored

KostyaSha: It for static analysing, nobody save you from npes in private methods :)

src/main/resources/org/jenkinsci/plugins/github/config/GitHubPluginConfig/config.groovy
```diff
@@ -38,6 +38,10 @@ f.section(title: descriptor.displayName) {
             }
         }
 
+        f.entry(title: _("GitHub client cache size (Mb)"), field: "clientCacheSize") {
```
KostyaSha: `MB`

src/main/java/org/jenkinsci/plugins/github/internal/GitHubLoginFunction.java
lanwen: It for extra checks for null when use something. If method private and used in one place - its overhead to annotate methods with nonnull

src/main/java/org/jenkinsci/plugins/github/config/GitHubPluginConfig.java
```diff
@@ -55,6 +53,13 @@
     public static final String GITHUB_PLUGIN_CONFIGURATION_ID = "github-plugin-configuration";
 
     /**
+     * Default value in MB for client cache size
+     *
+     * @see #getClientCacheSize()
+     */
+    private static final int DEFAULT_CLIENT_CACHE_SIZE_MB = 20;
```
KostyaSha: It should be in GitHubServerConfig

src/main/resources/org/jenkinsci/plugins/github/config/GitHubPluginConfig/help-clientCacheSize.html
```diff
@@ -0,0 +1,7 @@
+<div>
+    Cache size in MB used by GitHub client. This can speed up fetching data form GH and reduce rate limits consuming.<br/>
+    GH + okHttp do all work for results reliability 
+    (<a href="https://developer.github.com/v3/#conditional-requests">Conditional-requests in GitHub documentation</a>)<br/>
+    
+    Set to <b>0</b> to disable this feature
```
KostyaSha: doble `to`, maybe just `Set 0 to disable..`?

lanwen: yep

PR #89, opened 2015-09-23
[JENKINS-30626] Handle pings from organization webhooks

jglick:
[JENKINS-30626](https://issues.jenkins-ci.org/browse/JENKINS-30626)

[Reference](https://developer.github.com/v3/orgs/hooks/)

@reviewbybees esp. @recena who discovered this bug.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/github/webhook/subscriber/PingGHEventSubscriber.java
+++ b/src/main/java/org/jenkinsci/plugins/github/webhook/subscriber/PingGHEventSubscriber.java
@@ -10,7 +10,7 @@ import org.slf4j.LoggerFactory;
 import java.util.Set;
 
 import static com.google.common.collect.Sets.immutableEnumSet;
-import static net.sf.json.JSONObject.fromObject;
+import net.sf.json.JSONObject;
 import static org.kohsuke.github.GHEvent.PING;
 
 /**
@@ -52,8 +52,18 @@ public class PingGHEventSubscriber extends GHEventsSubscriber {
      */
     @Override
     protected void onEvent(GHEvent event, String payload) {
-        // something like <https://github.com/bar/foo>
-        String repo = fromObject(payload).getJSONObject("repository").getString("url");
-        LOGGER.info("{} webhook received from repo <{}>!", event, repo);
+        JSONObject parsedPayload = JSONObject.fromObject(payload);
+        JSONObject repository = parsedPayload.optJSONObject("repository");
+        if (repository != null) {
+            // something like <https://github.com/bar/foo>
+            LOGGER.info("{} webhook received from repo <{}>!", event, repository.getString("url"));
+        } else {
+            JSONObject organization = parsedPayload.optJSONObject("organization");
+            if (organization != null) {
+                LOGGER.info("{} webhook received from org <{}>!", event, organization.getString("url"));
+            } else {
+                LOGGER.warn("{} webhook received with unexpected payload", event);
+            }
+        }
     }
 }
--- a/src/main/java/org/jenkinsci/plugins/github/webhook/subscriber/PingGHEventSubscriber.java
+++ b/src/main/java/org/jenkinsci/plugins/github/webhook/subscriber/PingGHEventSubscriber.java
@@ -2,6 +2,7 @@ package org.jenkinsci.plugins.github.webhook.subscriber;
 
 import hudson.Extension;
 import hudson.model.Job;
+import net.sf.json.JSONObject;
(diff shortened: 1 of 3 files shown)


recena:
Of course! :bee: and :+1:

oleg-nenashev:
:bee: , but seems to be significant enough to have a JIRA issue. Does the original code fail with NPE?

amuniz:
:bee:

lanwen:
Please add a test with such payload same as for the ping.

src/main/java/org/jenkinsci/plugins/github/webhook/subscriber/PingGHEventSubscriber.java
```diff
@@ -10,7 +10,7 @@
 import java.util.Set;
 
 import static com.google.common.collect.Sets.immutableEnumSet;
-import static net.sf.json.JSONObject.fromObject;
+import net.sf.json.JSONObject;
```
KostyaSha: then it should be not in static block of imports

src/main/java/org/jenkinsci/plugins/github/webhook/subscriber/PingGHEventSubscriber.java
```diff
@@ -52,8 +52,18 @@ protected boolean isApplicable(Job<?, ?> project) {
      */
     @Override
     protected void onEvent(GHEvent event, String payload) {
-        // something like <https://github.com/bar/foo>
-        String repo = fromObject(payload).getJSONObject("repository").getString("url");
-        LOGGER.info("{} webhook received from repo <{}>!", event, repo);
+        JSONObject parsedPayload = JSONObject.fromObject(payload);
```
lanwen: why you remove static import?

KostyaSha: for using JSONObject

lanwen:
:+1: thanks. I'll going to squash this and merge manually

recena:
@lanwen Why closed? /cc @jglick

KostyaSha:
Please read what GH wrote: `@jglickjglick closed this in 03b1c87 8 minutes ago`

recena:
@KostyaSha Thanks. I have to learn why this way is better (than traditional merge).

KostyaSha:
It not better, GH closes relevant PR when changes committed in target branch directly. Faster fix ourself instead of commenting and waiting for changes from PR author.

recena:
@KostyaSha ah ok. Thanks for your time.

src/main/java/org/jenkinsci/plugins/github/webhook/subscriber/PingGHEventSubscriber.java
```diff
@@ -52,8 +53,18 @@ protected boolean isApplicable(Job<?, ?> project) {
      */
     @Override
     protected void onEvent(GHEvent event, String payload) {
-        // something like <https://github.com/bar/foo>
-        String repo = fromObject(payload).getJSONObject("repository").getString("url");
-        LOGGER.info("{} webhook received from repo <{}>!", event, repo);
+        JSONObject parsedPayload = fromObject(payload);
+        JSONObject repository = parsedPayload.optJSONObject("repository");
+        if (repository != null) {
+            // something like <https://github.com/bar/foo>
```
jglick: Not sure it matters, but not that this is not actually the form of the URL; it is something involving `api.github.com`.

jglick:
> GH closes relevant PR when changes committed in target branch

Well, when the special `closes` keyword is used in the squashed commit comment. It only treats the PR as having the merged status if the head commit of the PR actually becomes an ancestor of the head commit of the target branch, which will not happen when you squash.

KostyaSha:
Special keyword is not required. When diff has nothing GH closes it.

src/main/java/org/jenkinsci/plugins/github/webhook/subscriber/PingGHEventSubscriber.java
recena: @jglick For that, there are `GHEventsSubscriber` [validating](https://github.com/jenkinsci/github-plugin/blob/master/src/main/java/org/jenkinsci/plugins/github/webhook/subscriber/DefaultPushGHEventSubscriber.java#L73) the URL.

jglick: My point was just that the comment is wrong.

PR #90, opened 2015-09-27
fix checking creds on github enterprise

lanwen:
also additional fixes mentioned in #88

[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/90)


src/main/java/org/jenkinsci/plugins/github/config/GitHubServerConfig.java
```diff
@@ -257,6 +257,7 @@ public FormValidation doVerifyCredentials(
                 @QueryParameter String apiUrl, @QueryParameter String credentialsId) throws IOException {
 
             GitHubServerConfig config = new GitHubServerConfig(credentialsId);
+            config.setCustomApiUrl(isUrlCustom(apiUrl));
```
KostyaSha: that's why i prefered to have direct url without any boolean checkboxes

lanwen: when this component will be in jenkins, it will be much easier to avoid such logic here

oleg-nenashev:
LGTM

KostyaSha:
:+1:

PR #91, opened 2015-09-27
simply return on empty list of events to subscribe, without error

lanwen:
In case of any other plugin try to register hooks without checking for subscribers. We don't want to write error to log

[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/91)


KostyaSha:
> In case of any other plugin try to register hooks without checking for subscribers.

What is the case?

lanwen:
I've registered hooks when starting gihub-pullrequest plugin, but subscriber checks that trigger should be in hooks mode, when it actually was in cron mode. So the result of all events from all subscribers was empty list. 

Then i've got error in the log. (I'll fix such situation in gh-pullreq plugin, but think fix should be here too)
This situation never happened when using this plugin only.

KostyaSha:
LGTM

PR #93, opened 2015-10-04
Separate cache for each server config

lanwen:
This change requires java 1.7 to use java.nio as more effective way to operate files.
This implementation creates directory with name as hash of each pair `<api url+creds>`.
Redundant dirs removed on each save of global config.

- [x] add the javadocs
- [x] some tests for cleanup of cache dirs


pom.xml
```diff
@@ -180,8 +180,8 @@
             <plugin>
                 <artifactId>maven-compiler-plugin</artifactId>
                 <configuration>
-                    <source>1.6</source>
-                    <target>1.6</target>
+                    <source>1.7</source>
```
KostyaSha: TODO don't forget mark in changelog 1.7 requirement

src/main/resources/org/jenkinsci/plugins/github/config/GitHubServerConfig/config.groovy
```diff
@@ -14,20 +14,27 @@ f.entry(title: _("Credentials"), field: "credentialsId") {
...
+    }
+}
+
 f.block() {
     f.validateButton(
             title: _("Verify credentials"),
             progress: _("Verifying..."),
             method: "verifyCredentials",
-            with: "apiUrl,credentialsId"
+            with: "apiUrl,credentialsId,clientCacheSize"
```
KostyaSha: Does your IDE automatically adds spaces with autoformatting?

lanwen: in groovy strings - no

src/main/java/org/jenkinsci/plugins/github/internal/GitHubLoginFunction.java
```diff
@@ -101,16 +99,13 @@ private Proxy getProxy(String apiUrl) {
...
      * If cache size > 0, uses cache
      *
-     * @param apiUrl to build proxy
-     *
      * @return connector to be used as backend for client
      */
-    private OkHttpConnector connector(String apiUrl) {
-        OkHttpClient client = new OkHttpClient().setProxy(getProxy(apiUrl));
+    private OkHttpConnector connector(GitHubServerConfig config) {
+        OkHttpClient client = new OkHttpClient().setProxy(getProxy(defaultIfBlank(config.getApiUrl(), GITHUB_URL)));
```
KostyaSha: wouldn't this mask issue when something is broken and url get blank?
i.e. user entered url, jelly magic failed, connection created with public connection and user didn't get thrown error

src/main/resources/org/jenkinsci/plugins/github/config/GitHubServerConfig/config.groovy
KostyaSha: right, it string.

KostyaSha:
:+1: LGTM besides using defaults when real url is expected.

src/main/java/org/jenkinsci/plugins/github/internal/GitHubLoginFunction.java
lanwen: Insurance just in case. This should never happen until somebody will call directly classes from internals

PR #97, opened 2015-10-07
ignore files when clean root cache dir

lanwen:
add tests for notInCache filter

[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/97)


KostyaSha:
LGTM

PR #100, opened 2015-10-27
[FIXES JENKINS-27154] configurable build context and message in status builder and notifier

lanwen:
- [x] expandable message with token-macro support
- [x] configurable GH property
- [x] notifier with configurable section for message
- [x] build step with configurable section for message

[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/100)


pom.xml
```diff
@@ -123,6 +123,28 @@
...
+            <version>2.4.0</version>
+            <scope>test</scope>
+        </dependency>
+        
+        <!--to expand message content-->
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>email-ext</artifactId>
+            <version>2.38.2</version>
+            <optional>true</optional>
```
KostyaSha: I think we should check with @slide about providing better API or extract `$SCRIPT` tokens into separate plugin.

KostyaSha: And allow configure some additional directory instead `email-ext`

src/main/java/org/jenkinsci/plugins/github/common/ExpandableMessage.java
```diff
@@ -0,0 +1,126 @@
...
+ *
+ * uses https://wiki.jenkins-ci.org/display/JENKINS/Token+Macro+Plugin to expand vars
+ *
+ * @author Kanstantsin Shautsou
+ * @author Alina Karpovich
+ * @since 1.14.1
+ */
+public class ExpandableMessage extends AbstractDescribableImpl<ExpandableMessage> {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ExpandableMessage.class);
```
KostyaSha: Rename to LOG for shorter variable

src/main/java/org/jenkinsci/plugins/github/common/ExpandableMessage.java
```diff
@@ -0,0 +1,126 @@
...
+                return TokenMacro.expandAll((AbstractBuild) run, listener, content, false, loadPrivateTokens());
+            } catch (MacroEvaluationException e) {
+                LOGGER.error("Can't process token content {} in {} ({})",
+                        content, run.getParent().getFullName(), e.getMessage());
+                LOGGER.trace(e.getMessage(), e);
+                return content;
+            }
+        } else {
+            // fallback to env vars only because of token-macro allow only AbstractBuild in 1.11
+            return run.getEnvironment(listener).expand(trimToEmpty(content));
```
KostyaSha: this will only contribute vars from node=env. ParameterActions need to be extracted in case of parametrised run.

src/main/java/org/jenkinsci/plugins/github/common/ExpandableMessage.java
```diff
@@ -0,0 +1,126 @@
...
+
+    /**
+     * Expands all env vars. In case of AbstractBuild also expands token macro and build vars
+     *
+     * @param run      build context
+     * @param listener usually used to log something to console while building env vars
+     *
+     * @return string with expanded vars and tokens
+     */
+    public String expandAll(Run<?, ?> run, TaskListener listener) throws IOException, InterruptedException {
```
KostyaSha: can be converted to static with third argument

src/test/java/org/jenkinsci/plugins/github/common/ExpandableMessageTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+        assertThat("should not change", expander.getResult(), is(DEFAULT_EMAIL_EXT_TEMPLATE));
+    }
+
+
+    public static String asVar(String name) {
+        return format("${%s}", name);
+    }
+
+    private static class MessageExpander extends TestBuilder {
```
KostyaSha: `MessageExpanderBuilder` to exclude ```MessageExpander expander = new MessageExpander(new ExpandableMessage(```

src/main/java/org/jenkinsci/plugins/github/common/ExpandableMessage.java
lanwen: run haven't method to expand build vars :(

src/main/java/org/jenkinsci/plugins/github/common/ExpandableMessage.java
KostyaSha: Maybe better make it null safe?

src/main/java/org/jenkinsci/plugins/github/common/ExpandableMessage.java
KostyaSha: For workflow i plan experiment later, we can announce workflow support as "experimental"

src/main/java/org/jenkinsci/plugins/github/common/ExpandableMessage.java
lanwen: what for?

src/test/java/org/jenkinsci/plugins/github/common/ExpandableMessageTest.java
lanwen: don't understand

src/main/java/org/jenkinsci/plugins/github/common/ExpandableMessage.java
lanwen: doesn't matter i think

pom.xml
lanwen: can be just a module in plugin

src/test/java/org/jenkinsci/plugins/github/common/ExpandableMessageTest.java
KostyaSha: Name builder in different way as current naming is not obvious when your read code from up to down.
`new MessageExpander(new ExpandableMessage(` looks very confusing

KostyaSha:
> notifier with configurable sections for context and message
>  build step with configurable sections for context and message

Imho let's keep current trigger as is and deprecate in favor of future generic trigger. I don't think that current trigger will be possible convert to new.

pom.xml
slide: Will try and pull this out into the token-macro plugin soon.

src/test/java/org/jenkinsci/plugins/github/common/ExpandableMessageTest.java
lanwen: It expands message. How it does it - doesn't matter. Via builder, publisher etc.

lanwen:
this PR is not about trigger change. It only about steps (build and publish)

src/test/java/org/jenkinsci/plugins/github/common/ExpandableMessageTest.java
KostyaSha: When do you read this could firstly you have no idea what is this object about, that's why having Builder tail in tests enhances readability because allow understand that it constructs `Builder` that expected to be added to job.

lanwen: This test doesn't test any builder, it just test how the message expands. You shouldn't know how the way it does. So if in future the expander will be a publisher, it will stay same expander as here. Don't mix contexts of the test.

KostyaSha:
Ah, and scripts support from private email-ext macroses probably may be unsecure.

src/main/java/com/cloudbees/jenkins/GitHubCommitNotifier.java
```diff
@@ -34,6 +45,10 @@
  */
 public class GitHubCommitNotifier extends Notifier {
 
+    private static final Logger LOGGER = LoggerFactory.getLogger(GitHubCommitNotifier.class);
```
KostyaSha: `LOG` please as shorter variant

src/main/resources/com/coravy/hudson/plugins/github/GithubProjectProperty/config.groovy
```diff
@@ -0,0 +1,17 @@
...
+
+def f = namespace(lib.FormTagLib);
+
+f.optionalBlock(name: GITHUB_PROJECT_BLOCK_NAME, title: _('github.project'), checked: instance != null) {
+    f.entry(field: 'projectUrlStr', title: _('github.project.url')) {
+        f.textbox()
+    }
+
+    f.advanced() {
+        f.entry(title: _('github.build.status.context.for.commits'), field: 'statusContext') {
```
KostyaSha: maybe just `displayName` as existed jenkins pattern? I think it will be usable not only for contexts

lanwen: its local var which only status context and nothing else. So changing it to display name can mislead

src/main/resources/com/coravy/hudson/plugins/github/GithubProjectProperty/help-statusContext.html
```diff
@@ -0,0 +1,11 @@
+<div>
+    <p>
```
KostyaSha: And then just say in help that display name will be used for any UI places where job name is expected.

src/main/resources/com/coravy/hudson/plugins/github/GithubProjectProperty/help-statusContext.html
```diff
@@ -0,0 +1,11 @@
...
+        This value will be used as context name for
+        <a href="https://developer.github.com/v3/repos/statuses/">commit status</a> if status builder or 
+        status publisher is defined for this project.
+    </p>
+
+    <p>
+        If you leave it empty, job name will be used for builder and publisher.
+    </p>
+</div>
```
KostyaSha: new line, IDEA always not adding it :(

src/main/resources/com/coravy/hudson/plugins/github/GithubProjectProperty/config.groovy
KostyaSha: from github-plugin property and plugin viewpoint it name that will be displayed instead of existed job name.

src/main/resources/com/coravy/hudson/plugins/github/GithubProjectProperty/help-statusContext.html
lanwen: resource files can be ignored i think

KostyaSha: AFAIK keeping NL is an approach from history - you should have end marker for streams of data. Though today it not critical for high level langs :)

src/main/java/com/cloudbees/jenkins/GitHubCommitNotifier.java
```diff
@@ -109,38 +116,74 @@ public boolean perform(AbstractBuild<?, ?> build,
...
+                    GHCommitState.SUCCESS,
+                    Messages.CommitNotifier_Success(build.getDisplayName(), duration)
+            );
+        } else if (result.isBetterOrEqualTo(UNSTABLE)) {
+            return new StatusResult(
+                    GHCommitState.FAILURE,
+                    Messages.CommitNotifier_Unstable(build.getDisplayName(), duration)
+            );
+        } else {
+            return new StatusResult(
```
oleg-nenashev: What about ABORTED?

src/main/java/com/coravy/hudson/plugins/github/GithubProjectProperty.java
```diff
@@ -59,13 +84,29 @@ public GithubUrl getProjectUrl() {
...
+    /**
+     * Extracts value of display name from given job, or just returns full name if field or prop is not defined
+     *
+     * @param job project which wants to get current context name to use in GH status API
+     *
+     * @return display name or full job name if field is not defined
+     */
+    public static String displayNameFor(@Nonnull Job<?, ?> job) {
+        GithubProjectProperty ghProp = job.getProperty(GithubProjectProperty.class);
+        if (ghProp != null && isNotBlank(ghProp.getDisplayName())) {
```
oleg-nenashev: NIT: string trimming?

src/main/java/org/jenkinsci/plugins/github/common/ExpandableMessage.java
oleg-nenashev: @lanwen 
There are some substitution helpers here: https://github.com/jenkinsci/perforce-plugin/tree/master/src/main/java/hudson/plugins/perforce/utils 

Technically I can decouple it to a separate lib if it helps

oleg-nenashev:
I like the concept. Looking forward to see it finalized and merged :)

src/main/java/com/cloudbees/jenkins/GitHubCommitNotifier.java
lanwen: if build will be aborted - this code never be runned

src/main/java/com/coravy/hudson/plugins/github/GithubProjectProperty.java
lanwen: notInBlank will check trimming (what isNotEmpty will not)

src/main/java/org/jenkinsci/plugins/github/common/ExpandableMessage.java
lanwen: hmm, how it can help here?

src/main/java/com/cloudbees/jenkins/GitHubCommitNotifier.java
KostyaSha: right because there is no `runAfterFinalizer`

KostyaSha:
Btw, email-ext exposed some tokens.

pom.xml
```diff
@@ -123,6 +124,12 @@
             <version>1.1</version>
         </dependency>
 
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>token-macro</artifactId>
```
KostyaSha: Why not make optional? Plugin lookup and method extraction is pretty simple and shown in pr plugin.

src/test/java/org/jenkinsci/plugins/github/test/WithoutPlugins.java
```diff
@@ -0,0 +1,29 @@
...
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * @author lanwen (Merkushev Kirill)
```
KostyaSha: javadoc?

src/main/java/com/cloudbees/jenkins/GitHubCommitNotifier.java
```diff
@@ -45,9 +60,13 @@ public GitHubCommitNotifier(String resultOnFailure) {
...
 
-    @Deprecated
-    public GitHubCommitNotifier() {
-        this(getDefaultResultOnFailure().toString());
+    public ExpandableMessage getStatusMessage() {
+        return statusMessage;
+    }
+
+    @DataBoundSetter
+    public void setStatusMessage(ExpandableMessage statusMessage) {
```
KostyaSha: `since`?

src/main/java/com/cloudbees/jenkins/GitHubCommitNotifier.java
```diff
@@ -109,38 +116,74 @@ public boolean perform(AbstractBuild<?, ?> build,
...
-                        sha1, state, build.getAbsoluteUrl(), msg, build.getProject().getFullName());
+                        sha1, status.getState(), build.getAbsoluteUrl(),
+                        message,
+                        contextName
+                );
             }
         }
     }
 
+    private static StatusResult statusFrom(@Nonnull AbstractBuild<?, ?> build) {
```
KostyaSha: since

KostyaSha: ah, private, ok.

KostyaSha:
Please put `since` for public methods.

src/main/java/org/jenkinsci/plugins/github/common/ExpandableMessage.java
oleg-nenashev: At least it allows to extract variables from other sources like global vars and environment contributors

KostyaSha: That's all covered in token-macro expandAll() afair, but in PR plugin i made 2 optional deps (token-macro and email-ext for templates). But seems @lanwen decided add hard dep on token-macro.

pom.xml
lanwen: Lot of plugins uses token macro. Don't want to maintain one more option and handle functionality of token-macro

src/main/java/com/cloudbees/jenkins/GitHubCommitNotifier.java
```diff
@@ -34,9 +42,16 @@
  */
 public class GitHubCommitNotifier extends Notifier {
 
+    private ExpandableMessage statusMessage = new ExpandableMessage("");
```
KostyaSha-auto: Provide default?

lanwen: default provided in perform method, as it have some logic

KostyaSha: perform is called only during execution. User UI is empty now as i see.

lanwen: yep, if this field is empty - will be used default

KostyaSha: That's wrong for user that see empty config.

lanwen: It under the advanced section and has help.

KostyaSha: @abayer ^ that's why you have empty UI for gh plugin

KostyaSha: So if all plugins do such things, i will have fully empty configuration that i should click on advanced/help buttons to get what is really will be during execution.

lanwen: This doesn't change default behaviour which exists all time.
I can implement token-macro with that logic, but its overhead doing it in this pr i think.
What is your propose?

KostyaSha: Provide some default messages in UI. When use adds step it should have some message so it can change it if he doesn't like it.

lanwen: it has 4 default messages. So it requires 4 fields with your logic. No, thanks

KostyaSha: User should see real configuration that would be applied, if it 4 fields, then 4.

lanwen: User can now define own configuration. The previous behaviour did not change.

andynemzek:
Hello, 
I've seen the documentation for how to issue a one-off custom status via a step, but I'm wondering if there is any way to set the status context such that all statuses reported by the plugin use the specified context. This PR seems to imply there is a way, but I have found no documentation for it.  Thanks!

PR #103, opened 2015-12-20
[FIXES JENKINS-32132] Check status message is null (in case of wrong …

lanwen:
…deserialization)

[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/103)


lanwen:
test will be in next pr

KostyaSha:
If you are making this property ready for null, then mark it with `@CheckForNull` but i would prefer to see `readResolve()` and `@Nonnull` field or getter.

KostyaSha:
> test will be in next pr

Sorry where is it?

lanwen:
this is a hotfix, so implementation may change after test will be ready (im working on gh mock)

PR #105, opened 2016-01-13
Show the "GitHub" link for Workflow projects

arthurschreiber:
This will make the "GitHub" Link show up on Workflow Jobs and other similar types (like Multibranch Workflow Jobs).

[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/105)

### DIFF
--- a/src/main/java/com/coravy/hudson/plugins/github/GithubLinkAction.java
+++ b/src/main/java/com/coravy/hudson/plugins/github/GithubLinkAction.java
@@ -1,6 +1,12 @@
 package com.coravy.hudson.plugins.github;
 
+import java.util.Collection;
+import java.util.Collections;
+
+import hudson.Extension;
 import hudson.model.Action;
+import hudson.model.Job;
+import jenkins.model.TransientActionFactory;
 
 /**
  * Add the Github Logo/Icon to the sidebar.
@@ -30,4 +36,23 @@ public final class GithubLinkAction implements Action {
         return projectProperty.getProjectUrl().baseUrl();
     }
 
+    @SuppressWarnings("rawtypes")
+    @Extension
+    public static class GithubLinkActionFactory extends TransientActionFactory<Job> {
+        @Override
+        public Class<Job> type() {
+            return Job.class;
+        }
+
+        @Override
+        public Collection<? extends Action> createFor(Job j) {
+            GithubProjectProperty prop = ((Job<?, ?>) j).getProperty(GithubProjectProperty.class);
+
+            if (prop == null) {
+                return Collections.emptySet();
+            } else {
+                return Collections.singleton(new GithubLinkAction(prop));
+            }
+        }
+    }
 }
--- a/src/main/java/com/coravy/hudson/plugins/github/GithubProjectProperty.java
+++ b/src/main/java/com/coravy/hudson/plugins/github/GithubProjectProperty.java
@@ -2,7 +2,6 @@ package com.coravy.hudson.plugins.github;
 
 import com.cloudbees.jenkins.GitHubPushTrigger;
 import hudson.Extension;
-import hudson.model.Action;
 import hudson.model.Job;
 import hudson.model.JobProperty;
 import hudson.model.JobPropertyDescriptor;
@@ -14,8 +13,6 @@ import org.kohsuke.stapler.StaplerRequest;
 
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
-import java.util.Collection;
-import java.util.Collections;
 import java.util.logging.Logger;
 
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
@@ -81,14 +78,6 @@ public final class GithubProjectProperty extends JobProperty<Job<?, ?>> {
     @DataBoundSetter
     public void setDisplayName(String displayName) {
         this.displayName = displayName;
-    }
-
-    @Override
-    public Collection<? extends Action> getJobActions(Job<?, ?> job) {
-        if (null != projectUrl) {
-            return Collections.singleton(new GithubLinkAction(this));
-        }
-        return Collections.emptyList();
     }
 
     /**
--- a/src/test/java/com/coravy/hudson/plugins/github/GithubLinkActionFactoryTest.java
+++ b/src/test/java/com/coravy/hudson/plugins/github/GithubLinkActionFactoryTest.java
@@ -1,0 +1,57 @@
+package com.coravy.hudson.plugins.github;
+
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.instanceOf;
+import static org.hamcrest.Matchers.empty;
+import static org.junit.Assert.assertThat;
+
+import java.io.IOException;
+import java.util.Collection;
+
+import org.jenkinsci.plugins.workflow.job.WorkflowJob;
+import org.junit.Rule;
+import org.junit.Test;
+import org.jvnet.hudson.test.JenkinsRule;
+
+import com.coravy.hudson.plugins.github.GithubLinkAction.GithubLinkActionFactory;
+
+import hudson.model.Action;
+
+public class GithubLinkActionFactoryTest {
+    @Rule
+    public final JenkinsRule rule = new JenkinsRule();
+
+    private final GithubLinkActionFactory factory = new GithubLinkActionFactory();
+
+    private static final String PROJECT_URL = "https://github.com/jenkinsci/github-plugin/";
+
+    private WorkflowJob createExampleJob() throws IOException {
+        return rule.getInstance().createProject(WorkflowJob.class, "example");
+    }
+
+    private GithubProjectProperty createExampleProperty() {
+        return new GithubProjectProperty(PROJECT_URL);
+    }
+
+    @Test
+    public void shouldCreateGithubLinkActionForJobWithGithubProjectProperty() throws IOException {
+        final WorkflowJob job = createExampleJob();
+        final GithubProjectProperty property = createExampleProperty();
+        job.addProperty(property);
+
+        final Collection<? extends Action> actions = factory.createFor(job);
+        assertThat("factored actions list", actions.size(), is(1));
+
+        final Action action = actions.iterator().next();
+        assertThat("instance check", action, is(instanceOf(GithubLinkAction.class)));
+        assertThat("url of action", action.getUrlName(), is(property.getProjectUrlStr()));
+    }
+
+    @Test
+    public void shouldNotCreateGithubLinkActionForJobWithoutGithubProjectProperty() throws IOException {
+        final WorkflowJob job = createExampleJob();
+
+        final Collection<? extends Action> actions = factory.createFor(job);
+        assertThat("factored actions list", actions, is(empty()));
+    }
+}


src/main/java/com/coravy/hudson/plugins/github/GithubLinkAction.java
```diff
@@ -30,4 +35,24 @@ public String getUrlName() {
...
+    public static class Factory extends TransientActionFactory<Job> {
+        @Override
+        public Class<Job> type() {
+            return Job.class;
+        }
+
+        @Override
+        public Collection<? extends Action> createFor(Job j) {
+            GithubProjectProperty prop =
+                    (GithubProjectProperty) ((Job<?, ?>) j).getProperty(GithubProjectProperty.class);
```
KostyaSha: Are you sure that you need cast it? Why job is also recasted?

arthurschreiber: I remember my IDE complaining here for both of these.

src/main/java/com/coravy/hudson/plugins/github/GithubLinkAction.java
```diff
@@ -30,4 +35,24 @@ public String getUrlName() {
...
+    @SuppressWarnings({"rawtypes"})
+    @Extension
+    public static class Factory extends TransientActionFactory<Job> {
+        @Override
+        public Class<Job> type() {
+            return Job.class;
+        }
+
+        @Override
+        public Collection<? extends Action> createFor(Job j) {
```
KostyaSha: Have you tried `(Job<?, ?> j)` here and remove casts?

src/main/java/com/coravy/hudson/plugins/github/GithubLinkAction.java
```diff
@@ -30,4 +35,24 @@ public String getUrlName() {
         return projectProperty.getProjectUrl().baseUrl();
     }
 
+    @SuppressWarnings({"rawtypes"})
+    @Extension
+    public static class Factory extends TransientActionFactory<Job> {
+        @Override
+        public Class<Job> type() {
+            return Job.class;
```
lanwen: maybe it should be `ParameterizedJobMixIn.ParameterizedJob.class` as already in property class descriptor?

src/main/java/com/coravy/hudson/plugins/github/GithubLinkAction.java
```diff
@@ -30,4 +35,24 @@ public String getUrlName() {
         return projectProperty.getProjectUrl().baseUrl();
     }
 
+    @SuppressWarnings({"rawtypes"})
+    @Extension
+    public static class Factory extends TransientActionFactory<Job> {
```
lanwen: Please name it `GithubLinkActionFactory` for better stacks in case of errors and debug

src/main/java/com/coravy/hudson/plugins/github/GithubLinkAction.java
lanwen: it doesn't work, as it should be same signature as in generic

src/main/java/com/coravy/hudson/plugins/github/GithubLinkAction.java
jglick: If you cast to `Job<?,?>` you should get a well-typed return value from `getProperty`. Or you can use `(GithubProjectProperty) j.getProperty(GithubProjectProperty.class)`.

src/main/java/com/coravy/hudson/plugins/github/GithubLinkAction.java
jglick: Right, due to Java erasure it is not possible to do better, except perhaps with more exotic type tokens.

src/main/java/com/coravy/hudson/plugins/github/GithubLinkAction.java
lanwen: found same impl in `jenkins.tasks.SimpleBuildStep.LastBuildActionFactory`, so ignore prev comment

src/main/java/com/coravy/hudson/plugins/github/GithubProjectProperty.java
```diff
@@ -2,11 +2,9 @@
 
 import com.cloudbees.jenkins.GitHubPushTrigger;
 import hudson.Extension;
-import hudson.model.Action;
 import hudson.model.Job;
 import hudson.model.JobProperty;
 import hudson.model.JobPropertyDescriptor;
-import jenkins.model.ParameterizedJobMixIn;
```
lanwen: please return it back

src/main/java/com/coravy/hudson/plugins/github/GithubLinkAction.java
```diff
@@ -8,7 +14,6 @@
  * @author Stefan Saasen <stefan@coravy.com>
  */
 public final class GithubLinkAction implements Action {
-
```
lanwen: please don't remove this line

src/main/java/com/coravy/hudson/plugins/github/GithubLinkAction.java
```diff
@@ -30,4 +35,24 @@ public String getUrlName() {
         return projectProperty.getProjectUrl().baseUrl();
     }
 
+    @SuppressWarnings({"rawtypes"})
```
lanwen: can be `("rawtypes")` for single arg

lanwen:
Have lost my comment o_O
Please also add simple unit test which will call `createFor` on project with/without property

src/main/java/com/coravy/hudson/plugins/github/GithubLinkAction.java
arthurschreiber: Fixed.

src/main/java/com/coravy/hudson/plugins/github/GithubLinkAction.java
arthurschreiber: Fixed.

src/main/java/com/coravy/hudson/plugins/github/GithubProjectProperty.java
arthurschreiber: Sorry, no idea why that was deleted.

src/main/java/com/coravy/hudson/plugins/github/GithubLinkAction.java
arthurschreiber: Fixed.

arthurschreiber:
@lanwen I fixed the code issues and will add unit tests asap.

arthurschreiber:
Hey @lanwen @jglick @KostyaSha 

I just updated this PR to include test cases. Please keep the feedback coming, I usually don't do Java development, so my knowledge of "Java best-practices" is spotty at best and I'm happy for any suggestions that you might have.

arthurschreiber:
@lanwen Do I need to resolve the comments pointed out by @dummy-lanwen-bot ?

I don't feel that adding custom error messages to these assertions is going to help a lot, but if that's something that is important, I can fix this up.

KostyaSha:
@arthurschreiber yes, describe something. After 1 year it will be very difficult understand what happened in test when you get this error from some integration system.

src/test/java/com/coravy/hudson/plugins/github/GithubLinkActionFactoryTest.java
```diff
@@ -0,0 +1,45 @@
...
+    @Test
+    public void shouldCreateGithubLinkActionForJobWithGithubProjectProperty() throws IOException {
+        WorkflowJob p = j.jenkins.createProject(WorkflowJob.class, "p");
+        p.addProperty(new GithubProjectProperty("https://github.com/jenkinsci/github-plugin/"));
+
+        Collection<? extends Action> actions = factory.createFor(p);
+        assertThat(actions.size(), is(equalTo(1)));
+
+        GithubLinkAction action = (GithubLinkAction) actions.iterator().next();
+        assertThat(action.getUrlName(), is(equalTo("https://github.com/jenkinsci/github-plugin/")));
```
lanwen: please add here comment "url name".
Also `is(equalTo())` is the same as simple `is()`

src/test/java/com/coravy/hudson/plugins/github/GithubLinkActionFactoryTest.java
```diff
@@ -0,0 +1,45 @@
...
+public class GithubLinkActionFactoryTest {
+    @Rule
+    public JenkinsRule j = new JenkinsRule();
+
+    public GithubLinkActionFactory factory = new GithubLinkActionFactory();
+
+    @Test
+    public void shouldCreateGithubLinkActionForJobWithGithubProjectProperty() throws IOException {
+        WorkflowJob p = j.jenkins.createProject(WorkflowJob.class, "p");
+        p.addProperty(new GithubProjectProperty("https://github.com/jenkinsci/github-plugin/"));
```
lanwen: move string to constant, to not duplicate it twice

src/test/java/com/coravy/hudson/plugins/github/GithubLinkActionFactoryTest.java
```diff
@@ -0,0 +1,45 @@
...
+import org.junit.Test;
+import org.jvnet.hudson.test.JenkinsRule;
+
+import com.coravy.hudson.plugins.github.GithubLinkAction.GithubLinkActionFactory;
+
+import hudson.model.Action;
+
+public class GithubLinkActionFactoryTest {
+    @Rule
+    public JenkinsRule j = new JenkinsRule();
```
lanwen: please use name `jRule` as of using one-letter var name is bad practise

src/test/java/com/coravy/hudson/plugins/github/GithubLinkActionFactoryTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+    public GithubLinkActionFactory factory = new GithubLinkActionFactory();
```
lanwen: it should be private as nobody should use it except this test

src/test/java/com/coravy/hudson/plugins/github/GithubLinkActionFactoryTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+    @Test
+    public void shouldCreateGithubLinkActionForJobWithGithubProjectProperty() throws IOException {
+        WorkflowJob p = j.jenkins.createProject(WorkflowJob.class, "p");
```
lanwen: use getter for instance `getInstance()` as of accessing to internal fields is not so stable as using defined getters

src/test/java/com/coravy/hudson/plugins/github/GithubLinkActionFactoryTest.java
KostyaSha: @lanwen all jenkins using it, that only your single preference.

KostyaSha: Also it bad practise to add types into variable names, i.e. `myCoolList = new ArrayList()`

lanwen: yep, i try to use best-practices in tests for this plugin

lanwen: `Rule` is not a type in this context - it is behaviour or role of this object.

src/test/java/com/coravy/hudson/plugins/github/GithubLinkActionFactoryTest.java
arthurschreiber: Fixed.

src/test/java/com/coravy/hudson/plugins/github/GithubLinkActionFactoryTest.java
arthurschreiber: Fixed.

src/test/java/com/coravy/hudson/plugins/github/GithubLinkActionFactoryTest.java
arthurschreiber: Fixed.

src/test/java/com/coravy/hudson/plugins/github/GithubLinkActionFactoryTest.java
arthurschreiber: I added some descriptions for all assertions.

arthurschreiber:
@KostyaSha I'm not new to unit testing, but I often feel there's cases where adding a custom description is ending up as being nothing more than boilerplate that no one ever needs.

As long as the checks that are done inside the test cases (1 assertion per test), the test names should be enough description. And if your tests end up using many assertions, a custom matcher with a custom error formatting might be more appropriate.

Anyway, I performed all the suggested changes, feel free to review again.

src/test/java/com/coravy/hudson/plugins/github/GithubLinkActionFactoryTest.java
```diff
@@ -0,0 +1,51 @@
...
+    private GithubLinkActionFactory factory = new GithubLinkActionFactory();
+
+    private WorkflowJob createExampleJob() throws IOException {
+        return rule.getInstance().createProject(WorkflowJob.class, "example");
+    }
+
+    @Test
+    public void shouldCreateGithubLinkActionForJobWithGithubProjectProperty() throws IOException {
+        WorkflowJob job = createExampleJob();
+        String url = "https://github.com/jenkinsci/github-plugin/";
```
lanwen: please move it to const 
`private static final String GH_PROJECT = "..."`

src/test/java/com/coravy/hudson/plugins/github/GithubLinkActionFactoryTest.java
```diff
@@ -0,0 +1,51 @@
...
+        job.addProperty(new GithubProjectProperty(url));
+
+        Collection<? extends Action> actions = factory.createFor(job);
+        assertThat("Only one Action should have been created", actions.size(), is(1));
+
+        Action action = actions.iterator().next();
+        assertThat("Created Action should be instance of GithubLinkAction", action, is(instanceOf(GithubLinkAction.class)));
```
lanwen: can be collapsed to `instance check`

src/test/java/com/coravy/hudson/plugins/github/GithubLinkActionFactoryTest.java
```diff
@@ -0,0 +1,51 @@
...
+        assertThat("Action URL should equal url set in GithubProjectProperty", action.getUrlName(), is(url));
```
lanwen: can be collapsed to `url of action`

src/test/java/com/coravy/hudson/plugins/github/GithubLinkActionFactoryTest.java
arthurschreiber: Sorry to ask, but what would be the benefit of that? This value is not used in any other test case, so making it a property on the class does not make me feel like it has any obvious benefits.

Again, I'm no Java Expert, so if there's benefits to this, feel free to let me know. Learning new things is always good. :smile:

src/test/java/com/coravy/hudson/plugins/github/GithubLinkActionFactoryTest.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    @Test
+    public void shouldCreateGithubLinkActionForJobWithGithubProjectProperty() throws IOException {
+        WorkflowJob job = createExampleJob();
+        String url = "https://github.com/jenkinsci/github-plugin/";
+        job.addProperty(new GithubProjectProperty(url));
+
+        Collection<? extends Action> actions = factory.createFor(job);
+        assertThat("Only one Action should have been created", actions.size(), is(1));
```
lanwen: `factored actions list`

src/test/java/com/coravy/hudson/plugins/github/GithubLinkActionFactoryTest.java
lanwen: its not the test value itself, its just a some data which should remain the same in this test.
Vars should be used if they make some active role in the test
http://xunitpatterns.com/, page 192 (Irrelevant information)

lanwen:
@arthurschreiber u're right for 1 assertion for one test and custom matchers :heart: 
your test names is very good, so the assertion comment can bi omitted in cases where the matcher + test title says the cause of error. In case of more than one assertion in case, comment should help to define context of error. So it can explain the cause of this check, or simply point the place and the subject of checks.

I've commented the examples of such comments

arthurschreiber:
I'll perform your suggested changes (you are maintaining the project, overall). :heart: 

> So it can explain the cause of this check, or simply point the place and the subject of checks

Yeah, I feel that using plain `assert` method itself without any description is a big NO-NO. For specific matchers (like `instanceOf`), I usually feel that adding a custom description does usually not provide much, the matcher already has very specific error messages, and the stack trace provides line numbers anyway.

Did I miss anything?

lanwen:
only const for url now. Add it, and i merge and release

src/test/java/com/coravy/hudson/plugins/github/GithubLinkActionFactoryTest.java
arthurschreiber: Thanks for the explanation! :heart:

arthurschreiber:
@lanwen Can you check again? :+1:

src/test/java/com/coravy/hudson/plugins/github/GithubLinkActionFactoryTest.java
```diff
@@ -0,0 +1,57 @@
...
+
+    private final GithubLinkActionFactory factory = new GithubLinkActionFactory();
+
+    private static final String PROJECT_URL = "https://github.com/jenkinsci/github-plugin/";
+
+    private WorkflowJob createExampleJob() throws IOException {
+        return rule.getInstance().createProject(WorkflowJob.class, "example");
+    }
+
+    private GithubProjectProperty createExampleProperty() {
```
lanwen: could be inlined, but nvm

arthurschreiber:
Thank you so much for putting up with my lack of Java knowledge. And thanks for merging this and releasing it so quickly. :smile:

lanwen:
thanks for PR and quick review fixes

PR #108, opened 2016-01-28
[FIXED JENKINS-32666] Inject the Payload into the build

fbelzunc:
It might be interesting to inject the GitHub payload received into the build so it can be treated at build level.

The blog below explains what is the interest of injecting this information at build level.
* http://chloky.com/github-pull-req-webhook/
* http://chloky.com/github-json-payload-in-jenkins/

https://issues.jenkins-ci.org/browse/JENKINS-32666

@reviewbybees

[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/108)

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/GitHubPayload.java
+++ b/src/main/java/com/cloudbees/jenkins/GitHubPayload.java
@@ -1,0 +1,34 @@
+package com.cloudbees.jenkins;
+
+import hudson.EnvVars;
+import hudson.model.AbstractBuild;
+import hudson.model.EnvironmentContributingAction;
+import hudson.model.InvisibleAction;
+
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * Inject the payload received by GitHub into the build through $GITHUB_PAYLOAD so it can be processed
+ * @since January 28, 2016
+ * @version 1.17.1
+ */
+public class GitHubPayload extends InvisibleAction implements EnvironmentContributingAction {
+    private final String payload;
+
+    public GitHubPayload(String payload) {
+        this.payload = payload;
+    }
+
+    public String getPayload() {
+        return payload;
+    }
+
+    @Override
+    public void buildEnvVars(AbstractBuild<?, ?> abstractBuild, EnvVars envVars) {
+        LOGGER.log(Level.FINEST, "Injecting GITHUB_PAYLOAD: {0}", payload);
+        envVars.put("GITHUB_PAYLOAD", payload);
+    }
+
+    private static final Logger LOGGER = Logger.getLogger(GitHubPayload.class.getName());
+}
--- a/src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
+++ b/src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
@@ -8,6 +8,7 @@ import hudson.XmlFile;
 import hudson.console.AnnotatedLargeText;
 import hudson.model.AbstractProject;
 import hudson.model.Action;
+import hudson.model.CauseAction;
 import hudson.model.Item;
 import hudson.model.Job;
 import hudson.model.Project;
@@ -46,7 +47,6 @@ import java.util.Set;
 
 import static org.apache.commons.lang3.StringUtils.isEmpty;
 import static org.jenkinsci.plugins.github.Messages.github_trigger_check_method_warning_details;
-import static org.jenkinsci.plugins.github.util.JobInfoHelpers.asParameterizedJobMixIn;
 
 /**
  * Triggers a build when we receive a GitHub post-commit webhook.
@@ -70,7 +70,15 @@ public class GitHubPushTrigger extends Trigger<Job<?, ?>> implements GitHubTrigg
     /**
      * Called when a POST is made.
      */
+    @Deprecated
     public void onPost(String triggeredByUser) {
+        onPost(triggeredByUser, "");
+    }
+
+    /**
+     * Called when a POST is made.
+     */
+    public void onPost(String triggeredByUser, final String payload) {
         final String pushBy = triggeredByUser;
         getDescriptor().queue.execute(new Runnable() {
             private boolean runPolling() {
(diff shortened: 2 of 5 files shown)


KostyaSha:
:-1: this trigger implementation based on git polling would never match payload content vs triggered build.
You will have unrelated to build/branch injected payload that useless.

fbelzunc:
@KostyaSha I don't understand your point. 

Would you mind to elaborate a bit more so I can get why it is useless, please?

KostyaSha:
http://chloky.com/github-json-payload-in-jenkins/ that's absolutely different logic of triggering build that will work right according to it's implementation.

jtnord:
This will likely break a lot of builds.

Maven Ant and IIRC Gradle tasks take envVars and blindly inject them to the build as `-Dvar=value` which causes all sorts of havoc (see [JENKINS-25416](https://issues.jenkins-ci.org/browse/JENKINS-25416) for the tip of the iceberg.

Therefore I would be :bug: on this unless you can opt-out (which core can not yet do - so this needs to be covered by this PR - and I did not see the option to do this.

KostyaSha:
For those who has payed time but doesn't learn code:
GH sends payload, jenkins searches for interested projects and kicks git polling, if it return "YES THERE IS NEW SOMETHING IN SCM POLLING" it schedule build. After this, git may do build absolutely for other commit. You can follow stupid logic in git plugin https://issues.jenkins-ci.org/browse/JENKINS-30345 https://issues.jenkins-ci.org/browse/JENKINS-18079 
That's all makes me think that you never used your changes on real installations with real use cases. 

This trigger can't be changed to any other way of work. Because this plugin is generic - there is no sense in adding any other triggers to it. That's why i'm working on other triggers implementations in https://github.com/jenkinsci/github-integration-plugin

If you learn the code and found that i missed something and wrong with algorithms - feel free to reopen.

KostyaSha:
In 2012 year document that you found http://chloky.com/github-json-payload-in-jenkins/ build trigger gets EXACT payload for exact build/commit.

fbelzunc:
@KostyaSha 

> For those who has payed time but doesn't learn code:

I just think you are rude and unpolite - this is just my opinion. I don't think it is the way to treat people. *I don't care at all, what you know, and what you don't know*  if you don't know how to treat people contributing. You are thinking I do all on my business time and not on my spare time, but this is not correct at all.

People like you is what makes others to go away from Open Source projects like this one. You might want to learn how to work with others or be more polite and better not to reply if you are going to do it on this way.

Thank you anyway for the full explanation at the end.

KostyaSha:
@fbelzunc feel free to address anything personal directly to my emails.

KostyaSha:
@fbelzunc do you want to say that this commit is done in your personal time? I can excuse then. Do you want to say that you learn code and algorithm?
However i spent half hour today and half hour yesterday (for person that didn't read the docs) on CB reports and i can allow such sentence that was just what i think about changes that i see. I want spent time on opensource for opensource people with their real life cases.

btw, cc @abayer

KostyaSha:
@fbelzunc ok, if you describe how you used this modification and what use case do you want to solve - i will excuse for thing that i don't even think rude.

KostyaSha:
Let's clean this thread to keep technical side. I can remove sentence that touched person.

tfennelly:
> i will excuse

No, a better idea may be for you to __apologise__. Then, maybe @fbelzunc can __excuse__ you.

I think you know well when you are being rude (and everyone else does too). I don't understand why you make it personal. Your technical points are nearly always great, but then you go and ruin it by throwing in a personal insult. It makes no sense at all.

KostyaSha:
Cleaned. Next time please send to me directly will be glad to talk.

KostyaSha:
Let's back to technical side.
@tfennelly @fbelzunc please provide use case/project for this feature or goal that you try to solve.

tfennelly:
@KostyaSha totally bad form deleting other people's posts !!!

KostyaSha:
@tfennelly totally bad polluting PRs with such things. I didn't plan insult somebody. Please fill email for any personal things and discuss your and mine cultural views.
If this thread will continue with non technical things, i will lock it.

KostyaSha:
Also, https://github.com/jenkinsci/github-plugin/blob/master/CONTRIBUTING.md

fbelzunc:
Well, I thought that since before my change in `GitHubPushTrigger#run` was scheduling a build with:

```
asParameterizedJobMixIn(job).scheduleBuild(cause)
```
and `cause` is basically getting the `pushBy` from `onPost`, it will then be tied to the right build. However, now I understand that it only really says that the build was push by ... and not that the only commit which contains is the one from pushby.

> In 2012 year document that you found http://chloky.com/github-json-payload-in-jenkins/ build trigger gets EXACT payload for exact build/commit.

This is right, this example doesn't match exactly the pull request.

@KostyaSha Since this pull request doesn't really match the use case I wanted to implement, I am more than fine with your decision of closing the pull request. One thing that might work, is only to inject the payload in case there is only one commit in the change, but anyway it will not be then a clean use-case. 

@KostyaSha Thanks for your time and apologize if my lack ok knowledge about of the SCM made you loose your time - Everyday we learn new things, don't we?

Have a good day!

KostyaSha:
@fbelzunc even if your injection breaks plugins mentioned @jtnord i'm :+1: with you. It was bad idea on injecting everything into build without having options. But on other hand having many env variables is also bad idea. Also saving json request makes sense only for `light` (i call them) or `transient` when jenkins kicks the build would work only for limited number of cases i.e. ghprb-plugin and github-pullrequest-plugin providing more functionality like searching for special comments in PR that wouldn't work with payload. So you should query GH api back to get additional info and here there is no sense attaching anything at all. Inserting many vars is also bad. So i can suggest using groovy to access: either stored some info in action, either query GH api to get what you need for build. But then you will have time related issues (no atomic actions around GH api).

KostyaSha:
`asParameterizedJobMixIn(job).scheduleBuild(cause)` would also lie about next build number. It can match only because git schedule builds one-by-one and BuildData https://github.com/jenkinsci/git-plugin/pull/313 is always picked from latest build. But git plugin also provides specific features like (exclude commits for certain paths) that requires workspace.

That's why i'm in process of writing additional push/branch trigger that will work natively with GH hooks/API and trigger builds for exact expected actions.

PR #109, opened 2016-01-28
Fixed incorrect proxy configuration of GitHubLoginFunction

schulzh:
The [Jenkins ProxyConfiguration class](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/ProxyConfiguration.java) expects only the hostname (e.g. `git.example.com` vs `https://git.example.com/api/`) as a method parameter to `createProxy`, but the existing code gives it the whole URL.
As a result of this, the validation of no proxy hosts is not correct. In most cases this is not a problem, but it is for some enterprise GitHub instances.

The handling of the possible MalformedURLException might not be the most suitable, it may needs to be replaced.

[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/109)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/github/internal/GitHubLoginFunction.java
+++ b/src/main/java/org/jenkinsci/plugins/github/internal/GitHubLoginFunction.java
@@ -20,6 +20,7 @@ import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 import java.io.IOException;
 import java.net.HttpURLConnection;
+import java.net.MalformedURLException;
 import java.net.Proxy;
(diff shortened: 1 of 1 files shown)


KostyaSha:
Wouldn't be better made matching in core?

schulzh:
I thought of changing this in the core aswell, but it explicitly says `host` as the method-parameter so strictly speaking the code in the plugin is wrong.

KostyaSha:
Then :+1: , let's wait for @lanwen review

lanwen:
Thanks for the fix!

PR #110, opened 2016-02-19
Handle ssh:// URLs that doesn't contain git@

kad:
It is possible to specify and use github.com URLs as
`ssh://github.com/username/repo`, as credentials used
for SCM configuration usually include username.
This change would allow to properly parse such URLs.

[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/110)


KostyaSha:
LGTM, @lanwen ?

kad:
After second look, it would be good to update also test: github-plugin/src/test/java/com/coravy/hudson/plugins/github/GitHubRepositoryNameTest.java. I'll update PR in few hours.

KostyaSha:
Yes, text case required.

kad:
Test case updated.

lanwen:
thanks!

lanwen:
will be released as 1.17.1 soon

KostyaSha:
:+1:

PR #111, opened 2016-03-07
Don't let the build fail if the commit doesn't exist in upstream

kohsuke:
PR builds and other merge activities can create a new tip of the repository that's not in the github repository. This plugin causes the build to fail in such a case today, which is undesirable. We'd rather just have the build to succeed without a commit status.

This is the stack trace you get when that happens:

````
ERROR: Step ‘Set build status on GitHub commit’ aborted due to exception: 
java.io.FileNotFoundException: {"message":"No commit found for SHA: bb077596591672c8348ace0f8ddb2a13df2e0892","documentation_url":"https://developer.github.com/v3/repos/statuses/"}
	at org.kohsuke.github.Requester.handleApiError(Requester.java:527)
	at org.kohsuke.github.Requester._to(Requester.java:257)
	at org.kohsuke.github.Requester.to(Requester.java:203)
	at org.kohsuke.github.GHRepository.createCommitStatus(GHRepository.java:854)
	at com.cloudbees.jenkins.GitHubCommitNotifier.updateCommitStatus(GitHubCommitNotifier.java:142)
	at com.cloudbees.jenkins.GitHubCommitNotifier.perform(GitHubCommitNotifier.java:111)
	at hudson.tasks.BuildStepCompatibilityLayer.perform(BuildStepCompatibilityLayer.java:75)
	at hudson.tasks.BuildStepMonitor$1.perform(BuildStepMonitor.java:20)
	at hudson.model.AbstractBuild$AbstractBuildExecution.perform(AbstractBuild.java:795)
	at hudson.model.AbstractBuild$AbstractBuildExecution.performAllBuildSteps(AbstractBuild.java:736)
	at hudson.model.Build$BuildExecution.post2(Build.java:185)
	at hudson.model.AbstractBuild$AbstractBuildExecution.post(AbstractBuild.java:681)
	at hudson.model.Run.execute(Run.java:1766)
	at hudson.model.FreeStyleBuild.run(FreeStyleBuild.java:43)
	at hudson.model.ResourceController.execute(ResourceController.java:98)
	at hudson.model.Executor.run(Executor.java:408)
Caused by: java.io.FileNotFoundException: https://api.github.com/repos/cloudbees/blueocean/statuses/bb077596591672c8348ace0f8ddb2a13df2e0892
	at com.squareup.okhttp.internal.huc.HttpURLConnectionImpl.getInputStream(HttpURLConnectionImpl.java:240)
	at com.squareup.okhttp.internal.huc.DelegatingHttpsURLConnection.getInputStream(DelegatingHttpsURLConnection.java:210)
	at com.squareup.okhttp.internal.huc.HttpsURLConnectionImpl.getInputStream(HttpsURLConnectionImpl.java:25)
	at org.kohsuke.github.Requester.parse(Requester.java:483)
	at org.kohsuke.github.Requester._to(Requester.java:236)
	... 14 more
````

[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/111)

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/GitHubCommitNotifier.java
+++ b/src/main/java/com/cloudbees/jenkins/GitHubCommitNotifier.java
@@ -14,7 +14,6 @@ import hudson.tasks.Notifier;
 import hudson.tasks.Publisher;
 import hudson.util.ListBoxModel;
 import jenkins.tasks.SimpleBuildStep;
-
 import org.eclipse.jgit.lib.ObjectId;
 import org.jenkinsci.plugins.github.common.ExpandableMessage;
 import org.jenkinsci.plugins.github.util.BuildDataHelper;
@@ -55,7 +54,7 @@ public class GitHubCommitNotifier extends Notifier implements SimpleBuildStep {
     private final String resultOnFailure;
     private static final Result[] SUPPORTED_RESULTS = {FAILURE, UNSTABLE, SUCCESS};
 
-    private static final Logger LOGGER = LoggerFactory.getLogger(GitHubWebHook.class);
+    private static final Logger LOGGER = LoggerFactory.getLogger(GitHubCommitNotifier.class);
 
     @Restricted(NoExternalUse.class)
     public GitHubCommitNotifier() {
@@ -109,9 +108,9 @@ public class GitHubCommitNotifier extends Notifier implements SimpleBuildStep {
 
     @Override
     public void perform(Run<?, ?> build,
-                           FilePath ws,
-                           Launcher launcher,
-                           TaskListener listener) throws InterruptedException, IOException {
+                        FilePath ws,
+                        Launcher launcher,
+                        TaskListener listener) throws InterruptedException, IOException {
         try {
             updateCommitStatus(build, listener);
         } catch (IOException error) {
@@ -155,8 +154,8 @@ public class GitHubCommitNotifier extends Notifier implements SimpleBuildStep {
                     // doesn't exist in the upstream. Don't let the build fail
                     // TODO: ideally we'd like other plugins to designate a commit to put the status update to
                     LOGGER.debug("Failed to update commit status", e);
-                    listener.getLogger().println("Commit doesn't exist in "
-                            + repository.getFullName() + ". Status is not set");
+                    listener.getLogger()
+                            .format("Commit doesn't exist in %s. Status is not set%n", repository.getFullName());
                 }
             }
         }
--- a/src/main/java/com/cloudbees/jenkins/GitHubCommitNotifier.java
+++ b/src/main/java/com/cloudbees/jenkins/GitHubCommitNotifier.java
@@ -24,8 +24,11 @@ import org.kohsuke.github.GHCommitState;
 import org.kohsuke.github.GHRepository;
 import org.kohsuke.stapler.DataBoundConstructor;
 import org.kohsuke.stapler.DataBoundSetter;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nonnull;
+import java.io.FileNotFoundException;
 import java.io.IOException;
 
 import static com.cloudbees.jenkins.Messages.GitHubCommitNotifier_DisplayName;
@@ -51,6 +54,8 @@ public class GitHubCommitNotifier extends Notifier implements SimpleBuildStep {
 
     private final String resultOnFailure;
     private static final Result[] SUPPORTED_RESULTS = {FAILURE, UNSTABLE, SUCCESS};
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(GitHubWebHook.class);
 
     @Restricted(NoExternalUse.class)
     public GitHubCommitNotifier() {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/main/java/com/cloudbees/jenkins/GitHubCommitNotifier.java [hunk 1]
```diff
@@ -139,11 +140,19 @@ private void updateCommitStatus(@Nonnull Run<?, ?> build,
...
+                    repository.createCommitStatus(
+                            sha1, status.getState(), build.getAbsoluteUrl(),
+                            message,
+                            contextName
+                    );
+                } catch (FileNotFoundException e) {
+                    // PR builds and other merge activities can create a merge commit that
+                    // doesn't exist in the upstream. Don't let the build fail
+                    // TODO: ideally we'd like other plugins to designate a commit to put the status update to
+                    listener.getLogger().println("Commit doesn't exist in " +
```
michaelneale: `+` on newline to pass checkstyle

michaelneale:
:bee:  one checkstyle violation is fixed.

KostyaSha:
1) no associated jira issue
2) This changes the strategy of publisher that is totally wrong. Such error can appear because of wrong permissions and then it should fail. If you want special handling, then introduce error handler i.e. like https://github.com/jenkinsci/github-integration-plugin/blob/master/github-pullrequest-plugin/src/main/java/org/jenkinsci/plugins/github/pullrequest/utils/PublisherErrorHandler.java 
:-1:

kohsuke:
Note that I'm only catching `FileNotFoundException` so that it can differentiate other configuration errors from "no such commit" error

KostyaSha:
What error is used when user has no permissions to repository at all?

kohsuke:
Generic `IOExcception` is used for anything but 404 in github-api

kohsuke:
I created [JENKINS-33371](https://issues.jenkins-ci.org/browse/JENKINS-33371) as per your request

src/main/java/com/cloudbees/jenkins/GitHubCommitNotifier.java
```diff
@@ -52,6 +55,8 @@
     private final String resultOnFailure;
     private static final Result[] SUPPORTED_RESULTS = {FAILURE, UNSTABLE, SUCCESS};
 
+    private static final Logger LOGGER = Logger.getLogger(GitHubCommitNotifier.class.getName());
```
KostyaSha: use org.slf4j.Logger

src/main/java/com/cloudbees/jenkins/GitHubCommitNotifier.java
```diff
@@ -52,6 +56,8 @@
     private final String resultOnFailure;
     private static final Result[] SUPPORTED_RESULTS = {FAILURE, UNSTABLE, SUCCESS};
 
+    private static final Logger LOGGER = LoggerFactory.getLogger(GitHubWebHook.class);
```
lanwen: please use `GitHubCommitNotifier` class

KostyaSha: Interesting, would PMD be able to pick such errors?

lanwen:
think its ok to handle exception with message and not fail the build.
Please remove `UnusedImports: Unused import - java.util.logging.Level.` and change logger class and ill will merge it

KostyaSha:
> think its ok to handle exception with message and not fail the build.

Not really. 
1) First question would be where notifier extracting commit. If it come from push notification, then it may be not correct and then github-plugin should be fixed at all.
 -- Builder can set Pending, then publisher fail and GH will have wrong status. 
2) This is the strategy: fail on errors, skip errors silently. It should be configurable. Either people will PR changes every time or would report bugs for this traces.
3) Seems Builder forgotten again.

KostyaSha:
> 1) First question would be where notifier extracting commit. If it come from push notification, then it may be not correct and then github-plugin should be fixed at all.

That's `org.jenkinsci.plugins.github.util.BuildDataHelper#getCommitSHA1` that extracts commit from `BuildData`. AFAIR BuildData stores sha1 for commit that was defined in checkout section. So any GIT scm features shouldn't change initial commit. AFAIR git SCM design (and BuildData?) has real and effective hashes. 

Push trigger fully relies on git scm and git scm should provide correct hashes. Would be glad to hear real example and hash changes situation for git scm.

KostyaSha:
`hudson.plugins.git.util.Build#marked` should be right commit, right?

lanwen:
This builder should be splitted to 3 parts - the hash finder, status and message logic and the result handler (of plugin itself), so now build-safe behaviour can be applied

KostyaSha:
@kohsuke please try change BuilDataHelper to use
```
---        final Revision lastBuildRevision = buildData.getLastBuiltRevision();
+++        final Revision lastBuildRevision = buildData.lastBuild.marked;
```

KostyaSha:
@lanwen absent commit is bug or force push. Force push of course should fail the build as commit doesn't exist anymore and for example pushing tags after it makes no sense.

kohsuke:
Added what Kostya wanted in another commit and squashed the original one.

KostyaSha:
@kohsuke does that work? Please remove try then as it would be bug in git plugins. GH commit kicks build and it should exist, either build should be failed.

KostyaSha:
cc @oleg-nenashev as BuildDataHelper seems his class

KostyaSha:
If try would be removed back, then :+1: as i think it was initial bug on using wrong commits because BuilData design is to crazy for understanding.

src/test/java/com/cloudbees/jenkins/GitHubCommitNotifierTest.java
```diff
@@ -73,6 +73,7 @@
         @Override
         protected void before() throws Throwable {
             when(data.getLastBuiltRevision()).thenReturn(rev);
+            data.lastBuild = new hudson.plugins.git.util.Build(rev,rev,0,Result.SUCCESS);
```
KostyaSha: please use imports

KostyaSha: and spaces (even if checkstyle doesn't cover test classes).

lanwen:
Please don't forget about https://github.com/jenkinsci/github-plugin/pull/111/files#diff-752cdecf02b12bb22c1fe3285f4d99a5R58

michaelneale:
bump - I guess the remaining changes are to fix the import - is that all? 
(whilst this won't fail the build, it won;t correctly set the status of the actual commit what was merge validated).

lanwen:
import + logger. Will fix that on the weekend by myself

PR #112, opened 2016-03-08
[JENKINS-33228] Misc global config page improvements

kohsuke:
See [JENKINS-33228](https://issues.jenkins-ci.org/browse/JENKINS-33228) for the context of this discussion.

This PR consists of a number of small changes to slightly improve the global configuration page.

`GitHubServerConfig` and especially its help text is still not really adequate for reuse by other plugins, but those requires further work.

[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/112)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/github/config/GitHubPluginConfig.java
+++ b/src/main/java/org/jenkinsci/plugins/github/config/GitHubPluginConfig.java
@@ -171,7 +171,7 @@ public class GitHubPluginConfig extends GlobalConfiguration {
 
     @Override
     public String getDisplayName() {
-        return "GitHub Plugin Configuration";
+        return "GitHub Configuration";
     }
 
     @SuppressWarnings("unused")
--- a/src/main/java/org/jenkinsci/plugins/github/config/GitHubPluginConfig.java
+++ b/src/main/java/org/jenkinsci/plugins/github/config/GitHubPluginConfig.java
@@ -171,7 +171,7 @@ public class GitHubPluginConfig extends GlobalConfiguration {
 
     @Override
     public String getDisplayName() {
-        return "GitHub Configuration";
+        return "GitHub";
     }
 
     @SuppressWarnings("unused")
--- a/src/main/java/org/jenkinsci/plugins/github/config/GitHubServerConfig.java
+++ b/src/main/java/org/jenkinsci/plugins/github/config/GitHubServerConfig.java
@@ -261,7 +261,7 @@ public class GitHubServerConfig extends AbstractDescribableImpl<GitHubServerConf
 
         @Override
         public String getDisplayName() {
-            return "GitHub Server Config";
+            return "GitHub Server";
         }
 
         @SuppressWarnings("unused")
--- a/src/main/java/org/jenkinsci/plugins/github/config/GitHubServerConfig.java
+++ b/src/main/java/org/jenkinsci/plugins/github/config/GitHubServerConfig.java
@@ -83,11 +83,6 @@ public class GitHubServerConfig extends AbstractDescribableImpl<GitHubServerConf
     private int clientCacheSize = DEFAULT_CLIENT_CACHE_SIZE_MB;
 
     /**
-     * only to set to default apiUrl when uncheck. Can be removed if optional block can nullify value if unchecked
-     */
-    private transient boolean customApiUrl;
-
-    /**
      * To avoid creation of new one on every login with this config
      */
     private transient GitHub cachedClient;
@@ -98,28 +93,26 @@ public class GitHubServerConfig extends AbstractDescribableImpl<GitHubServerConf
     }
 
     /**
-     * {@link #customApiUrl} field should be defined earlier. Because of we get full content of optional block,
-     * even if it already unchecked. So if we want to return api url to default value - custom value should affect
+     * Set the API endpoint.
      *
      * @param apiUrl custom url if GH. Default value will be used in case of custom is unchecked or value is blank
      */
     @DataBoundSetter
     public void setApiUrl(String apiUrl) {
-        if (customApiUrl) {
-            this.apiUrl = defaultIfBlank(apiUrl, GITHUB_URL);
-        } else {
-            this.apiUrl = GITHUB_URL;
-        }
+        this.apiUrl = defaultIfBlank(apiUrl, GITHUB_URL);
     }
 
     /**
-     * Should be called before {@link #setApiUrl(String)}
      *
      * @param customApiUrl true if optional block "Custom GH Api Url" checked in UI
+     * @deprecated
+     *      just set {@link #setApiUrl(String)}
      */
     @DataBoundSetter
     public void setCustomApiUrl(boolean customApiUrl) {
-        this.customApiUrl = customApiUrl;
+        if (!customApiUrl) {
+            this.apiUrl = GITHUB_URL;
+        }
     }
 
     /**
@@ -138,6 +131,7 @@ public class GitHubServerConfig extends AbstractDescribableImpl<GitHubServerConf
 
     /**
      * @see #isUrlCustom(String)
+     * @deprecated
      */
     @Restricted(NoExternalUse.class)
     public boolean isCustomApiUrl() {
@@ -291,7 +285,6 @@ public class GitHubServerConfig extends AbstractDescribableImpl<GitHubServerConf
                 @QueryParameter Integer clientCacheSize) throws IOException {
 
             GitHubServerConfig config = new GitHubServerConfig(credentialsId);
-            config.setCustomApiUrl(isUrlCustom(apiUrl));
             config.setApiUrl(apiUrl);
             config.setClientCacheSize(clientCacheSize);
             GitHub gitHub = new GitHubLoginFunction().apply(config);
--- a/src/main/java/org/jenkinsci/plugins/github/config/GitHubServerConfig.java
+++ b/src/main/java/org/jenkinsci/plugins/github/config/GitHubServerConfig.java
@@ -103,19 +103,6 @@ public class GitHubServerConfig extends AbstractDescribableImpl<GitHubServerConf
     }
 
     /**
-     *
-     * @param customApiUrl true if optional block "Custom GH Api Url" checked in UI
-     * @deprecated
-     *      just set {@link #setApiUrl(String)}
-     */
-    @DataBoundSetter
-    public void setCustomApiUrl(boolean customApiUrl) {
-        if (!customApiUrl) {
-            this.apiUrl = GITHUB_URL;
-        }
-    }
-
-    /**
      * This server config will be used to manage GH Hooks if true
      *
      * @param manageHooks false to ignore this config on hook auto-management
@@ -127,15 +114,6 @@ public class GitHubServerConfig extends AbstractDescribableImpl<GitHubServerConf
 
     public String getApiUrl() {
         return apiUrl;
-    }
-
-    /**
-     * @see #isUrlCustom(String)
-     * @deprecated
-     */
-    @Restricted(NoExternalUse.class)
-    public boolean isCustomApiUrl() {
-        return isUrlCustom(apiUrl);
     }
 
     public boolean isManageHooks() {
--- a/src/main/java/org/jenkinsci/plugins/github/config/GitHubServerConfig.java
+++ b/src/main/java/org/jenkinsci/plugins/github/config/GitHubServerConfig.java
@@ -19,8 +19,6 @@ import org.jenkinsci.plugins.github.util.misc.NullSafeFunction;
 import org.jenkinsci.plugins.github.util.misc.NullSafePredicate;
 import org.jenkinsci.plugins.plaincredentials.StringCredentials;
 import org.jenkinsci.plugins.plaincredentials.impl.StringCredentialsImpl;
-import org.kohsuke.accmod.Restricted;
-import org.kohsuke.accmod.restrictions.NoExternalUse;
 import org.kohsuke.github.GitHub;
 import org.kohsuke.stapler.DataBoundConstructor;
 import org.kohsuke.stapler.DataBoundSetter;
--- a/src/main/java/org/jenkinsci/plugins/github/config/GitHubServerConfig.java
+++ b/src/main/java/org/jenkinsci/plugins/github/config/GitHubServerConfig.java
@@ -281,12 +281,11 @@ public class GitHubServerConfig extends AbstractDescribableImpl<GitHubServerConf
         @SuppressWarnings("unused")
         public FormValidation doVerifyCredentials(
                 @QueryParameter String apiUrl,
-                @QueryParameter String credentialsId,
-                @QueryParameter Integer clientCacheSize) throws IOException {
+                @QueryParameter String credentialsId) throws IOException {
 
             GitHubServerConfig config = new GitHubServerConfig(credentialsId);
             config.setApiUrl(apiUrl);
-            config.setClientCacheSize(clientCacheSize);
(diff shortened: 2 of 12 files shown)


kohsuke:
@reviewbybees

src/main/java/org/jenkinsci/plugins/github/config/GitHubPluginConfig.java
```diff
@@ -171,7 +171,7 @@ public boolean configure(StaplerRequest req, JSONObject json) throws FormExcepti
 
     @Override
     public String getDisplayName() {
-        return "GitHub Plugin Configuration";
+        return "GitHub Configuration";
```
KostyaSha: :-1:

KostyaSha:
@kohsuke please do PRs from your forked repository, jenkinsci is full of your stalled branches (example https://github.com/jenkinsci/jenkins/branches/stale)

Also for UI improvements please attach screenshot before and after.

This changes seems what i said before to @lanwen, but he decided to do target for easier default UI. Let's wait his review.

src/main/java/org/jenkinsci/plugins/github/config/GitHubPluginConfig.java
KostyaSha: And yes, i know about help buttons, but they may be broken by core like in different plugins today.

src/test/java/org/jenkinsci/plugins/github/internal/GitHubClientCacheCleanupTest.java
```diff
@@ -103,7 +102,10 @@ private void it(String comment, int count) throws IOException {
     }
 
     private void makeCachedRequestWithCredsId(String credsId) throws IOException {
-        jRule.getInstance().getDescriptorByType(GitHubServerConfig.DescriptorImpl.class)
-                .doVerifyCredentials(github.serverConfig().getApiUrl(), credsId, 1);
+        GitHubServerConfig config = new GitHubServerConfig(credsId);
+        config.setApiUrl(github.serverConfig().getApiUrl());
+        config.setClientCacheSize(1);
```
KostyaSha: @kohsuke maybe it would be easy for you to implement this https://issues.jenkins-ci.org/browse/JENKINS-32503 ?

src/main/java/org/jenkinsci/plugins/github/config/GitHubServerConfig.java
```diff
@@ -138,6 +131,7 @@ public String getApiUrl() {
 
     /**
      * @see #isUrlCustom(String)
+     * @deprecated
      */
     @Restricted(NoExternalUse.class)
     public boolean isCustomApiUrl() {
```
KostyaSha: Just remove if it not used anymore, it was only for UI checkboxes because of pure jelly forms.

jglick:
> please do PRs from your forked repository

Agreed, origin PRs should be reserved for cases where collaboration is expected.

src/main/java/org/jenkinsci/plugins/github/config/GitHubPluginConfig.java
jglick: > they may be broken by core

What is this referring to?

jglick: > it [is] extremely critical to know what plugin contributes what configuration

But it is precisely the point of this PR to make this configuration be more readily _applicable_ to multiple plugins, regardless of which plugin _contributes_ it.

From the user perspective, they are configuring GitHub settings in Jenkins. If they need to troubleshoot why something is not working as expected, there are multiple techniques for doing that, such as enabling logging, searching known bug reports, etc. The link in help text may indeed be used to determine which plugin is defining a given configuration fragment, but this is not necessarily the source of a problem anyway.

KostyaSha: Github plugin provides connection for plugins, it has own documentation that guides on how configure this connection. I.e. in pullrequest-plugin i have reference: 1) configure github-plugin 2) configure PR plugin. 
Please remember that non enterprise usage it's not marketing feature where you want mask everything and resolve all issues via support. FOSS users should understand what and where should be configured. 

I think we can review and bit update github-plugin documentation to indicate that it could be used by other plugins.Technically it has API + ancient push->GitSCM.poll() parts. Other plugins should also reference to github-plugin (i.e. if configured user should have additional permissions).

kohsuke: I'd also point out that most other plugins do not put "Plugin" nor "Configuration" in it. In fact I should have further removed "Configuration" from the subject line.

kohsuke:
Before:
before

After:
after

src/main/resources/org/jenkinsci/plugins/github/config/GitHubPluginConfig/config.groovy
```diff
@@ -5,15 +5,12 @@ import com.cloudbees.jenkins.GitHubPushTrigger
 def f = namespace(lib.FormTagLib);
 
 f.section(title: descriptor.displayName) {
-    f.entry(title: _("Servers configs with credentials to manage GitHub integrations"),
-            description: _("List of GitHub Servers to manage hooks, set commit statuses etc."),
```
lanwen: think description is useful, as of some users confused what is this section for

src/main/resources/org/jenkinsci/plugins/github/config/GitHubServerConfig/config.groovy
```diff
@@ -6,21 +6,26 @@ def f = namespace(lib.FormTagLib);
...
 
 f.entry(title: _("Credentials"), field: "credentialsId") {
     c.select()
 }
 
-f.optionalBlock(title: _("Custom GitHub API URL"),
-        inline: true,
-        field: "customApiUrl",
-        checked: instance?.customApiUrl) {
-    f.entry(title: _("GitHub API URL"), field: "apiUrl") {
```
lanwen: Disagree. What for we should see the url all the time if it not enterprise?

KostyaSha:
@kohsuke thanks for screenshots
> After:

I think it would make sense place test button under `manage hooks` as ideally it should also verify ability to manage hooks (that is not possible to check :atm: because github-api doesn't expose headers AFAIR).

src/main/java/org/jenkinsci/plugins/github/config/GitHubPluginConfig.java
KostyaSha: Yeah, "configuration" sounds bit redundant. Would it be possible to place some UI element that would show that this section comes from gh plugin, then i would be ok to remove "plugin" word?

src/main/resources/org/jenkinsci/plugins/github/config/GitHubPluginConfig/config.groovy
```diff
@@ -5,15 +5,12 @@ import com.cloudbees.jenkins.GitHubPushTrigger
...
+    f.entry(title: _("GitHub Servers"),
             help: descriptor.getHelpFile()) {
         
         f.repeatableHeteroProperty(
                 field: "configs",
-                hasHeader: "true",
-                addCaption: _("Add GitHub Server Config"),
```
lanwen: Maybe just set to Add GitHub Server? As of its better for end-users than simple "Add" (it adds a userfriendly context)

KostyaSha: I also don't like default "add" "delete" as they are not very useful on what they really do because of pure jelly form selection (it visible, but with all not aligned checkboxes not very convenient).

KostyaSha: (Small reminder: verify documentation and screenshots if this changes would be merged.)

src/test/java/org/jenkinsci/plugins/github/internal/GitHubClientCacheCleanupTest.java
```diff
@@ -103,7 +102,10 @@ private void it(String comment, int count) throws IOException {
...
+        GitHub gitHub = new GitHubLoginFunction().apply(config);
```
lanwen: please use `GitHubServerConfig#loginToGithub()` instead

src/main/resources/org/jenkinsci/plugins/github/config/GitHubServerConfig/help-clientCacheSize.html
```diff
@@ -1,7 +1,11 @@
 <div>
-    Cache size in MB used by GitHub client. This can speed up fetching data form GH and reduce rate limits consuming.<br/>
-    GH + okHttp do all work for results reliability 
-    (<a href="https://developer.github.com/v3/#conditional-requests">Conditional-requests in GitHub documentation</a>)<br/>
-    
-    Set <b>0</b> to disable this feature
+    <p>
+        Jenkins will use this much space in <tt>$JENKINS_HOME</tt> to cache data retrieved from GitHub API calls.
```
lanwen: please add mention about MB

src/main/java/org/jenkinsci/plugins/github/config/GitHubPluginConfig.java
lanwen: this ok for me (it was here for historical reason)

KostyaSha: At all, would it render help button if we place somewhere help.html?

recena: :+1:

src/main/resources/org/jenkinsci/plugins/github/config/GitHubPluginConfig/config.groovy
kohsuke: In the interest of getting other changes in, if this is something you feel strongly I'm happy to give up, but  you should see it when you have one or two `GitHubServerConfig` instances. This text will really look out of place. Existing help file should explain what this field does. In addition, if this plugin is to be better reused from elsewhere, the wording is incorrect, too.

KostyaSha: Yes, github servers configuration part should be generified in UI and docs. But clicking help buttons is also not always convenient. Try open all collapsed items in configuration page to get review of configurations - you will tire clicking.

kohsuke:
> I think it would make sense place test button under manage hooks as ideally it should also verify ability to manage hooks (that is not possible to check :atm: because github-api doesn't expose headers AFAIR).

When the plugin actually develops the capability to check it, please make that change. This is out of scope of this PR.

KostyaSha:
> This is out of scope of this PR.

According to commits description you reordering items. Changing later would require redoing screenshots. Placing now probably would also exclude bad aligned help buttons.

src/main/resources/org/jenkinsci/plugins/github/config/GitHubServerConfig/config.groovy
kohsuke: I still think this is a step in the right direction. `github.com` endpoint is the default value there, so those who are only using it doesn't have to do anything, and for GHE users this is a little better.

src/main/resources/org/jenkinsci/plugins/github/config/GitHubPluginConfig/config.groovy
```diff
@@ -5,15 +5,13 @@ import com.cloudbees.jenkins.GitHubPushTrigger
...
-    f.entry(title: _("Servers configs with credentials to manage GitHub integrations"),
-            description: _("List of GitHub Servers to manage hooks, set commit statuses etc."),
+    f.entry(title: _("GitHub Servers"),
             help: descriptor.getHelpFile()) {
         
         f.repeatableHeteroProperty(
                 field: "configs",
                 hasHeader: "true",
-                addCaption: _("Add GitHub Server Config"),
-                deleteCaption: _("Delete GitHub Server Config"))
```
KostyaSha: Restore `deleteCaption`if you restored `addCaption` ?

KostyaSha:
FYI @kohsuke you may wait for both: my and @lanwen comments before doing changes.

src/main/resources/org/jenkinsci/plugins/github/config/GitHubServerConfig/config.groovy
KostyaSha: I also was initially for showing API url explicitly without any checkboxes or empty fields.

KostyaSha: CC @oleg-nenashev also participated in this debates before.

src/main/resources/org/jenkinsci/plugins/github/config/GitHubPluginConfig/config.groovy
lanwen: think add caption is enought

src/main/resources/org/jenkinsci/plugins/github/config/GitHubServerConfig/config.groovy
lanwen: my point is to hide as much as possible in default configuration from end user. If user wants to customize url - he clicks to the checkbox. Before this action he doesn't want to see the exact url. He just wants working state

KostyaSha: But global configuration is for admins - technical users that should keep eye on configuration. We should expect that they are not silly users :)

lanwen: ok, seems here is too much not silly users to continue debates on this checkbox

lanwen:
so, ok for me :+1: 
wait for a few more hours to additional reviews and will merge this

lanwen:
Will release it as `1.18.0` together with #111 (after it will be completed)

src/main/java/org/jenkinsci/plugins/github/config/GitHubServerConfig.java
```diff
@@ -98,28 +91,13 @@ public GitHubServerConfig(String credentialsId) {
...
-        }
-    }
-
-    /**
-     * Should be called before {@link #setApiUrl(String)}
-     *
-     * @param customApiUrl true if optional block "Custom GH Api Url" checked in UI
-     */
-    @DataBoundSetter
-    public void setCustomApiUrl(boolean customApiUrl) {
```
amuniz: This change broke backward compatibility, for example [here](https://github.com/jenkinsci/github-branch-source-plugin/blob/fadfb91c4ee9b62f18c6b427d2fb4814e244036c/src/main/java/org/jenkinsci/plugins/github_branch_source/Connector.java#L90).

recena: Exactly!

KostyaSha: See description of the field. This was used only for UI because of pure jelly form.

amuniz: > This was used only for UI because of pure jelly form.

It doesn't matter, it's public API.

lanwen: hmmm, thanks for the report, will return that in a few hours.

lanwen: #115

PR #114, opened 2016-03-15
prevent JS in link to github

lanwen:
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/114)

### DIFF
--- a/src/main/java/com/coravy/hudson/plugins/github/GithubLinkAction.java
+++ b/src/main/java/com/coravy/hudson/plugins/github/GithubLinkAction.java
@@ -1,12 +1,13 @@
 package com.coravy.hudson.plugins.github;
 
-import java.util.Collection;
-import java.util.Collections;
-
 import hudson.Extension;
 import hudson.model.Action;
 import hudson.model.Job;
 import jenkins.model.TransientActionFactory;
+import org.jenkinsci.plugins.github.util.XSSApi;
+
+import java.util.Collection;
+import java.util.Collections;
 
 /**
  * Add the Github Logo/Icon to the sidebar.
@@ -33,7 +34,7 @@ public final class GithubLinkAction implements Action {
 
     @Override
     public String getUrlName() {
-        return projectProperty.getProjectUrl().baseUrl();
+        return XSSApi.asValidHref(projectProperty.getProjectUrl().baseUrl());
     }
 
     @SuppressWarnings("rawtypes")
--- a/src/main/java/org/jenkinsci/plugins/github/util/XSSApi.java
+++ b/src/main/java/org/jenkinsci/plugins/github/util/XSSApi.java
@@ -1,0 +1,27 @@
+package org.jenkinsci.plugins.github.util;
+
+import java.net.MalformedURLException;
+import java.net.URL;
+
+/**
+ * @author lanwen (Merkushev Kirill)
+ */
+public final class XSSApi {
+    private XSSApi() {
+    }
+
+    /**
+     * Method to filter invalid url for XSS. This url can be inserted to href safely
+     *
+     * @param urlString unsafe url
+     *
+     * @return safe url
+     */
+    public static String asValidHref(String urlString) {
+        try {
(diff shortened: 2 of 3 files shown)


src/main/java/com/coravy/hudson/plugins/github/GithubLinkAction.java
```diff
@@ -33,7 +34,7 @@ public String getIconFileName() {
 
     @Override
     public String getUrlName() {
-        return projectProperty.getProjectUrl().baseUrl();
+        return XSSApi.asValidHref(projectProperty.getProjectUrl().baseUrl());
```
KostyaSha: `Functions.*escape*` `hudson.Util#escape`

src/main/java/com/coravy/hudson/plugins/github/GithubLinkAction.java
```diff
@@ -1,12 +1,13 @@
...
-import java.util.Collections;
-
 import hudson.Extension;
 import hudson.model.Action;
 import hudson.model.Job;
 import jenkins.model.TransientActionFactory;
+import org.jenkinsci.plugins.github.util.XSSApi;
+
+import java.util.Collection;
+import java.util.Collections;
```
KostyaSha: Add checkstyle verification for imports?

src/main/java/org/jenkinsci/plugins/github/util/XSSApi.java
```diff
@@ -0,0 +1,27 @@
+package org.jenkinsci.plugins.github.util;
+
+import java.net.MalformedURLException;
+import java.net.URL;
+
+/**
+ * @author lanwen (Merkushev Kirill)
+ */
+public final class XSSApi {
```
daniel-beck: Maybe `@Restricted` if you care about plugins using APIs of this plugin?

src/main/java/org/jenkinsci/plugins/github/util/XSSApi.java
```diff
@@ -0,0 +1,27 @@
...
+    /**
+     * Method to filter invalid url for XSS. This url can be inserted to href safely
+     *
+     * @param urlString unsafe url
+     *
+     * @return safe url
+     */
+    public static String asValidHref(String urlString) {
+        try {
+            return new URL(urlString).toExternalForm();
```
daniel-beck: What happens to relative links? Protocol-relative URLs?

src/main/java/com/coravy/hudson/plugins/github/GithubLinkAction.java
lanwen: Already is. What wrong now?

src/main/java/com/coravy/hudson/plugins/github/GithubLinkAction.java
lanwen: nope. They can't escape js

src/main/java/org/jenkinsci/plugins/github/util/XSSApi.java
lanwen: Expecting github url is **scheme + fqdn + project path** all other urls will be filtered

src/main/java/com/coravy/hudson/plugins/github/GithubLinkAction.java
KostyaSha: Import order.

lanwen: now it alphabetically ordered. As for me its ok

KostyaSha: Yes, but re-ordering mean that it not defined in checkstyle. http://checkstyle.sourceforge.net/config_imports.html#ImportOrder_Example

lanwen: this is not what we should care about

PR #120, opened 2016-04-28
Improved logging format

jglick:
I saw a custom logger showing things like

```
… INFO com.cloudbees.jenkins.GitHubPushTrigger$1 run
SCM changes detected in master. Triggering  #123
```

Probably this was some project inside of a folder whose full name would have been useful to know, but it was not visible here.

Also note the gratuitous space.

@reviewbybees

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/120)

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
+++ b/src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
@@ -108,7 +108,6 @@ public class GitHubPushTrigger extends Trigger<Job<?, ?>> implements GitHubTrigg
 
             public void run() {
                 if (runPolling()) {
-                    String name = " #" + job.getNextBuildNumber();
                     GitHubPushCause cause;
                     try {
                         cause = new GitHubPushCause(getLogFile(), pushBy);
@@ -117,9 +116,9 @@ public class GitHubPushTrigger extends Trigger<Job<?, ?>> implements GitHubTrigg
                         cause = new GitHubPushCause(pushBy);
                     }
                     if (asParameterizedJobMixIn(job).scheduleBuild(cause)) {
-                        LOGGER.info("SCM changes detected in " + job.getName() + ". Triggering " + name);
+                        LOGGER.info("SCM changes detected in " + job.getFullName() + ". Triggering #" + job.getNextBuildNumber());
                     } else {
-                        LOGGER.info("SCM changes detected in " + job.getName() + ". Job is already in the queue");
+                        LOGGER.info("SCM changes detected in " + job.getFullName() + ". Job is already in the queue");
                     }
                 }
             }
--- a/src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
+++ b/src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
@@ -116,7 +116,8 @@ public class GitHubPushTrigger extends Trigger<Job<?, ?>> implements GitHubTrigg
                         cause = new GitHubPushCause(pushBy);
                     }
                     if (asParameterizedJobMixIn(job).scheduleBuild(cause)) {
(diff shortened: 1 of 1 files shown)


KostyaSha:
LGTM

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -108,7 +108,6 @@ private boolean runPolling() {
 
             public void run() {
                 if (runPolling()) {
-                    String name = " #" + job.getNextBuildNumber();
```
jglick: I was concerned that for flyweight tasks merely calling `scheduleBuild` would cause the build to start _immediately_ and thus produce an inaccurate estimate if this variable were inlined, but the following test confirms that starting is asynchronous:

```java
@Test
public void nextBuildNumber() throws Exception {
    WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
    p.setDefinition(new CpsFlowDefinition(""));
    assertEquals(1, p.getNextBuildNumber());
    QueueTaskFuture<WorkflowRun> b = p.scheduleBuild2(0);
    assertEquals(1, p.getNextBuildNumber());
    b.get();
    assertEquals(2, p.getNextBuildNumber());
}
```

KostyaSha: next buildnumber would never be predictable with current polling logic.

recena:
🐝

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
jglick: Sure, it is always an estimate. IIRC there was some core PR to not even bother reporting `nextBuildNumber` in the build history widget. But `SCMTrigger` (from which this code seems to have been copied) does log it.

jglick:
https://github.com/jenkinsci/jenkins/commit/fac1cd26432899c816f7ef04a4b0126104ac835d seems to have been the corresponding core fix.

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
KostyaSha: I would suggest just remove it.

KostyaSha: `triggering build/run`

KostyaSha:
Let's remove this number prediction.

lanwen:
will release it a few hours later as 1.19.0

KostyaSha:
@lanwen 


> Let's remove this number prediction.

lanwen:
think predicted number is better than nothing

KostyaSha:
It wrong according to existed polling logic.

lanwen:
yep but it close to real in 1-2.

KostyaSha:
> I buy a 3k screen so that Checkstyle can complain I am using more than the leftmost third of it?

@jglick Do you want to buy us the same monitors?

jglick:
W.r.t. the build number prediction: it is probably pretty accurate in this case. The usual reason why it would be off is that you are scheduling multiple copies of the same task with different parameters (or some other `QueueAction`). `GitHubPushTrigger` itself does not seem to add any such action, so the only possibility would be some _other_ code happening to schedule a build with parameters at about the same time.

KostyaSha:
@jglick could you remind the full polling log (have no running instances under hand)?

jglick:
> could you remind the full polling log

From `SCMTrigger`? Have none handy.

KostyaSha:
Polling log that appears in sidepanel Action, afair it prints the body of hook event that says about one branch, then do polling and says that job XX triggered. User expecting to see this branch build in XXX build, but git scm decides checkout other branch. That's i think that such information falsehoods for user.

PR #124, opened 2016-05-04
use tagname without artifact prefix for releases

lanwen:
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/124)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -282,6 +282,19 @@
                     </execution>
                 </executions>
             </plugin>
+
+            <plugin>
+                <artifactId>maven-release-plugin</artifactId>
+                <configuration>
+                    <tagNameFormat>v@{project.version}</tagNameFormat>
+                    <mavenExecutorId>forked-path</mavenExecutorId>
+                    <useReleaseProfile>false</useReleaseProfile>
+                    <preparationGoals>clean install</preparationGoals>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -282,6 +282,19 @@
                     </execution>
                 </executions>
             </plugin>
+
+            <plugin>
+                <artifactId>maven-release-plugin</artifactId>
+                <configuration>
+                    <tagNameFormat>v@{project.version}</tagNameFormat>
```
KostyaSha: Why  with `v`? Why not just use version?

pom.xml
```diff
@@ -282,6 +282,19 @@
...
+                    <mavenExecutorId>forked-path</mavenExecutorId>
+                    <useReleaseProfile>false</useReleaseProfile>
+                    <preparationGoals>clean install</preparationGoals>
```
KostyaSha: neve do install when it not required to really install files into maven cache, should be `clean verify`

lanwen: copypaste from parent pom

pom.xml
lanwen: tagname looks better with `v` (imho)
https://github.com/VerbalExpressions/JavaVerbalExpressions/releases

PR #126, opened 2016-06-22
[FIXED JENKINS-36144] Borrow the SCMTrigger's queue

stephenc:
https://issues.jenkins-ci.org/browse/JENKINS-36144

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/126)

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
+++ b/src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
@@ -11,13 +11,17 @@ import hudson.model.Action;
 import hudson.model.Item;
 import hudson.model.Job;
 import hudson.model.Project;
+import hudson.triggers.SCMTrigger;
 import hudson.triggers.Trigger;
 import hudson.triggers.TriggerDescriptor;
 import hudson.util.FormValidation;
+import hudson.util.NamingThreadFactory;
 import hudson.util.SequentialExecutionQueue;
 import hudson.util.StreamTaskListener;
+import java.lang.reflect.Field;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadFactory;
 import jenkins.model.Jenkins;
-import jenkins.model.Jenkins.MasterComputer;
 import jenkins.model.ParameterizedJobMixIn;
 import jenkins.triggers.SCMTriggerItem.SCMTriggerItems;
 import org.apache.commons.jelly.XMLOutput;
@@ -45,7 +49,6 @@ import java.util.List;
 import java.util.Set;
 
 import static org.apache.commons.lang3.StringUtils.isEmpty;
-import static org.jenkinsci.plugins.github.Messages.github_trigger_check_method_warning_details;
 import static org.jenkinsci.plugins.github.util.JobInfoHelpers.asParameterizedJobMixIn;
 
 /**
@@ -72,7 +75,9 @@ public class GitHubPushTrigger extends Trigger<Job<?, ?>> implements GitHubTrigg
      */
     public void onPost(String triggeredByUser) {
         final String pushBy = triggeredByUser;
-        getDescriptor().queue.execute(new Runnable() {
+        DescriptorImpl d = getDescriptor();
+        d.checkThreadPoolSize();
+        d.queue.execute(new Runnable() {
             private boolean runPolling() {
                 try {
                     StreamTaskListener listener = new StreamTaskListener(getLogFile());
@@ -226,7 +231,7 @@ public class GitHubPushTrigger extends Trigger<Job<?, ?>> implements GitHubTrigg
     @Extension
     public static class DescriptorImpl extends TriggerDescriptor {
         private final transient SequentialExecutionQueue queue =
-                new SequentialExecutionQueue(MasterComputer.threadPoolForRemoting);
+                new SequentialExecutionQueue(Executors.newSingleThreadExecutor(threadFactory()));
 
         private transient String hookUrl;
 
@@ -234,6 +239,43 @@ public class GitHubPushTrigger extends Trigger<Job<?, ?>> implements GitHubTrigg
 
         @Inject
         private transient GitHubHookRegisterProblemMonitor monitor;
+
+        @Inject
+        private transient SCMTrigger.DescriptorImpl scmTrigger;
+
+        private transient int maximumThreads = Integer.MIN_VALUE;
+
+        private static ThreadFactory threadFactory() {
+            return new NamingThreadFactory(Executors.defaultThreadFactory(), "GitHubPushTrigger");
+        }
+
+        public DescriptorImpl() {
+            checkThreadPoolSize();
+        }
+
+        /**
+         * Update the {@link java.util.concurrent.ExecutorService} instance.
+         */
+        /*package*/
+        synchronized void checkThreadPoolSize() {
+            if (scmTrigger != null) {
+                int count = scmTrigger.getPollingThreadCount();
+                if (maximumThreads != count) {
+                    maximumThreads = count;
+                    try {
+                        Field getQueue = SCMTrigger.DescriptorImpl.class.getDeclaredField("queue");
+                        getQueue.setAccessible(true);
+                        SequentialExecutionQueue q = (SequentialExecutionQueue) getQueue.get(scmTrigger);
+                        this.queue.setExecutors(q.getExecutors());
+                    } catch (NoSuchFieldException | IllegalAccessException | ClassCastException e) {
+                        queue.setExecutors(
+                                (count == 0
+                                        ? Executors.newCachedThreadPool(threadFactory())
+                                        : Executors.newFixedThreadPool(maximumThreads, threadFactory())));
+                    }
+                }
+            }
+        }
 
         @Override
         public boolean isApplicable(Item item) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -235,6 +241,47 @@ public void writeLogTo(XMLOutput out) throws IOException {
...
+         */
+        /*package*/
+        synchronized void checkThreadPoolSize() {
+            if (scmTrigger != null) {
+                int count = scmTrigger.getPollingThreadCount();
+                if (maximumThreads != count) {
+                    maximumThreads = count;
+                    try {
+                        Field getQueue = SCMTrigger.DescriptorImpl.class.getDeclaredField("queue");
+                        getQueue.setAccessible(true);
```
KostyaSha: hack?

stephenc: of course... but after a more careful analysis I decided against that and rather just mirror the pool size... which sadly means you have two pools... but this does point that perhaps we should have an explicit pool for polling that plugins can use rather than the current status where each push notifier has to roll their own... with hillarious consequences

jtnord:
:bee: @stephenc I assume you are going to file a issue for the generic issue

> but after a more careful analysis I decided against that and rather just mirror the pool size... which sadly means you have two pools... but this does point that perhaps we should have an explicit pool for polling that plugins can use rather than the current status where each push notifier has to roll their own... with hillarious consequences

rsandell:
:bee: A future further improvement could perhaps also be to not schedule the polling if the same repo and branch is already in the queue.

KostyaSha:
Queue skipping polling for the same project AFAIK

KostyaSha:
> A future further improvement could perhaps also be to not schedule the polling if the same repo and branch is already in the queue.

Could you describe how queue works now?

stephenc:
https://issues.jenkins-ci.org/browse/JENKINS-36147 filed @jtnord

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -333,6 +362,10 @@ public static boolean allowsHookUrlOverride() {
             return ALLOW_HOOKURL_OVERRIDE;
         }
 
+        private static ThreadFactory threadFactory() {
```
lanwen: it should be method that returns executors - `threadPoolWithCapacity(int)`

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -235,6 +239,31 @@ public void writeLogTo(XMLOutput out) throws IOException {
...
+        }
+
+        /**
+         * Update the {@link java.util.concurrent.ExecutorService} instance.
+         */
+        /*package*/
+        synchronized void checkThreadPoolSize() {
+            if (scmTrigger != null) {
+                int count = scmTrigger.getPollingThreadCount();
+                if (maximumThreads != count) {
```
lanwen: why just not `count >= 0`?

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -235,6 +239,31 @@ public void writeLogTo(XMLOutput out) throws IOException {
...
+
+        public DescriptorImpl() {
+            checkThreadPoolSize();
+        }
+
+        /**
+         * Update the {@link java.util.concurrent.ExecutorService} instance.
+         */
+        /*package*/
+        synchronized void checkThreadPoolSize() {
```
lanwen: should be named `updateThreadPoolSize` as of it not only checks it

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -235,6 +239,31 @@ public void writeLogTo(XMLOutput out) throws IOException {
...
+            if (scmTrigger != null) {
+                int count = scmTrigger.getPollingThreadCount();
+                if (maximumThreads != count) {
+                    queue.setExecutors(
+                            (count == 0
+                                    ? Executors.newCachedThreadPool(threadFactory())
+                                    : Executors.newFixedThreadPool(maximumThreads, threadFactory())));
```
jtnord: Does this not cause a thread leak?
Say `ScmTrigger.pollingThreadCount` was set to `100`, this kicks in and creates a new fixed size thread pool, then you have a lot of notifcations and you use all 100 Threads...
then the admin sees this and sets `ScmTrigger.pollingThreadCount` to a more sane value of `10`, and this creates a new fixed size thread pool Executor...

But the original ExecutorService (with 100 threads) is not shutdown, and not does it wait for all the tasks to complete...

Basically you are waiting for GC to kick (and hoping that it does the right thing)...

jtnord:
:bug: as I think the Executor service has the potential to leak threads (at least temporarily until some GC kicks in, the point of which is non-deterministic)

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
stephenc: When you call `queue.setExecutor(queue.getExecutor())` it will basically shutdown its own executor (which is why I stopped trying to steal SCMTrigger's queue because `setExecutor` shuts down the old one (without a no-op check beforehand... Nice eh))

So yes the admin will have 110 threads until the 100 previously executing tasks complete. No need to wait for GC

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
stephenc: Because we want to catch when the thread pool size is changed in any way (up or down)

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
stephenc: You mean `checkThreadPoolSizeAndUpdateIfNecessary` if you want to be exact. 

Most calls will just be a check so I named after the normal case

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
stephenc: I don't see the value, seems like change for change's sake. The other method is not too complex whereas we'd need to add a three way test to use this as an executor service factory method

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
lanwen: i'm for the verbose naming in such cases

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
lanwen: so we need the ability to change maximumThreads. Can't see it now

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
lanwen: Don't think the complexity reduction in such simple way is "like change for change's sake". At minimal state it should be named as `namedThreadFactory` as of its simple shortcut.

jtnord:
Yes, my bad.  :bee:

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
stephenc: We piggy-back off `SCMTrigger.getPollingThreadCount()` which gives the UI. The `maximumThreads` field is a cache of the last seen value

stephenc: aha! a bug caught! cool!

KostyaSha:
Btw, the same queue i'm using in other trigger plugins. Is it possible to make queue public and place configuration in global config?

KostyaSha:
@reviewbybees @stephenc relogin!

stephenc:
Nahh just depends which email I reply to!

stephenc:
Anyway, my phone is the holder of the 2FA for that account...

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -235,6 +239,32 @@ public void writeLogTo(XMLOutput out) throws IOException {
...
+
+        public DescriptorImpl() {
+            checkThreadPoolSizeAndUpdateIfNecessary();
+        }
+
+        /**
+         * Update the {@link java.util.concurrent.ExecutorService} instance.
+         */
+        /*package*/
+        synchronized void checkThreadPoolSizeAndUpdateIfNecessary() {
```
lanwen: it should be a test for this logic

stephenc: I have lost any energy to continue this PR any further. I want to concentrate my time on the critical root cause, i.e. https://issues.jenkins-ci.org/browse/JENKINS-36147 which once implemented will mean we can rip out this code. If you want to add tests to this code which should be removed once I implement https://issues.jenkins-ci.org/browse/JENKINS-36147, feel free. I will not be updating this PR any further at this point.

KostyaSha: It would require updating core that is not good for this plugin. Or we can backport your core improvements after.

stephenc: well my plan is to make the API available via scm-api without a core bump... and then on newer cores it either will revert to the core implementation (classloader priority) or the scm-api variant will use reflection to redirect to the core implementation

stephenc: However, it will take me a while to get that API right (as I don't want to end up with the same mistakes of the main Queue API and end up having to have a "big fat lock") so for now, to save users from commit storms killing their Jenkins, this fix would be good to have released.

lanwen: ookay, i'll merge this. But this is an example of permanent state "will fix that better in future releases". It can't be finished while you've start do things right everywhere

PR #127, opened 2016-07-05
[JENKINS-36445] Reduce length of displayed SHA1 in changes

MarkEWaite:
Most git repository browsing systems (bitbucket, cgit, github, gitweb)
present a short form of the SHA1 (commonly the first 7 characters) with
the clickable hyperlink that will take them to a page that includes the
full SHA1 of the commit.

This change reduces the visible text to the first 7 characters of the
SHA1, consistent with those other repository browsers.  It also adds
assertions that the expected format is used.

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/127)

### DIFF
--- a/src/test/java/com/coravy/hudson/plugins/github/GithubLinkAnnotatorTest.java
+++ b/src/test/java/com/coravy/hudson/plugins/github/GithubLinkAnnotatorTest.java
@@ -2,12 +2,31 @@ package com.coravy.hudson.plugins.github;
 
 import static org.junit.Assert.assertEquals;
 import hudson.MarkupText;
+import hudson.plugins.git.GitChangeSet;
+import java.util.ArrayList;
+import org.junit.Before;
 
 import org.junit.Test;
 
 public class GithubLinkAnnotatorTest {
 
     private final static String GITHUB_URL = "http://github.com/juretta/iphone-project-tools/";
+    private final static String SHA1 = "deadbeef36cd854f4dd6fa40bf94c0c657681dd5";
+    private GitChangeSet changeSet;
+
+    @Before
+    public void createChangeSet() throws Exception {
+        ArrayList<String> lines = new ArrayList<String>();
+        lines.add("commit " + SHA1);
+        lines.add("tree 66236cf9a1ac0c589172b450ed01f019a5697c49");
+        lines.add("parent e74a24e995305bd67a180f0ebc57927e2b8783ce");
+        lines.add("author Author Name <author.name@nospam.com> 1363879004 +0100");
+        lines.add("committer Committer Name <committer.name@nospam.com> 1364199539 -0400");
+        lines.add("");
+        lines.add("    Committer and author are different in this commit.");
+        lines.add("");
+        changeSet = new GitChangeSet(lines, true);
+    }
 
     @Test
     public final void testAnnotateStringMarkupText() {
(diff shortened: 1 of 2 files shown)


src/test/java/com/coravy/hudson/plugins/github/GithubLinkAnnotatorTest.java
```diff
@@ -25,13 +44,44 @@ public final void testAnnotateStringMarkupText() {
                         + "issues/9876/find'>close #9876</a> link");
     }
 
+    @Test
+    public final void testChangeSetAnnotateStringMarkupText() {
```
lanwen: did you know about "Matchers" pattern?

code can be organized in form:
```java
assertThat(originalText, annotatedWith(changeSet, equalTo(expectedChangeSetAnnotation)));
```

Please try to use 
```java
public static Matcher<String> annotatedWith(GitChangeSet chSet, Matcher<String> matcher) {
   return new FeatureMatcher<String, String>(matcher, "annotated text", "") {
    
      @Override
       public String featureValueOf(String actual) {
            return annotate(actual, chSet);
       }
}
```

More info: http://hamcrest.org/JavaHamcrest/

Please also organize multiply cases same as in:
https://github.com/jenkinsci/github-plugin/blob/master/src/test/java/org/jenkinsci/plugins/github/util/XSSApiTest.java
More info: https://github.com/TNG/junit-dataprovider

lanwen:
Thanks for your PR and tests! Please try to improve test code. Feel free to ask any questions if you need more detailed explanation

src/test/java/com/coravy/hudson/plugins/github/GithubLinkAnnotatorTest.java
MarkEWaite: Interesting. I diligently tried to retain the pattern and style that was already in the test, rather than switching to a Hamcrest based pattern.

I'm happy to switch to a Hamcrest based pattern and to use similar structure to XSSApiTest.

Are you willing to consider the behavioral change at the root of this request?  When the git plugin was rewritten (from git plugin 1.x to git plugin 2.x), one of the behaviors that changed was to shorten the commit messages.  I didn't implement that change in the git plugin, but I believe it is the right change.  

However, there are at least a [few git plugin 1.x users](https://issues.jenkins-ci.org/browse/JENKINS-29977) who object to the git plugin 2.x message truncation behavior.

I'm not sure if GitHub plugin users are as vocal about truncation of the SHA1, or rather if they are surprised (like I was) to see the full text of the SHA1 displayed in the web user interface.

lanwen: I'm completely for the new good practices in tests! :)

Does not think that truncating a commit message on the server side is good idea. 
The commit message length is just a recommended approach, so can be ignored. With current UI jenkins can't provide good UX to see full change log with truncated messages. But commit sha mostly is just a way to fast jump to exact commit in github interface and its ok to limit it with 7 symbols

MarkEWaite: Thanks for the clarity.  As time allows, I'll rework this pull request using Hamcrest assertions and will resubmit.  I'll close this pull request and replace it with a new pull request when the Hamcrest based assertions are ready.

lanwen:
You can just rewrite your branch to autoupdate PR, no need to send a new one

KostyaSha:
You can create new branch `checkout -b continueWork` then push it to this remote PR branch and PR will automatically get new code, optionally locally you will have your previous work.

MarkEWaite:
Thanks for the recommendations.  It was fun reworking the test to use Hamcrest assertions.  Are you willing to review PR128 (which is the reworked pull request), or do I need to close that pull request and reopen this one with the changes?

lanwen:
No need to reopen. Will continue in new one

PR #129, opened 2016-07-08
[JENKINS-33974] Payload verification using shared secrets

martinmine:
This pull requests adds the support for verification of payloads from GitHub using SHA1 HMAC digest of payloads.
This pull request solves issue [JENKINS-33974](https://issues.jenkins-ci.org/browse/JENKINS-33974?jql=project%20%3D%20JENKINS%20AND%20status%20in%20(Open%2C%20%22In%20Progress%22%2C%20Reopened)%20AND%20component%20%3D%20%27github-plugin%27).

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/129)


src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -14,23 +14,24 @@
 import hudson.triggers.SCMTrigger;
 import hudson.triggers.Trigger;
 import hudson.triggers.TriggerDescriptor;
-import hudson.util.FormValidation;
-import hudson.util.NamingThreadFactory;
-import hudson.util.SequentialExecutionQueue;
-import hudson.util.StreamTaskListener;
+import hudson.util.*;
```
KostyaSha: don't collapse imports

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -57,8 +58,18 @@
  */
 public class GitHubPushTrigger extends Trigger<Job<?, ?>> implements GitHubTrigger {
 
+    private Secret sharedSecret;
+
     @DataBoundConstructor
-    public GitHubPushTrigger() {
```
KostyaSha: breaks backward compatibility

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -57,8 +58,18 @@
  */
 public class GitHubPushTrigger extends Trigger<Job<?, ?>> implements GitHubTrigger {
 
+    private Secret sharedSecret;
```
KostyaSha: Must be stored in Credentials

KostyaSha: also it should be in connector API and not in trigger.

src/main/java/org/jenkinsci/plugins/github/extension/GHEventsSubscriber.java
```diff
@@ -55,12 +55,13 @@
...
      * parse logic
-     * Don't call it directly, use {@link #processEvent(GHEvent, String)} static function
+     * Don't call it directly, use {@link #processEvent(GHEvent, String, String)} static function
      *
      * @param event   gh-event (as of PUSH, ISSUE...). One of returned by {@link #events()} method. Never null.
      * @param payload payload of gh-event. Never blank. Can be parsed with help of GitHub#parseEventPayload
+     * @param signature X-Hub-Signature header value, HMAC hex digest of payload from GitHub.
      */
-    protected void onEvent(GHEvent event, String payload) {
+    protected void onEvent(GHEvent event, String payload, String signature) {
```
KostyaSha: breaks backward compatibility for API!

src/main/java/org/jenkinsci/plugins/github/webhook/GHSignatureHeader.java
```diff
@@ -0,0 +1,50 @@
...
+
+import static java.lang.annotation.ElementType.PARAMETER;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+import static org.slf4j.LoggerFactory.getLogger;
+
+/**
+ * InjectedParameter annotation to use on WebMethod parameters.
+ * Grabs the X-Hub-Signature header value from HTTP requests and injects them to function parameters.
+ *
+ * @author lanwen (Merkushev Kirill)
```
KostyaSha: Are you @lanwen ?

KostyaSha:
👎 
Secret exists in hook configuration, hooks a management by global config and this API provided for this and other plugins. Every changes constructor/method in API breaks backward compatibility and other plugins. Secrets shouldn't be stored directly, use Credentials api and store id.

src/main/java/org/jenkinsci/plugins/github/webhook/GHSignatureHeader.java
martinmine: My bad, the code was mostly a copy-paste of a class he had written. I should be more explicit and say it was based off code he had written.

KostyaSha: or add other author. or document that it copy of other class.

martinmine:
I'll take care of the breaking API changes and the credentials API over the weekend. When attempting to run the checkstyle (mvn checkstyle:checkstyle), I'm receiving the following error on my box:
`[ERROR] [ERROR] Unresolveable build extension: Plugin org.jenkins-ci.tools:maven-hpi-plugin:1.117 or one of its dependencies could not be resolved: Failure to find org.jenkins-ci.tools:maven-hpi-plugin:jar:1.117 in http://repo.maven.apache.org/maven2 was cached in the local repository, resolution will not be reattempted until the update interval of central has elapsed or updates are forced @`
Any specific reason for this or am I running the wrong command for executing checkstyle? Also, when you say 

> Secret exists in hook configuration, hooks a management by global config and this API provided for this and other plugins.

do you mean if there is already support for storing secrets (before this pull request)?

KostyaSha:
> do you mean that there is already support for storing secrets (before this pull request)?

No, i mean that it needed only during hook processing and shouldn't be available for end user.

src/main/java/org/jenkinsci/plugins/github/extension/GHEventsSubscriber.java
```diff
@@ -55,12 +55,13 @@
      * This method called when root action receives webhook from GH and this extension is interested in such
      * events (provided by {@link #events()} method). By default do nothing and can be overrided to implement any
      * parse logic
-     * Don't call it directly, use {@link #processEvent(GHEvent, String)} static function
+     * Don't call it directly, use {@link #processEvent(GHEvent, String, String)} static function
```
lanwen: this class should not know about any sign.

src/main/java/org/jenkinsci/plugins/github/webhook/subscriber/DefaultPushGHEventSubscriber.java
```diff
@@ -100,4 +95,37 @@ public void run() {
             LOGGER.warn("Malformed repo url {}", repoUrl);
         }
     }
+
+    public void triggerJobs(final GitHubRepositoryName changedRepository, final String payload,
```
lanwen: trigger should not know about signature too

src/main/java/org/jenkinsci/plugins/github/extension/GHEventsSubscriber.java
martinmine: @lanwen Considering that there are project-specific secrets, another solution would be to parse the payload twice as the validation would be dependent on the repo which is specified inside the payload, thus making it required to parse the message before validating. Do you think this sounds like an acceptable solution or do you have any other idea how to approach this?

KostyaSha: Why do you need to make it project-specific?

martinmine: More flexibility in terms of security. Only having the ability to use a global secret means more people would potentially have to share the key among themselves if setting up the hook manually. Having the option to use a project-specific key (that would also override any global key) mitigates this issues by letting them specify their own key.

lanwen: It should be not project-specific, but user-specific, as of user with admin access can view and change signature keys for all projects.
So if you want flexible solution - just register one admin user per project.

martinmine: I'm mostly thinking about risks outside the CI environment, for example if an attacker obtains a secret, it would be possible to track down which project the secret belongs to. Having project-specific secrets would also prevent them from starting to execute jobs belonging to other project as the key is tied to a specific project. From my point of view, it is not the issue that an admin has access to the keys for all the projects, but that the keys are loosely tied to the projects when they are system-wide and if there is a large amount of hooks that have different repos with different repo owners.

martinmine:
The implementation is now using the credentials API, as suggested by @KostyaSha and there is also no project-specific keys.

lanwen:
Sorry for the delay, need to look closer

src/main/java/org/jenkinsci/plugins/github/webhook/RequirePostWithGHHookPayload.java
```diff
@@ -114,6 +130,55 @@ protected void shouldContainParseablePayload(Object[] arguments) throws Invocati
...
 
         /**
+         * Checks that an incoming request has a valid signature, if there is specified a signature in the config.
+         *
+         * @param req Incoming request.
+         * @throws InvocationTargetException if any of preconditions is not satisfied
+         */
+        protected void shouldProvideValidSignature(StaplerRequest req, Object[] args) throws InvocationTargetException {
+            final String signature = parseSHA1Value(req.getHeader(SIGNATURE_HEADER));
+            final Secret secret = Jenkins.getInstance()
```
lanwen: `GitHubPlugin.configuration().getHookSecretConfig().getHookSecret();`

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -14,12 +14,14 @@
 import hudson.triggers.SCMTrigger;
 import hudson.triggers.Trigger;
 import hudson.triggers.TriggerDescriptor;
+
```
lanwen: seems useless change

src/main/java/org/jenkinsci/plugins/github/config/HookSecretConfig.java
```diff
@@ -0,0 +1,95 @@
...
+
+import static com.cloudbees.plugins.credentials.CredentialsMatchers.firstOrDefault;
+import static com.cloudbees.plugins.credentials.CredentialsMatchers.withId;
+import static com.cloudbees.plugins.credentials.CredentialsProvider.lookupCredentials;
+import static org.apache.commons.lang.StringUtils.isEmpty;
+
+/**
+ * Manages storing/retrieval of the shared secret for the hook.
+ */
+@XStreamAlias("hook-config")
```
lanwen: actually hook-secret-config

src/main/resources/org/jenkinsci/plugins/github/config/HookSecretConfig/help-sharedSecret.html
```diff
@@ -0,0 +1,4 @@
+<div>
+    A shared secret token GitHub will use to sign requests in order for Jenkins to verify that the request came from GitHub.
+    If left blank, this feature will not be used.
```
lanwen: Please add a link to corresponding section of GitHub help

src/main/java/org/jenkinsci/plugins/github/config/GitHubPluginConfig.java
```diff
@@ -61,6 +61,7 @@
 
     private List<GitHubServerConfig> configs = new ArrayList<GitHubServerConfig>();
     private URL hookUrl;
+    private HookSecretConfig hookSecretConfig = new HookSecretConfig(null);
```
lanwen: should be part of the github-server-configuration, but not the entire jenkins

lanwen:
👍  will add some changes by myself

PR #133, opened 2016-07-27
Support aborted build status reporting

KevinPike:


KostyaSha:
> does jenkins call publishers on aborted build? Don't think so.

No, it doesn't. AFAIK

KevinPike:
The behavior that I'm seeing that is when a build starts, a pending github commit status is created. When the build is aborted, another pending github commit status is created. I believe that second status is pending because it is the default status. It does appear that the github plugin is getting called for an aborted build.

lanwen:
please provide a test to reproduce (you can find examples with mocked github in tests for notifier)

PR #134, opened 2016-08-11
[JENKINS-33974] Payload verification using shared secrets

lanwen:
Closes #129

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/134)


MartinNowak:
This seems to fail for some form encoded payloads, like the one attached below @lanwen.
[payload.encoded.txt](https://github.com/jenkinsci/github-plugin/files/818832/payload.encoded.txt)
[payload.json.txt](https://github.com/jenkinsci/github-plugin/files/818833/payload.json.txt)

Running `URLEncode(URLDecode(payload))` does not return the same result, apparently b/c `~` is encoded by Java but not by GH.

lanwen:
@MartinNowak can you write a test to reproduce? (And may be a full fix?:)

MartinNowak:
No, too much time already spent on figuring this out, and I'm neither familiar with nor keen to dig further into this mass of Java code/plugins.
The obvious fix is to checksum the original request body instead of trying to recreate it from the decoded form.

There isn't much wrong with recommending application/json as workaround, though GH uses application/x-www-form-urlencoded by default which applies to the hooks installed by github-organization (or github-branch-source or github-api or whatever).

lanwen:
Ok, thanks in any case, will try to do something

nh2:
I'm also still seeing `Provided signature [aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa] did not match to calculated` for specific branch names, when using `application/x-www-form-urlencoded`.

I suspect in my case it's because the branch name contains a `.`

PR #135, opened 2016-08-12
Print warning when repositories differ by case

rcoh:
There are many instances where people have run into hard to debug problems because Github webhooks in the plugin are case sensitive where the rest of Jenkins is case insensitive. Ideally, I'd like to change the equality behavior to be case insensitive, but I recognize that could be backwards incompatible for some people.

As a compromise, I'd like to add a WARN when comparing 2 repositories that would be equal but aren't because the case is different. At least in this case, when users look at the logs, it will be obvious what is happening.

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/135)


lanwen:
Thanks for the PR, but warning message in equals method is not right approach.

PR #137, opened 2016-09-14
Verbosive logging for commit status setter

lanwen:
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/137)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/github/status/GitHubCommitStatusSetter.java
+++ b/src/main/java/org/jenkinsci/plugins/github/status/GitHubCommitStatusSetter.java
@@ -143,6 +143,15 @@ public class GitHubCommitStatusSetter extends Notifier implements SimpleBuildSte
             String message = result.getMsg();
             GHCommitState state = result.getState();
 
+            listener.getLogger().printf(
+                    "[%s] %s on repos %s (sha:%7.7s) with context:%s%n",
+                    getDescriptor().getDisplayName(),
+                    state,
+                    repos,
+                    sha,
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/github/status/GitHubCommitStatusSetter.java
```diff
@@ -128,6 +128,11 @@ public void perform(@Nonnull Run<?, ?> run, @Nonnull FilePath workspace, @Nonnul
             String message = result.getMsg();
             GHCommitState state = result.getState();
 
+            listener.getLogger().printf(
+                    "[GitHub Commit Status Setter] Going to set status %s on repos %s (sha:%7.7s) with context:%s%n",
```
KostyaSha: normally prefix should be short and name plugin and then the same UI name i.e. `[GitHub] [getDescriptor().getDisplayName()]`

lanwen: Yep, but display name is too long for now

KostyaSha: Could be renamed to `Set GitHub commit status [universal]` btw and then rename log output without `going blabla` as it already what `Descriptor.getDisplayName()` claim.

lanwen: You are right, will do

PR #139, opened 2016-09-19
[FIXED JENKINS-38347] Use Initializer levels.

KostyaSha:
- migrator: Don't throw NPE because Descriptor wasn't ready.
- aliases: user annotation initializer.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/github/GitHubPlugin.java
+++ b/src/main/java/org/jenkinsci/plugins/github/GitHubPlugin.java
@@ -1,16 +1,19 @@
 package org.jenkinsci.plugins.github;
 
 import hudson.Plugin;
+import hudson.init.Initializer;
 import org.jenkinsci.plugins.github.config.GitHubPluginConfig;
 import org.jenkinsci.plugins.github.migration.Migrator;
 
 import javax.annotation.Nonnull;
 
+import static hudson.init.InitMilestone.PLUGINS_PREPARED;
+import static hudson.init.InitMilestone.PLUGINS_STARTED;
 import static org.apache.commons.lang3.ObjectUtils.defaultIfNull;
 
 /**
  * Main entry point for this plugin
- *
+ * <p>
  * Launches migration from old config versions
  * Contains helper method to get global plugin configuration - {@link #configuration()}
  *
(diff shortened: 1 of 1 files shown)


KostyaSha:
@lanwen @atonasenko for review

src/main/java/org/jenkinsci/plugins/github/GitHubPlugin.java
```diff
@@ -21,21 +24,17 @@
      * Launched before plugin starts
      * Adds alias for {@link GitHubPlugin} to simplify resulting xml
      */
-    public static void init() {
+    @Initializer(before = PLUGINS_STARTED)
```
KostyaSha: level needs to be checked

KostyaSha: Added workaround to ensure order

src/main/java/org/jenkinsci/plugins/github/GitHubPlugin.java
```diff
@@ -21,25 +24,31 @@
      * Launched before plugin starts
      * Adds alias for {@link GitHubPlugin} to simplify resulting xml
      */
-    public static void init() {
+//    @Initializer(before = PLUGINS_STARTED)
```
lanwen: ?

lanwen: can be removed i think

KostyaSha: I kept it just to not start thinking in future what was the right level.

atanasenko:
+1

src/main/java/org/jenkinsci/plugins/github/GitHubPlugin.java
lanwen: move it to javadoc plz - now it looks like a forgotten block of code

KostyaSha: done

lanwen:
thanks!

atanasenko:
Thanks guys!

PR #149, opened 2016-10-17
WIP Deprecate magic method.

KostyaSha:
Relying on this manageHooks may end to situation that it will try register hook for GH server that has disabled hooks. 

PS One line fluent is hardly readable for eyes.


src/main/java/org/jenkinsci/plugins/github/config/GitHubPluginConfig.java
```diff
@@ -90,8 +90,15 @@ public void setConfigs(List<GitHubServerConfig> configs) {
         return configs;
     }
 
+    /**
+     * @deprecated get GitHubServerConfig matching to exact remote GH server, filter and only then check allowToManage
```
lanwen: this method exists only to check - can we ever manage hooks somewhere

lanwen:
should be notice, not deprecation

src/main/java/org/jenkinsci/plugins/github/config/GitHubPluginConfig.java
KostyaSha: But it useless.

lanwen: nope, we can skip some actions if no any server is allowed to manage hooks (such as cleanup)

PR #150, opened 2016-10-19
[FIXED JENKINS-36446] Ensure all plugin extensions are initialized before migration

stephenc:
See [JENKINS-36446](https://issues.jenkins-ci.org/browse/JENKINS-36446)

@reviewbybees

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/150)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/github/GitHubPlugin.java
+++ b/src/main/java/org/jenkinsci/plugins/github/GitHubPlugin.java
@@ -1,14 +1,13 @@
 package org.jenkinsci.plugins.github;
 
 import hudson.Plugin;
+import hudson.init.InitMilestone;
 import hudson.init.Initializer;
 import org.jenkinsci.plugins.github.config.GitHubPluginConfig;
 import org.jenkinsci.plugins.github.migration.Migrator;
 
 import javax.annotation.Nonnull;
 
-import static hudson.init.InitMilestone.PLUGINS_PREPARED;
-import static hudson.init.InitMilestone.PLUGINS_STARTED;
 import static org.apache.commons.lang3.ObjectUtils.defaultIfNull;
 
 /**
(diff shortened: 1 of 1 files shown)


stephenc:
Unsure what is up with Travis and its test failure... Jenkins had no issue building

KostyaSha:
Travis is known issue, i see no exact version from reporter and initializers was adjusted in last release. Did you reproduce and fix issues locally or blindly adopted initializers as you feeling?

stephenc:
@KostyaSha it is incorrect to run code that invokes other plugins prior to `EXTENSIONS_AUGMENTED` as that is the first stage in the lifecycle where the `ExtensionList` is committed. So doesn't matter that you refactored the code, the migration wants to migrate credentials, therefore it needs to invoke the credentials plugin, therefore the migration *must* be at least after `EXTENSIONS_AUGMENTED` or the Jenkins master will fail to start

KostyaSha:
@stephenc ok, could you create some test with TestPluginExtension that will show behaviour?

rsandell:
:bee:

stephenc:
@KostyaSha I do not have time to work on this issue further. It is not infinite fixing BTW, it is the complete inability to have fixing work. To reproduce I suspect all you need to do is start with an old version of the plugin with some legacy credentials defined, then upgrade this plugin and credentials plugin at the same time... restart... boom!

KostyaSha:
Would be greate to block this PR until you fix https://github.com/jenkinsci/workflow-job-plugin/pull/28 , but because of users of course it will be solved 10x time faster then you handle community.

KostyaSha:
cc @lanwen

stephenc:
@kostyasha I cannot be everywhere all at once... ;-)

stephenc:
Thanks @lanwen

PR #152, opened 2016-11-05
Crumb exclusion should be more forgiving if the user leaves off the trailing slash

i386:
Discovered that if I left out the trailing slash on my webhook configuration in Github (e.g. `/github-webook` rather than `/github-webhook/`) then incoming requests would not be covered by the crumb exclusion.

This change also adds tests for `GitHubWebHookCrumbExclusion`.

PTAL @jenkinsci/code-reviewers @KostyaSha

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/152)

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/GitHubWebHookCrumbExclusion.java
+++ b/src/main/java/com/cloudbees/jenkins/GitHubWebHookCrumbExclusion.java
@@ -9,6 +9,8 @@ import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import java.io.IOException;
 
+import static org.apache.commons.lang3.StringUtils.isEmpty;
+
 @Extension
 public class GitHubWebHookCrumbExclusion extends CrumbExclusion {
 
@@ -16,11 +18,16 @@ public class GitHubWebHookCrumbExclusion extends CrumbExclusion {
     public boolean process(HttpServletRequest req, HttpServletResponse resp, FilterChain chain)
             throws IOException, ServletException {
         String pathInfo = req.getPathInfo();
-        if (pathInfo != null && pathInfo.equals(getExclusionPath())) {
-            chain.doFilter(req, resp);
-            return true;
+        if (isEmpty(pathInfo)) {
+            return false;
         }
-        return false;
+        // Github will not follow redirects https://github.com/isaacs/github/issues/574
+        pathInfo = pathInfo.endsWith("/") ? pathInfo : pathInfo + '/';
+        if (!pathInfo.equals(getExclusionPath())) {
+            return false;
+        }
+        chain.doFilter(req, resp);
+        return true;
     }
 
     public String getExclusionPath() {
--- a/src/test/java/com/cloudbees/jenkins/GitHubWebHookCrumbExclusionTest.java
+++ b/src/test/java/com/cloudbees/jenkins/GitHubWebHookCrumbExclusionTest.java
@@ -1,0 +1,67 @@
+package com.cloudbees.jenkins;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import javax.servlet.FilterChain;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+import static junit.framework.Assert.assertFalse;
+import static junit.framework.TestCase.assertTrue;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+public class GitHubWebHookCrumbExclusionTest {
+
+    private GitHubWebHookCrumbExclusion exclusion;
+    private HttpServletRequest req;
+    private HttpServletResponse resp;
+    private FilterChain chain;
+
+    @Before
+    public void before() {
+        exclusion = new GitHubWebHookCrumbExclusion();
+        req = mock(HttpServletRequest.class);
+        resp = mock(HttpServletResponse.class);
+        chain = mock(FilterChain.class);
+    }
+
+    @Test
+    public void testFullPath() throws Exception {
+        when(req.getPathInfo()).thenReturn("/github-webhook/");
+        assertTrue(exclusion.process(req, resp, chain));
+        verify(chain, times(1)).doFilter(req, resp);
+    }
+
+    @Test
+    public void testFullPathWithoutSlash() throws Exception {
+        when(req.getPathInfo()).thenReturn("/github-webhook");
+        assertTrue(exclusion.process(req, resp, chain));
+        verify(chain, times(1)).doFilter(req, resp);
+    }
+
+    @Test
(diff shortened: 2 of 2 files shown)


jglick:
What is the purpose of this? If Jenkins receives a request to `/github-webhook` it will simply redirect you to `/github-webhook/` before servicing `GitHubWebHook.doIndex`:

```
$ curl -IL https://jenkins.ci.cloudbees.com/github-webhook
HTTP/1.1 302 Found
Date: Mon, 07 Nov 2016 22:32:45 GMT
Expires: Thu, 01 Jan 1970 00:00:00 GMT
Location: https://jenkins.ci.cloudbees.com/github-webhook/
Server: Jetty(9.2.z-SNAPSHOT)
Set-Cookie: JSESSIONID.f61ecab1=1cu3g6hvkmjo21mdyuj1ixkhgu;Path=/;Secure;HttpOnly
X-Content-Type-Options: nosniff
Connection: keep-alive

HTTP/1.1 405 Method Not Allowed
Content-Type: text/plain;charset=UTF-8
Date: Mon, 07 Nov 2016 22:32:45 GMT
Server: Jetty(9.2.z-SNAPSHOT)
X-Content-Type-Options: nosniff
Connection: keep-alive

```

so you would better use the form ending with `/` to begin with.

KostyaSha:
I also don't understand profit of it. Url format should be also documented. A lot of APIs are sensitive and users should follow docs.

lanwen:
Agree with @KostyaSha. So many code to avoid user mistake...

i386:
@jglick from my testing it seems Github wasn't following the redirect.

michaelneale:
github does not follow the redirect. @lanwen its a net increase of 4 lines of code for something that bites users daily - is that a lot of code?

i386:
@KostyaSha @lanwen I strongly disagree - it's very easy to leave a slash off of the end of the URL and not pick up on why it's not working. This removes the problem all together at the cost of a few lines of code in the Git plugin. 

We can make more users satisfied with Jenkins if we are more forgiving with their mistakes. So closing this PR as WONTFIX is 🐛

daniel-beck:
I'll just leave this link here:
https://en.wikipedia.org/wiki/Robustness_principle

KostyaSha:
> github does not follow the redirect.

Where did you found this?

i386:
@KostyaSha https://github.com/isaacs/github/issues/574

michaelneale:
@KostyaSha years of using github with random webhook services that send 302's

KostyaSha:
Could you please stop editing your comments? I can't collect everything to follow.

src/main/java/com/cloudbees/jenkins/GitHubWebHookCrumbExclusion.java
```diff
@@ -16,11 +16,15 @@
...
             throws IOException, ServletException {
         String pathInfo = req.getPathInfo();
-        if (pathInfo != null && pathInfo.equals(getExclusionPath())) {
-            chain.doFilter(req, resp);
-            return true;
+        if (pathInfo == null || pathInfo.equals("")) {
+            return false;
         }
-        return false;
+        pathInfo = !pathInfo.endsWith("/") ? pathInfo + '/' : pathInfo;
```
KostyaSha: Please place comment here and with link to source of crazy GH API nuances.

src/main/java/com/cloudbees/jenkins/GitHubWebHookCrumbExclusion.java
```diff
@@ -16,11 +16,15 @@
     public boolean process(HttpServletRequest req, HttpServletResponse resp, FilterChain chain)
             throws IOException, ServletException {
         String pathInfo = req.getPathInfo();
-        if (pathInfo != null && pathInfo.equals(getExclusionPath())) {
-            chain.doFilter(req, resp);
-            return true;
+        if (pathInfo == null || pathInfo.equals("")) {
```
KostyaSha: Use `StringUtils.isEmpty()` (AFAIR) with static import

KostyaSha:
> @KostyaSha isaacs/github#574

Second time on this week i shocked how GH handle HTTP.

KostyaSha:
@michaelneale blocked me on GH so i have no notifications about his comments.

i386:
@KostyaSha I've updated the PR with the requested changes.

src/main/java/com/cloudbees/jenkins/GitHubWebHookCrumbExclusion.java
```diff
@@ -9,18 +9,25 @@
...
             throws IOException, ServletException {
         String pathInfo = req.getPathInfo();
-        if (pathInfo != null && pathInfo.equals(getExclusionPath())) {
-            chain.doFilter(req, resp);
-            return true;
+        if (isEmpty(pathInfo)) {
+            return false;
+        }
+        // Github will not follow redirects https://github.com/isaacs/github/issues/574
+        pathInfo = !pathInfo.endsWith("/") ? pathInfo + '/' : pathInfo;
```
KostyaSha: please swap arguments to exclude negative logic 
`pathInfo = pathInfo.endsWith("/") ? pathInfo : pathInfo + '/' ;`

src/main/java/com/cloudbees/jenkins/GitHubWebHookCrumbExclusion.java
```diff
@@ -9,18 +9,25 @@
...
             throws IOException, ServletException {
         String pathInfo = req.getPathInfo();
-        if (pathInfo != null && pathInfo.equals(getExclusionPath())) {
-            chain.doFilter(req, resp);
-            return true;
+        if (isEmpty(pathInfo)) {
+            return false;
+        }
+        // Github will not follow redirects https://github.com/isaacs/github/issues/574
+        pathInfo = pathInfo.endsWith("/") ?  pathInfo : pathInfo + '/';
```
KostyaSha: extra space...

i386: Fixed

src/main/java/com/cloudbees/jenkins/GitHubWebHookCrumbExclusion.java
i386: Fixed

i386:
Thanks @KostyaSha

KostyaSha:
Will keep for @lanwen review. 4am. 💤

i386:
Squashed these commits.

KostyaSha:
@i386 no need, github has squash button.

KostyaSha:
screenshot 2016-11-16 04 25 40

lanwen:
ok guys, you've convinced me

lanwen:
released as 1.23.1

michaelneale:
thanks! @lanwen hahaha yes. I have no idea why github don't follow, but I suspect that it doesn't as it adds latency that is "hidden" (my theory). Anyway, they do what they do, and they are the 400 pound gorrilla so we have to follow their rules ;)

i386:
Thank you!

PR #153, opened 2016-11-07
[FIXED JENKINS-39533] Make GitHub plugin BuildableItem aware

stephenc:
See [JENKINS-39533](https://issues.jenkins-ci.org/browse/JENKINS-39553)

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
+++ b/src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
@@ -377,8 +377,8 @@ public class GitHubPushTrigger extends Trigger<Job<?, ?>> implements GitHubTrigg
          * @since TODO
          */
         @SuppressWarnings("unused")
-        public FormValidation doCheckHookRegistered(@AncestorInPath Job<?, ?> job) {
-            Preconditions.checkNotNull(job, "Job can't be null if wants to check hook in monitor");
+        public FormValidation doCheckHookRegistered(@AncestorInPath Item job) {
+            Preconditions.checkNotNull(job, "Item can't be null if wants to check hook in monitor");
 
             Collection<GitHubRepositoryName> repos = GitHubRepositoryNameContributor.parseAssociatedNames(job);
 
--- a/src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
+++ b/src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
@@ -7,6 +7,7 @@ import hudson.ExtensionPoint;
 import hudson.Util;
 import hudson.model.AbstractProject;
 import hudson.model.EnvironmentContributor;
+import hudson.model.Item;
 import hudson.model.Job;
 import hudson.model.TaskListener;
 import hudson.plugins.git.GitSCM;
@@ -36,41 +37,54 @@ public abstract class GitHubRepositoryNameContributor implements ExtensionPoint 
      * Looks at the definition of {@link AbstractProject} and list up the related github repositories,
      * then puts them into the collection.
      *
-     * @deprecated Use {@link #parseAssociatedNames(Job, Collection)}
+     * @deprecated Use {@link #parseAssociatedNames(Item, Collection)}
      */
     @Deprecated
     public void parseAssociatedNames(AbstractProject<?, ?> job, Collection<GitHubRepositoryName> result) {
-        parseAssociatedNames((Job) job, result);
+        parseAssociatedNames((Item) job, result);
     }
 
     /**
      * Looks at the definition of {@link Job} and list up the related github repositories,
      * then puts them into the collection.
+     * @deprecated Use {@link #parseAssociatedNames(Item, Collection)}
      */
+    @Deprecated
     public /*abstract*/ void parseAssociatedNames(Job<?, ?> job, Collection<GitHubRepositoryName> result) {
-        if (overriddenMethodHasDeprecatedSignature(job)) {
-            parseAssociatedNames((AbstractProject) job, result);
-        } else {
-            throw new AbstractMethodError("you must override the new overload of parseAssociatedNames");
-        }
+        parseAssociatedNames((Item) job, result);
     }
 
     /**
-     * To select backward compatible method with old extensions
-     * with overridden {@link #parseAssociatedNames(AbstractProject, Collection)}
-     *
-     * @param job - parameter to check for old class
-     *
-     * @return true if overridden deprecated method
+     * Looks at the definition of {@link Item} and list up the related github repositories,
+     * then puts them into the collection.
      */
-    private boolean overriddenMethodHasDeprecatedSignature(Job<?, ?> job) {
-        return Util.isOverridden(
+    @SuppressWarnings("deprecation")
+    public /*abstract*/ void parseAssociatedNames(Item item, Collection<GitHubRepositoryName> result) {
+        if (Util.isOverridden(
+                GitHubRepositoryNameContributor.class,
+                getClass(),
+                "parseAssociatedNames",
+                Job.class,
+                Collection.class
+        )) {
+            // if this impl is legacy, it cannot contribute to non-jobs, so not an error
+            if (item instanceof Job) {
+                parseAssociatedNames((Job<?, ?>) item, result);
+            }
+        } else  if (Util.isOverridden(
                 GitHubRepositoryNameContributor.class,
                 getClass(),
                 "parseAssociatedNames",
                 AbstractProject.class,
                 Collection.class
-        ) && job instanceof AbstractProject;
+        )) {
+            // if this impl is legacy, it cannot contribute to non-projects, so not an error
+            if (item instanceof AbstractProject) {
+                parseAssociatedNames((AbstractProject<?, ?>) item, result);
+            }
+        } else {
+            throw new AbstractMethodError("you must override the new overload of parseAssociatedNames");
+        }
     }
 
     public static ExtensionList<GitHubRepositoryNameContributor> all() {
@@ -82,13 +96,21 @@ public abstract class GitHubRepositoryNameContributor implements ExtensionPoint 
      */
     @Deprecated
     public static Collection<GitHubRepositoryName> parseAssociatedNames(AbstractProject<?, ?> job) {
-        return parseAssociatedNames((Job) job);
+        return parseAssociatedNames((Item) job);
     }
 
+    /**
+     * @deprecated Use {@link #parseAssociatedNames(Item)}
+     */
+    @Deprecated
     public static Collection<GitHubRepositoryName> parseAssociatedNames(Job<?, ?> job) {
+        return parseAssociatedNames((Item) job);
+    }
+
+    public static Collection<GitHubRepositoryName> parseAssociatedNames(Item item) {
         Set<GitHubRepositoryName> names = new HashSet<GitHubRepositoryName>();
         for (GitHubRepositoryNameContributor c : all()) {
-            c.parseAssociatedNames(job, names);
+            c.parseAssociatedNames(item, names);
         }
         return names;
     }
@@ -99,9 +121,9 @@ public abstract class GitHubRepositoryNameContributor implements ExtensionPoint 
     @Extension
     public static class FromSCM extends GitHubRepositoryNameContributor {
         @Override
-        public void parseAssociatedNames(Job<?, ?> job, Collection<GitHubRepositoryName> result) {
+        public void parseAssociatedNames(Item job, Collection<GitHubRepositoryName> result) {
             SCMTriggerItem item = SCMTriggerItems.asSCMTriggerItem(job);
-            EnvVars envVars = buildEnv(job);
+            EnvVars envVars = job instanceof Job ? buildEnv((Job) job) : new EnvVars();
             if (item != null) {
                 for (SCM scm : item.getSCMs()) {
                     addRepositories(scm, envVars, result);
--- a/src/main/java/com/cloudbees/jenkins/GitHubWebHook.java
+++ b/src/main/java/com/cloudbees/jenkins/GitHubWebHook.java
@@ -3,6 +3,7 @@ package com.cloudbees.jenkins;
 import com.google.common.base.Function;
 import hudson.Extension;
 import hudson.ExtensionPoint;
+import hudson.model.Item;
 import hudson.model.Job;
 import hudson.model.RootAction;
 import hudson.model.UnprotectedRootAction;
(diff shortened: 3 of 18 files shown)


src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
```diff
@@ -36,41 +37,54 @@
...
                 AbstractProject.class,
                 Collection.class
-        ) && job instanceof AbstractProject;
+        )) {
+            // if this impl is legacy, it cannot contribute to non-projects, so not an error
+            if (item instanceof AbstractProject) {
+                parseAssociatedNames((AbstractProject<?, ?>) item, result);
+            }
+        } else {
+            throw new AbstractMethodError("you must override the new overload of parseAssociatedNames");
```
oleg-nenashev: 🐛 So it will expose the error when somebody uses Predicates like `withTrigger` e.g. in `DefaultPushGHEventSubscriber.#isApplicable()`. Instead of returning `false` this method will start throwing `Error` after the patch  if there is an incompatible contributor. Just if I read it right...

IMHO there should be one of the following fixes:
* Log warning and do nothing
* Replace the Error by non-runtime exception and implement correct handling in the upper logic

src/main/java/com/cloudbees/jenkins/GitHubWebHook.java
```diff
@@ -70,21 +71,35 @@ public String getUrlName() {
...
     }
 
     /**
+     * If any wants to auto-register hook, then should call this method
+     * Example code:
+     * {@code GitHubWebHook.get().registerHookFor(item);}
+     *
+     * @param item not null item to register hook for
+     */
+    public void registerHookFor(Item item) {
```
oleg-nenashev: `@since`

src/main/java/org/jenkinsci/plugins/github/extension/GHEventsSubscriber.java
```diff
@@ -32,16 +35,71 @@
  */
 public abstract class GHEventsSubscriber implements ExtensionPoint {
     private static final Logger LOGGER = LoggerFactory.getLogger(GHEventsSubscriber.class);
+    private transient Boolean hasIsApplicableItem;
```
oleg-nenashev: I would explicitly add `CheckForNull`, just for the future

src/main/java/org/jenkinsci/plugins/github/util/FluentIterableWrapper.java
```diff
@@ -80,6 +80,15 @@
     }
 
     /**
+     * Returns the elements from this fluent iterable that are instances of the supplied type. The
+     * resulting fluent iterable's iterator does not support {@code remove()}.
+     */
+    @CheckReturnValue
+    public final <F extends E> FluentIterableWrapper<F> filter(Class<F> clazz) {
```
oleg-nenashev: `@since`

src/main/java/org/jenkinsci/plugins/github/util/JobInfoHelpers.java
```diff
@@ -33,49 +35,49 @@ private JobInfoHelpers() {
      *
      * @return predicate with true on apply if job contains trigger of given class
      */
-    public static Predicate<Job> withTrigger(final Class<? extends Trigger> clazz) {
-        return new Predicate<Job>() {
-            public boolean apply(Job job) {
-                return triggerFrom(job, clazz) != null;
+    public static <ITEM extends Item> Predicate<ITEM> withTrigger(final Class<? extends Trigger> clazz) {
```
oleg-nenashev: I'm not 100% sure it's a binary compatible change, same for below. Need to check it

src/main/java/org/jenkinsci/plugins/github/util/JobInfoHelpers.java
```diff
@@ -87,11 +89,26 @@ public boolean apply(Job job) {
...
+    /**
+     * @param item    job to search trigger in
+     * @param tClass trigger with class which we want to receive from job
+     * @param <T>    type of trigger
+     *
+     * @return Trigger instance with required class or null
+     * TODO use standard method in 1.621+
+     */
+    @CheckForNull
+    public static <T extends Trigger> T triggerFrom(Item item, Class<T> tClass) {
```
oleg-nenashev: `@since`

src/main/java/org/jenkinsci/plugins/github/webhook/WebhookManager.java
```diff
@@ -79,24 +80,45 @@ public static WebhookManager forHookUrl(URL endpoint) {
...
+     * New events will be merged with old ones from existent hook.
+     *
+     * By default only push event is registered
+     *
+     * @param item to find for which repos we should create hooks
+     *
+     * @return runnable to create hooks on run
+     * @see #createHookSubscribedTo(List)
+     */
+    public Runnable registerFor(final Item item) {
```
oleg-nenashev: `@since`

src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
stephenc: Actually this is more a design error in the original API. These methods taking the collection should actually have been `protected` as they form part of the SPI not the API.

If you have an old implementation, then you will have overridden `parseAssociatedNames(Job,Collection)` or `parseAssociatedNames(AbstractProject,Collection)` and thus we delegate to those.

If you have a new implementation, then you are supposed to override `parseAssociatedNames(Item,Collection)` and if you don't then this exception should blow up in your face and give you the AME.

If we had these SPI methods as `protected` then we could make `parseAssociatedNames(Item,Collection)` abstract and use the API methods like `parseAssociatedNames(Item)` to route through to the implementation of existing extension points... but instead, as this is public API we have to do it this way and we have no way to force new implementations to override `parseAssociatedNames(Item,Collection)` other than this exception

src/main/java/org/jenkinsci/plugins/github/util/JobInfoHelpers.java
stephenc: The type gets erased. It may cause compiler errors if compiling against the newer API but the erased type is still `Predicate`

stephenc:
@lanwen can we get this merged and released?

KostyaSha:
@stephenc when it will be reviewed and verified

stephenc:
@KostyaSha the changes are good, oleg just misunderstood... right now this is required to get the github-branch-source 2.0 release out.

KostyaSha:
@stephenc situation is that all existing github plugins was ignored, we also suggested to use shared connection and other things, but your multibranch still inventing own things. Now different stuff is pushed into plugins and we will be forced to maintain it. That's look not very good, my gh plugins fully using github-plugin so i need firstly identify that it wouldn't break anything.  github-api-plugin already got portion of sh.. without answers https://github.com/jenkinsci/github-api-plugin/commit/185541aded1daeb3e549595d0008629a950470dd and i already raised questions to jenkinsci board why non-maintainer people making changes into plugin. 

@stephenc i will try review as soon as i can.

stephenc:
@KostyaSha I am not happy with some of the mess that I have found, e.g. GitHub org folders plugin should never have been released... I'm working on getting these aligned better, but it's a marathon not a sprint.

KostyaSha:
> @KostyaSha I am not happy with some of the mess that I have found,

Compare what is done in CB plugins and what exists and working well for a year here https://github.com/KostyaSha/github-integration-plugin (it was under jenkinsci and was planned to go into one powerful github-plugin, but i was forced to go out of jenkins infra).

src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
stephenc: Also please check the previous implementation of `parseAssociatedNames(Job<?, ?> job, Collection<GitHubRepositoryName> result)` which threw the exact same exception. All I am doing here is moving the exception throwing from the previously most generic method to the new most generic method.

The exception will only be thrown if you have an implementation that does not override *any* `parseAssociatedNames` methods. If this had been designed as an SPI then the implementation methods would be protected and we could have shifted the abstract and used a guard.

To clarify for @KostyaSha this is not a bug just migration of existing pattern

stephenc:
@KostyaSha said:

> Oleg's comments are not addressed

Incorrect, I fixed or added a rationale for every one of oleg's comments

src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
KostyaSha: @stephenc only since

stephenc:
FTR this is now blocking releasing Branch API 2.0, so if we could get some progress on a decision it would be super awesome as I have a lot of plugins to juggle getting releases at or near the same time

KostyaSha:
And please don't merge branches into branch this complicates search in tree.

stephenc:
Sorry merge commit got pushed by accident when I was trying to diagnose the events being missing (thank you fail2ban... though not having these changes was also a culprit, fail2ban was stopping them even getting to my server)

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -377,8 +379,9 @@ private static ThreadFactory threadFactory() {
          * @since TODO
          */
         @SuppressWarnings("unused")
-        public FormValidation doCheckHookRegistered(@AncestorInPath Job<?, ?> job) {
-            Preconditions.checkNotNull(job, "Job can't be null if wants to check hook in monitor");
+        @Restricted(NoExternalUse.class) // invoked from Stapler
+        public FormValidation doCheckHookRegistered(@AncestorInPath Item job) {
```
KostyaSha: rename job to item

src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
```diff
@@ -99,9 +124,9 @@ private boolean overriddenMethodHasDeprecatedSignature(Job<?, ?> job) {
     @Extension
     public static class FromSCM extends GitHubRepositoryNameContributor {
         @Override
-        public void parseAssociatedNames(Job<?, ?> job, Collection<GitHubRepositoryName> result) {
+        public void parseAssociatedNames(Item job, Collection<GitHubRepositoryName> result) {
```
KostyaSha: job -> item

stephenc:
@oleg-nenashev yes, the correct solution is to use the SPI pattern like I have done in SCM API where the methods that implementers are required to implement are `protected` and the methods that consumers are supposed to call are `public final`.

Then when you need to adapt the API, what you do is have the `public final` check the current `protected abstract` and fall back through previous SPI methods. You will be guaranteed to get a match as:

* A new implementation will not compile unless they implement the current `protected abstract`
* An old implementation has to have implemented one of the previous `protected abstract` that you have now deprecated or they would not have compiled against the old SPI

stephenc:
@reviewbybees done

KostyaSha:
Have you tried open jenkins global settings?

stephenc:
@KostyaSha you mean the `Caused by: java.lang.NoSuchMethodError: jenkins.model.Jenkins.getActiveInstance()Ljenkins/model/Jenkins;
` errors because this needs the bump to 1.609 baseline to work with credentials (which is already in master but you didn't want me to include the merge commit?

KostyaSha:
I placed new gh plugin version into my test instance and it failed with 
```
target/tmp/webapp/WEB-INF/lib/jenkins-core-1.625.3.jar!/lib/form/rowSet.jelly:47:23: <d:invokeBody> No such property: configuration for class: org.jenkinsci.plugins.github.GitHubPlugin
	at 
```

stephenc:
@KostyaSha are you sure that isn't an issue for you anyway with either master or the base version this PR was created against? there is no change in this PR that should create that issue for you

KostyaSha:
i'm checking around...

KostyaSha:
revert to 1.22.1 works.

stephenc:
Well this PR is off of 1.22.4 so I claim your issue is outside of the scope of this PR

KostyaSha:
1.22.4 also works.

KostyaSha:
```
Caused by: 
java.lang.NoSuchMethodError: com.cloudbees.plugins.credentials.common.StandardListBoxModel.includeEmptyValue()Lcom/cloudbees/plugins/credentials/common/AbstractIdCredentialsListBoxModel;
	at org.jenkinsci.plugins.github.config.GitHubServerConfig$DescriptorImpl.doFillCredentialsIdItems(GitHubServerConfig.java:291)
```

stephenc:
Have you upgraded credentials?

src/main/java/org/jenkinsci/plugins/github/extension/GHEventsSubscriber.java
```diff
@@ -32,16 +36,72 @@
...
+        if (hasIsApplicableItem == null) {
+            boolean implemented = false;
+            // cannot use Util.isOverridden because method is protected and isOverridden only checks public methods
+            Class<?> clazz = getClass();
+            while (clazz != null && clazz != GHEventsSubscriber.class) {
+                try {
+                    Method isApplicable = clazz.getDeclaredMethod("isApplicable", Item.class);
+                    if (isApplicable.getDeclaringClass() != GHEventsSubscriber.class) {
+                        // ok this is the first method we have found that could be an override
+                        // if somebody overrode an inherited method with and `abstract` then we don't have the method
```
KostyaSha: `override`?

KostyaSha:
> Have you upgraded credentials?

seems no, i'm using `hpi:run` from my plugin and comparing on working variant. So yes, maybe caused by super latest requirement.

KostyaSha:
hpi from pr build seems working, looks like something strange in your PR head even after rebase

KostyaSha:
Workflow messed all plugins to work with Job that is not suitable in many cases. Now github-plugin should extend to `Item`. I see no any methods that could be useful for `Item`. Why top-level hierarchy item should deal with gh?? If this PR about `BuildableItem`, then it should be `BuildableItem` . And even in `BuildableItem` i see no any useful types for storing something useful. There is no properties, actions. Only methods to build and display. So type expansion is under huge question.  

`AbstractTopLevelItem`/`AbstractItem` at least has actions!

KostyaSha:
screenshot 2016-12-12 00 22 24
I reviewed all methods and types. Maybe i see bit more than core provides. I can switch to core sources.

stephenc:
@KostyaSha this is not about workflow, but about removing the assumption that the only thing that would want to know about events are jobs and that the only things to trigger are Jobs.

For example https://github.com/jenkinsci/cloudbees-folder-plugin/blob/41f1e2d63f8e4ef82bb845a01fc81ea5e427f465/src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java#L97 is a BuildableItem and may want to listen for events from the event handler in order to generate items based on the changes.

stephenc:
IMHO we are better allowing the events to be picked up by any generic `Item` (as that way we can at least mix-in the Item interface rather than relying on the abstract base class.

The trigger should work on anything that is a `BuildableItem` because that is where a trigger makes sense

KostyaSha:
Ideally should be `TrigerableItem` :/ Though if `item` wants and subscriber can get. Then ok. Penalty is that now hook iteration will deal with `Jenkins.getAllItems()` that literally is all items.

KostyaSha:
I just fear another brainf.. like Parametrized.JobMixIn that doesn't allow to mock tests because methods needs crazy generics and etc. 

abstract method (Item) picks for old binary child abstract(Job) calls, right?

stephenc:
Thanks for merging. Any estimate on when a release can be cut?

KostyaSha:
Fixing different since. Maybe it possible add empty checkstyle in parent pom that will have single `@since {TODO/FIXME//}` on fail build level?

KostyaSha:
released.

PR #154, opened 2016-11-07
[FIXED JENKINS-35550] Upgrade to Credentials 2.1.0+ API for populating credentials drop-down

stephenc:
See [JENKINS-35550](https://issues.jenkins-ci.org/browse/JENKINS-35550)

- In two of the cases the context is explicitly `Jenkins.getInstance()` and no user credentials should be considered.
- In one case - conversion - then user credentials stores should be included

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/154)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -46,8 +46,8 @@
     </issueManagement>
 
     <properties>
-        <jenkins.version>1.580</jenkins.version>
-        <jenkins-test-harness.version>1.580</jenkins-test-harness.version>
+        <jenkins.version>1.609</jenkins.version>
+        <jenkins-test-harness.version>1.609</jenkins-test-harness.version>
         <release.skipTests>false</release.skipTests>
         <maven.javadoc.skip>true</maven.javadoc.skip>
         <findbugs-maven-plugin.version>3.0.2</findbugs-maven-plugin.version>
@@ -105,7 +105,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>credentials</artifactId>
-            <version>1.22</version>
+            <version>2.1.8</version>
         </dependency>
 
         <dependency>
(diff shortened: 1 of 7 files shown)


pom.xml
```diff
@@ -105,7 +105,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>credentials</artifactId>
-            <version>1.22</version>
+            <version>2.1.8</version>
```
KostyaSha: why??

stephenc: recommended version in the 2.1.x series

KostyaSha: So it will require everybody update to latest credentials without alternative?

stephenc: Yes. if you want to go down as far as 2.1.6 you can... but the diff between 2.1.6 and 2.1.8 is recommended given that I have not seen any issues in 2.1.8 in over 2 weeks

stephenc: so 2.1.8 was soaked for 2 weeks now... no new bugs on top of 2.1.7... so I am declaring it the preferred version... there is still some issues that plugins have with the 2.1.7+ versions when there is a plugin that iterates an `ExtensionList` before `InitMilestone.EXTENSIONS_AUGMENTED` and can therefore see partially initialized plugins... but that is a problem with the plugins doing bold things, not the credentials plugin ;-)

lanwen:
will merge and release it at friday

pom.xml
KostyaSha: You shouldn't require the latest if there is no reason in APIs. Git 3.0 was so cool that i was forced to revert it to 2.x and hopefully nobody enforced latest version. So revert to latest possible according to APIs.

KostyaSha:
https://github.com/jenkinsci/github-plugin/pull/154#discussion_r87164069

lanwen:
@KostyaSha github now provides ability to "Require changes" if you wait to fix smth after review.

Also - this fix requires 2.1+ api, so why not to use latest stable release?

KostyaSha:
I provided real life example. Now there is no any change to use non latest version.

PR #155, opened 2016-11-08
[JENKINS-39590] Switch to `GitHub.parseEventPayload` for event parsing

stephenc:
See [JENKINS-39590](https://issues.jenkins-ci.org/browse/JENKINS-39590)

Work In Progress until https://github.com/kohsuke/github-api/pull/306 is merged and https://github.com/jenkinsci/github-api-plugin/ gets a release (also needs to be on top of https://github.com/jenkinsci/github-plugin/pull/153)

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/155)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -93,7 +93,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>github-api</artifactId>
-            <version>1.69</version>
+            <version>1.80</version>
         </dependency>
 
         <dependency>
--- a/src/main/java/org/jenkinsci/plugins/github/webhook/subscriber/DefaultPushGHEventSubscriber.java
+++ b/src/main/java/org/jenkinsci/plugins/github/webhook/subscriber/DefaultPushGHEventSubscriber.java
@@ -8,10 +8,14 @@ import com.cloudbees.jenkins.GitHubWebHook;
 import hudson.Extension;
 import hudson.model.Item;
 import hudson.security.ACL;
+import java.io.IOException;
+import java.io.StringReader;
+import java.net.URL;
 import jenkins.model.Jenkins;
-import net.sf.json.JSONObject;
 import org.jenkinsci.plugins.github.extension.GHEventsSubscriber;
 import org.kohsuke.github.GHEvent;
+import org.kohsuke.github.GHEventPayload;
+import org.kohsuke.github.GitHub;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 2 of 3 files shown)


pom.xml
```diff
@@ -46,8 +46,8 @@
     </issueManagement>
 
     <properties>
-        <jenkins.version>1.580</jenkins.version>
-        <jenkins-test-harness.version>1.580</jenkins-test-harness.version>
+        <jenkins.version>1.615</jenkins.version>
```
oleg-nenashev: I would vouch for using LTS baseline like `1.625` since you go beyond the Java 6 compat threshold anyway

src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
```diff
@@ -36,41 +37,54 @@
...
                 AbstractProject.class,
                 Collection.class
-        ) && job instanceof AbstractProject;
+        )) {
+            // if this impl is legacy, it cannot contribute to non-projects, so not an error
+            if (item instanceof AbstractProject) {
+                parseAssociatedNames((AbstractProject<?, ?>) item, result);
+            }
+        } else {
+            throw new AbstractMethodError("you must override the new overload of parseAssociatedNames");
```
oleg-nenashev: see another PR

stephenc:
@oleg-nenashev I created this PR (and tagged as WIP) so that the *upstream* PRs can be reviewed with an intended usage example... i.e. This PR

stephenc:
@reviewbybees

stephenc:
@reviewbybees ping

src/main/java/org/jenkinsci/plugins/github/webhook/subscriber/PingGHEventSubscriber.java
```diff
@@ -59,18 +61,23 @@ protected boolean isApplicable(Item project) {
      */
     @Override
     protected void onEvent(GHEvent event, String payload) {
```
KostyaSha: Does onEvent report errors itself? Then catch not needed.

KostyaSha: https://github.com/jenkinsci/github-plugin/blob/93d40692ff3866705175624e93ec584d4ac88132/src/main/java/org/jenkinsci/plugins/github/extension/GHEventsSubscriber.java#L208-L214
so try-catch useless wrapper, undo change and check that parser reports meaningful error.

stephenc: That Throwable catch should be the last ditch catch. Propagation to it is not a good pattern.

I'll see if I can decipher what you mean tomorrow on computer not phone. But in general you should catch and log as close and as specific as possible while you know what sense to make...

KostyaSha: In just you can trust to any underlying extension as it may throw NPE, so catching throwables and reporting is standard save ass pattern. If exception throws meaningful information then additional exception steps doesn't add anything useful for debugging.

lanwen:
LGTM

src/main/java/org/jenkinsci/plugins/github/webhook/subscriber/DefaultPushGHEventSubscriber.java
```diff
@@ -78,13 +87,15 @@ public void run() {
                     for (Item job : Jenkins.getInstance().getAllItems(Item.class)) {
                         GitHubTrigger trigger = triggerFrom(job, GitHubPushTrigger.class);
                         if (trigger != null) {
-                            LOGGER.debug("Considering to poke {}", job.getFullDisplayName());
-                            if (GitHubRepositoryNameContributor.parseAssociatedNames(job).contains(changedRepository)) {
-                                LOGGER.info("Poked {}", job.getFullDisplayName());
+                            String fullDisplayName = job.getFullDisplayName();
+                            LOGGER.debug("Considering to poke {}", fullDisplayName);
+                            if (GitHubRepositoryNameContributor.parseAssociatedNames(job)
+                                    .contains(changedRepository)) {
```
jglick: Please avoid gratuitous reformatting of otherwise unmodified lines.

stephenc:
@reviewbybees done

lanwen:
released as 1.25.1

PR #157, opened 2016-11-17
[Jenkins-39822] GitHub plugin functional tests broken against 1.651+

raul-arabaolaza:
[JENKINS-39822](https://issues.jenkins-ci.org/browse/JENKINS-39822)

@reviewbybees

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/157)

### DIFF
--- a/src/test/java/com/cloudbees/jenkins/GitHubCommitNotifierTest.java
+++ b/src/test/java/com/cloudbees/jenkins/GitHubCommitNotifierTest.java
@@ -12,6 +12,7 @@ import hudson.model.Result;
 import hudson.plugins.git.GitSCM;
 import hudson.plugins.git.Revision;
 import hudson.plugins.git.util.BuildData;
+import hudson.util.VersionNumber;
 import org.eclipse.jgit.lib.ObjectId;
 import org.jenkinsci.plugins.github.config.GitHubPluginConfig;
 import org.jenkinsci.plugins.github.test.GHMockRule;
@@ -97,7 +98,12 @@ public class GitHubCommitNotifierTest {
         FreeStyleProject prj = jRule.createFreeStyleProject();
         prj.setScm(new GitSCM("http://non.existent.git.repo.nowhere/repo.git"));
         prj.getPublishersList().add(new GitHubCommitNotifier());
-        Build b = prj.scheduleBuild2(0, new Cause.UserIdCause(), new BuildData()).get();
+        Build b = null;
+        if (jRule.getPluginManager().getPlugin("git").getVersionNumber().isNewerThan(new VersionNumber("2.4.0"))) {
+            b = prj.scheduleBuild2(0, new Cause.UserIdCause(), new BuildData()).get();
+        } else {
+            b = prj.scheduleBuild2(0).get();
+        }
         jRule.assertBuildStatus(Result.FAILURE, b);
         jRule.assertLogContains(BuildDataHelper_NoLastRevisionError(), b);
     }
--- a/src/test/java/com/cloudbees/jenkins/GitHubCommitNotifierTest.java
+++ b/src/test/java/com/cloudbees/jenkins/GitHubCommitNotifierTest.java
@@ -98,6 +98,7 @@ public class GitHubCommitNotifierTest {
         FreeStyleProject prj = jRule.createFreeStyleProject();
         prj.setScm(new GitSCM("http://non.existent.git.repo.nowhere/repo.git"));
         prj.getPublishersList().add(new GitHubCommitNotifier());
+        //Git plugin 2.4.1 + does not include BuildData if checkout fails, so we add it if needed
         Build b = safelyGenerateBuild(prj);
         jRule.assertBuildStatus(Result.FAILURE, b);
         jRule.assertLogContains(BuildDataHelper_NoLastRevisionError(), b);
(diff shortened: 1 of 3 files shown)


KostyaSha:
Plugin is on 1.609 and tests are fine. Security 170 could be fixed in other way. 
👎

raul-arabaolaza:
@KostyaSha I understand, but given that these changes do not upgrade the baseline nor they broke the existing tests I thought it could be interesting to add them.

I do not understand the SECURITY-170 comment, are you referring that you prefer a separate PR for that or is that you prefer another solution? If you prefer another solution can you give me a hint, please?

src/test/java/com/cloudbees/jenkins/GitHubCommitNotifierTest.java
```diff
@@ -96,7 +97,7 @@ public void testNoBuildRevision() throws Exception {
         FreeStyleProject prj = jRule.createFreeStyleProject();
         prj.setScm(new GitSCM("http://non.existent.git.repo.nowhere/repo.git"));
         prj.getPublishersList().add(new GitHubCommitNotifier());
-        Build b = prj.scheduleBuild2(0).get();
+        Build b = prj.scheduleBuild2(0, new Cause.UserIdCause(), new BuildData()).get();
```
KostyaSha: Why it broken in 2.0?

raul-arabaolaza: On 2.x when the `testNoBuidRevision` executes there is no `BuildData` attached to the run,   `build.getAction(BuildData.class); ` returns null on `ObjectId#getCommitSHA1` so the end result is exactly the same as when `testNoBuildData` is executed 

So instead of trying to mock GitHub or change any version I thought the best curse of action is to make sure there is always a non null `BuildData` on that test

KostyaSha: So maybe create conditional on jenkins version?

src/test/java/org/jenkinsci/plugins/github/common/ExpandableMessageTest.java
```diff
@@ -43,6 +46,9 @@ public void shouldExpandEnvAndBuildVars() throws Exception {
         ));
 
         FreeStyleProject job = jRule.createFreeStyleProject();
+        ParameterDefinition paramDef = new StringParameterDefinition(CUSTOM_BUILD_PARAM, "", "");
+        ParametersDefinitionProperty paramsDef = new ParametersDefinitionProperty(paramDef);
+        job.addProperty(paramsDef);
```
KostyaSha: and conditional here

KostyaSha:
> I do not understand the SECURITY-170 comment,

If github-plugin inject some parameters, then they should be fixed in other way instead of injecting in job bypassing properties. Good example is in https://github.com/jenkinsci/gerrit-trigger-plugin/pull/285/files#diff-54e50e5c447d3a893fd786bd419ee8b4R264
i would even like to see this code as separate plugin.

raul-arabaolaza:
@KostyaSha Will work on your requested changes. Thanks for your time

lanwen:
> If github-plugin inject some parameters 

Its not. Just expands some vars for status setter

raul-arabaolaza:
@KostyaSha It seems that the null build data problem does not come from jenkins core but git plugin. To be able to execute against core 2.19.3 you have to use a newer version of git. Git plugin since 2.4.1 does not store BuildData if checkout goes wrong (see [this](https://github.com/jenkinsci/git-plugin/commit/88739e102c29d405eb4edf587c38eaa24ab25566#diff-f1f2ff967f38c8b53a4901be3169035e) commit). That means a conditional does not make sense, unless is removed when git dependency is upgraded, which IMHO results in an unneeded TODO and ugly code

So, this PR does not make the existing test broke and if for some reason git dependency is upgraded it makes the test work against it without code changes, but is unnecessary for the current version of the dependency. Do not know if you plan to upgrade git dependency shortly (to 2.4.4 for example) or not. If you plan that it may made sense to continue with this PR, if not it may not and I will change it to only fix the SECURITY-170 issue. Opinions?

lanwen:
I think we can try to update to git plugin of latest 2.x.x

KostyaSha:
Just add conditional on plugin version. And wrap this place. Examples should be somewhere in core/plugins.

raul-arabaolaza:
@KostyaSha @lanwen Tried to address your comments. Is the code better now??

lanwen:
Wow, as for me previous version was much better, @KostyaSha

KostyaSha:
> Wow, as for me previous version was much better, @KostyaSha

After few years you will say thanks :D but still there is no meaningful description WHY this hacks are needed. Security170 is not a description, what exactly and why?

src/test/java/org/jenkinsci/plugins/github/common/ParametersActionHelper.java
```diff
@@ -0,0 +1,44 @@
+package org.jenkinsci.plugins.github.common;
+
+import hudson.model.ParametersAction;
+
+import java.lang.reflect.Field;
+import java.lang.reflect.Modifier;
+
+public class ParametersActionHelper {
```
lanwen: please add javadoc with complete description for such classes

raul-arabaolaza:
@KostyaSha @lanwen  Clarifying comments and javadoc added.

lanwen:
Please also make codacy happy :)
https://www.codacy.com/app/lanwen/github-plugin/pullRequest?prid=422575

KostyaSha:
@lanwen maybe comments should be in Java doc for methods?

lanwen:
No, i think in this case they should be inlined

PR #159, opened 2016-11-20
add field for setting context for pending step

koenpunt:
I've updated the pending step with the context field also available in the after build step.

But I couldn't get the failing test updated to make it work again, so if anyone can give me some pointers there? 

I'm totally new to java, and somethings are easy to derive from existing code, but I actually have no idea what I'm doing 💩. But the implementation is functional in my installation.

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/159)

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/GitHubSetCommitStatusBuilder.java
+++ b/src/main/java/com/cloudbees/jenkins/GitHubSetCommitStatusBuilder.java
@@ -11,6 +11,7 @@ import hudson.tasks.BuildStepDescriptor;
 import hudson.tasks.Builder;
 import jenkins.tasks.SimpleBuildStep;
 import org.jenkinsci.plugins.github.common.ExpandableMessage;
+import org.jenkinsci.plugins.github.extension.status.GitHubStatusContextSource;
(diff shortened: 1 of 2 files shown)


lanwen:
Please add read resolve method to handle null-deserialization
Example:
https://github.com/jenkinsci/github-plugin/blob/master/src/main/java/org/jenkinsci/plugins/github/status/GitHubCommitStatusSetter.java#L173-L178

src/main/java/com/cloudbees/jenkins/GitHubSetCommitStatusBuilder.java
```diff
@@ -48,13 +50,25 @@ public ExpandableMessage getStatusMessage() {
...
+        return contextSource;
+    }
+
+    /**
      * @since 1.14.1
      */
     @DataBoundSetter
     public void setStatusMessage(ExpandableMessage statusMessage) {
         this.statusMessage = statusMessage;
     }
```
lanwen: please add `since FIXME` javadoc tag, i'll fix actual version before release

koenpunt:
I've updated the code per the requested changes. Should I squash some of these commits?

lanwen:
No, github can squash by itself

PR #160, opened 2016-11-28
Make token-macro optional

KostyaSha:
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/160)


KostyaSha:
Found, you didn't migrate conditional resolver to make token-macro optional....

PR #164, opened 2017-02-07
[FIXED JENKINS-41811] Expose event origin to listeners

stephenc:
See [JENKINS-41811](https://issues.jenkins-ci.org/browse/JENKINS-41811)

Downstream of https://github.com/jenkinsci/scm-api-plugin/pull/27

- [x] Remove WIP label once https://github.com/jenkinsci/scm-api-plugin/pull/27 merged and released and this PR updated to release version

Upstream of changes in GitHub Branch Source to start exposing the event origin

See also https://github.com/jenkinsci/branch-api-plugin/pull/79 (includes screenshots of events from GitHub Branch Source with this PRs changes integrated) 

@reviewbybees

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/164)


pom.xml
```diff
@@ -99,7 +99,13 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>git</artifactId>
-            <version>2.4.0</version>
+            <version>2.4.0</version> <!-- should really be 2.6.4 to match scm-api but ok as git wass forward compat -->
```
stephenc: Workflow tests need rework for the updated git dependency, but as git plugin didn't break the rules of SCM API we don't *need* to update to 2.6.4 / 3.0.4... just would be missing the SCMEvent integration in git plugin, but the tests in this plugin do not rely on that functionality, so no harm

src/main/java/com/cloudbees/jenkins/GitHubWebHook.java
```diff
@@ -153,7 +157,10 @@ public static Jenkins getJenkinsInstance() throws IllegalStateException {
      * Other plugins may be interested in listening for these updates.
      *
      * @since 1.8
+     * @deprecated we do not think this API is required any more, if we are wrong, please raise a JIRA
      */
+    @Deprecated
+    @Restricted(NoExternalUse.class)
```
stephenc: From what I can see the only consumer of this API is the 1.x GitHub Branch Source. If there are other use cases for consumers remaining post SCM API 2.0.x updates then this would need to be augmented to pass through the origin.

I'd much rather kill off the code path, so by adding the `@Restricted` we will flush out any impls without breaking them until they update their dependency

pom.xml
```diff
@@ -99,7 +99,13 @@
...
+        </dependency>
+
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>scm-api</artifactId>
+            <version>2.0.3-SNAPSHOT</version>
```
KostyaSha: 👎  on scm-api dependency

src/main/java/com/cloudbees/jenkins/GitHubTrigger.java
```diff
@@ -24,6 +24,8 @@
     // TODO: document me
     void onPost(String triggeredByUser);
 
+    void onPost(String origin, String triggeredByUser);
```
KostyaSha: should wait until java8

src/main/java/com/cloudbees/jenkins/GitHubWebHook.java
```diff
@@ -153,7 +157,10 @@ public static Jenkins getJenkinsInstance() throws IllegalStateException {
      * Other plugins may be interested in listening for these updates.
      *
      * @since 1.8
+     * @deprecated we do not think this API is required any more, if we are wrong, please raise a JIRA
```
KostyaSha: we? You?

pom.xml
jglick: typo

pom.xml
jglick: preferably specify a timestamp

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -87,6 +96,7 @@ private boolean runPolling() {
                         PrintStream logger = listener.getLogger();
                         long start = System.currentTimeMillis();
                         logger.println("Started on " + DateFormat.getDateTimeInstance().format(new Date()));
+                        logger.println("Started by event from " + origin);
```
jglick: Can `origin` be `null`? Or would it be `"?"` here?

src/main/java/com/cloudbees/jenkins/GitHubTrigger.java
jglick: Could instead introduce an extension `interface`.

src/main/java/com/cloudbees/jenkins/GitHubWebHook.java
jglick: [the list](https://wiki.jenkins-ci.org/display/JENKINS/Extension+points#Extensionpoints-com.cloudbees.jenkins.GitHubWebHook.Listener)

src/main/java/org/jenkinsci/plugins/github/webhook/subscriber/DefaultPushGHEventSubscriber.java
```diff
@@ -92,7 +93,7 @@ public void run() {
                             if (GitHubRepositoryNameContributor.parseAssociatedNames(job)
                                     .contains(changedRepository)) {
                                 LOGGER.info("Poked {}", fullDisplayName);
-                                trigger.onPost(pusherName);
+                                trigger.onPost(origin, pusherName);
```
jglick: Must at least catch `AbstractMethodError` here.

pom.xml
jglick: Why? `scm-api` is harmless—it has no UI.

KostyaSha: Because it adds useless dependency that participate in multibranch/branch api chaos with cycle dependencies and huge pages with instructions what to do with broken instances. I also have gh plugins depending on github-plugin and i don't want to deal with redundant dependencies. github-plugin as trigger and eventing mechanism may work without additional deps. Freely add Listener calls instead of deprecating it.  I would also like to see firstly such changes in gitlab and other trigger plugins, not only gh.

src/main/java/com/cloudbees/jenkins/GitHubWebHook.java
stephenc: @jglick that single impl is not in GHBS 2.0.x

pom.xml
stephenc: @KostyaSha you already have the dependency via git 2.4.0 (but just an older version) and we need the `SCMEvent.originOf(req)` helper method

stephenc:
@KostyaSha I am not adding a dependency on scm-api, the current [transitive dependency](https://github.com/jenkinsci/git-plugin/blob/git-2.4.0/pom.xml#L258-L262) on scm-api is 

```
    <dependency>
      <artifactId>scm-api</artifactId>
      <version>0.2</version>
    </dependency>
```

KostyaSha:
> @KostyaSha I am not adding a dependency on scm-api, the current transitive dependency on scm-api is

That's terrible.

stephenc:
@KostyaSha 
> That's terrible.

exactly, that version is so ancient which is why we need to bump it ;-)

KostyaSha:
typed huge sentence on how i disappointed with jenkins & CB and deleted.

src/main/java/com/cloudbees/jenkins/GitHubTrigger.java
```diff
@@ -15,6 +15,7 @@
  * and triggers a build.
  *
  * @author aaronwalker
+ * @deprecated extend {@link GitHubTrigger2} instead
```
jglick: Careful—does this produce deprecation warnings for implementers of `GitHubTrigger2`?

src/main/java/org/jenkinsci/plugins/github/webhook/subscriber/DefaultPushGHEventSubscriber.java
```diff
@@ -4,6 +4,7 @@
 import com.cloudbees.jenkins.GitHubRepositoryName;
 import com.cloudbees.jenkins.GitHubRepositoryNameContributor;
 import com.cloudbees.jenkins.GitHubTrigger;
+import com.cloudbees.jenkins.GitHubTrigger2;
```
jglick: Unused import AFAICT.

src/main/java/com/cloudbees/jenkins/GitHubTrigger.java
stephenc: Who has implemented it? it is doubtful the interface is even required.

stephenc: Doesn't show as deprecated in my IDE
screen shot 2017-02-07 at 19 10 08

KostyaSha:
So just to show useless information in UI you are adding useless hard dependencies, deprecating and hacking everything around? That will all end that nobody outside CB will be able to consume and implement plugins based on such code. For example credentials-plugin - nobody knows how to use it.

stephenc:
@KostyaSha 

> typed huge sentence on how i disappointed with jenkins & CB and deleted.

Well writing out your anger certainly helps process it. But SCM API update was only a problem for GitHub Branch Source and BitBucket Branch Source plugins. 

I didn't even need to update Subversion as it was fine with the updated SCM API.

I only updated GIT and Mercurial because they could both significantly benefit from the event apis, (and because I needed to in order to fix GitHub Branch Source and BitBucket Branch Source plugins), but from a "oh we upgraded SCM API" perspective, the 1.x -> 2.x transition for SCM API does not mandate an upgrade to the Git plugin (but we gain improvements from the update)

SCM API is designed to be as minimal as possible.

In the longer term, we're probably going to pull all the SCM stuff out of core and into SCM API where we can provide plugin authors with better guidance and try to consolidate and tidy up a lot of the mess that has evolved in the various plugins (everyone implementing their own thread pools for polling; lack of consolidated eventing; etc)

I would really like to know what your objection is to SCM API

stephenc:
> For example credentials-plugin - nobody knows how to use it.

Next on my todo list are implementation, consumer and user guides for credentials plugin.

KostyaSha:
> stuff out of core

Like
https://issues.jenkins-ci.org/browse/JENKINS-31153?focusedCommentId=246004&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-246004


> (everyone implementing their own thread pools for polling; lack of consolidated eventing; etc)

GH plugin implementing eventing and it works fine it doesn't require to hack multiple plugins. I have good experience with github-plugin API + github-integration/pullrequest-plugins. They work great without any additional plugins, dependency tree is/was clear. 
Common APIs doesn't work well, they always limited and locked with long reviews/releases (fix for workflow trigger mess for example).

Keeping in mind that i want switch one day from github-api library i'm not so strong on polluting this plugin ;)

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -75,6 +77,13 @@ public void onPost() {
      * Called when a POST is made.
      */
     public void onPost(String triggeredByUser) {
+        onPost(SCMEvent.originOf(Stapler.getCurrentRequest()), triggeredByUser);
+    }
+
+    /**
+     * Called when a POST is made.
+     */
+    public void onPost(final String origin, String triggeredByUser) {
```
KostyaSha: S maybe replace it with more common object with fields that will be expandable in future?

src/main/java/com/cloudbees/jenkins/GitHubTrigger.java
KostyaSha: i think @jglick mean compiler warnings, but i guess they work only on annotations.

stephenc:
> So just to show useless information in UI you are adding useless hard dependencies, deprecating and hacking everything around?

We have customers with phantom builds that they don't know where they are being triggered from. I have seen non-customer users with the same problem.

So I disagree that the information is useless. It is only useless until you have builds being triggered causing unnecessary deployment int production... when you have that situation it is essential.

I am bumping a dependency that this plugin should really be depending on. Triggering scm builds will need to be consolidated into the scm event path for this plugin anyway.

KostyaSha:
Origin is only one of multiple things that may be needed for debugging, `(everyone implementing their own thread pools for polling; lack of consolidated eventing; etc)` exactly this appeared because nobody do design before implementation.  Hack, merge, release, keep 10y ... make support business... Single string looks the same. If you would need one more you will create GitHubTrigger3, GitHubTrigger4??

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -75,6 +77,13 @@ public void onPost() {
      * Called when a POST is made.
      */
     public void onPost(String triggeredByUser) {
+        onPost(SCMEvent.originOf(Stapler.getCurrentRequest()), triggeredByUser);
```
KostyaSha: So why not call `Stapler.getCurrentRequest()` in `GitHubTrigger` implementation, it still the same thread?

stephenc:
> From all what i see (threading) you can call Stapler.getCurrentRequest() in eventsubscriber/trigger impl

That assumes that all callers of the api are doing the call from a stapler request.

If somebody is relaying requests between masters, there may be no request but the origin has been relayed (and relay appended)

Hence better to have the origin as an explicit parameter

stephenc:
> System logger is standart pattern for debugging issues in jenkins.

And that is really crappy when trying to trace why a job got built. The logs have rotated away before you can find out.

Jobs already retain the cause of their build. They should include tracing information for that cause.

KostyaSha:
> If somebody is relaying requests between masters, there may be no request but the origin has been relayed (and relay appended)

So it only CB enterprise issue?

> Hence better to have the origin as an explicit parameter

So wrap it to object along with `triggeredByUser` it will be possible extend it in future.

KostyaSha:
> System logger is standart pattern for debugging issues in jenkins.
> And that is really crappy when trying to trace why a job got built. The logs have rotated away before you can find out.
> 
> Jobs already retain the cause of their build. They should include tracing information for that cause.

Crap is system settings that requires restart master, you and other still adding them into jenkins and plugins and ignoring my suggestions to stop doing it without ui/system settings.

https://github.com/jenkinsci/git-plugin/blob/master/src/main/java/hudson/plugins/git/GitSCM.java#L1782 wonderful example! Introduced because some people don't want to see verbose/too much information about build.

stephenc:
> Crap is system settings that requires restart master, you and other still adding them into jenkins and plugins and ignoring my suggestions to stop doing it without ui/system settings.

I don't add ones that require a restart unless there is a very good reason. In general I add as a non-final static field and I shall rejoice when the core dependency is high enough to switch to the new API

stephenc:
> Introduced because some people don't want to see verbose/too much information about build.

Don't get me started on how the git plugin is "death by a million check boxes"

The recent changes have certainly made it better, but omg it is still too complex for users to configure and trying to do too much imho... so then you need a verbose mode to debug why it's going "wrong"... shock horror!

stephenc:
> Keeping in mind that i want switch one day from github-api library i'm not so strong on polluting this plugin ;)

Well that is a different kettle of fish. AIUI we added the dependency on this plugin because you complained that we weren't reusing existing functionality.

If you really want to take it in a different direction then perhaps we should review that decision and rip out all the adapter code... though that would make me sad as then users would have double the configuration to do.

If that is really your vision for this plugin then we need to know sooner rather than later because the next thing on my list for GHBS is removing the duplicate server definitions... then fixing up credentials (which will likely require fixes here too)

KostyaSha:
> Well that is a different kettle of fish. AIUI we added the dependency on this plugin because you complained that we weren't reusing existing functionality.

Everything already reinvented! I'm not alone maintainer here. I have no any idea what is happening in CB private chats or issue trackers - what multi master you have and for what cases it not enough to call stapler methods. You PRed single argument for single case with deprecations and 0 extensibility in future - i'm reviewing, reviewing with examples.

> if you really want to take it in a different direction

For my own plugins unfortunately i forced it to do. This plugin is generic. In any case this subscription current mechanism can't work for oauth integration ala travis without rewrote in github-api library fixes (that usually seats and waiting SPOF guy). But it other story

stephenc:
> So wrap it to object along with triggeredByUser it will be possible extend it in future.

if you want me to switch to an event object, fair enough... though it would likely end up being most of `SCMEvent` anyway which would probably result in a larger refactoring

KostyaSha:
I guess github-plugin depends on git only because of sucking PushTrigger, i expected that it has no any relation to `SCMEvent` or whatever is inventing for common APIs. Will check in free time whether it possible at all to put git into optional dep. Common API layering ends in infinite hell with missing options. Here github-plugin is for github and made as much as more generic, so having it extensible is needed.

stephenc:
@lanwen care to weigh in?

src/main/java/com/cloudbees/jenkins/GitHubTriggerEvent.java
```diff
@@ -0,0 +1,125 @@
...
+                "timestamp=" + timestamp +
+                ", origin='" + origin + '\'' +
+                ", triggeredByUser='" + triggeredByUser + '\'' +
+                '}';
+    }
+
+    /**
+     * Builder for {@link GitHubTriggerEvent} instances..
+     */
+    public static class Builder {
```
KostyaSha: Why do you need to keep fields final? Simple class with `withXX()` + newEvent() static creator should enough imho.

stephenc: Because then you cannot have an equals method that compares fields and you cannot trust consumers of the event not to mutate. Immutable is best (I don't like `@DataBoundSetter` BTW)

KostyaSha: Do you expected this could effect anything? All info is optional.

lanwen:
Wow, what a big discussion :) Will try to review in next 24h

KostyaSha:
@lanwen just code enough ;)

KostyaSha:
processEvent() and everything else still on String arg

src/main/java/com/cloudbees/jenkins/GitHubTriggerEvent.java
```diff
@@ -0,0 +1,125 @@
...
+            this.triggeredByUser = triggeredByUser;
+            return this;
+        }
+
+        public GitHubTriggerEvent build() {
+            return new GitHubTriggerEvent(timestamp, origin, triggeredByUser);
+        }
+
+        @Override
+        public String toString() {
```
jglick: Gratuitous for a throwaway builder.

src/main/java/com/cloudbees/jenkins/GitHubTriggerEvent.java
```diff
@@ -0,0 +1,125 @@
...
+    public String getOrigin() {
+        return origin;
+    }
+
+    public String getTriggeredByUser() {
+        return triggeredByUser;
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
lanwen: time to try lombok?

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -57,7 +59,7 @@
  *
  * @author Kohsuke Kawaguchi
  */
-public class GitHubPushTrigger extends Trigger<Job<?, ?>> implements GitHubTrigger {
+public class GitHubPushTrigger extends Trigger<Job<?, ?>> implements GitHubTrigger2 {
```
lanwen: can we just drop this Interface implementation? Do you know somebody who use it?

src/test/java/org/jenkinsci/plugins/github/webhook/subscriber/DefaultPushGHEventListenerTest.java
```diff
@@ -50,10 +52,16 @@ public void shouldParsePushPayload() throws Exception {
...
         prj.setScm(GIT_SCM_FROM_RESOURCE);
 
-        new DefaultPushGHEventSubscriber()
-                .onEvent(GHEvent.PUSH, classpath("payloads/push.json"));
+        GHSubscriberEvent subscriberEvent =
+                new GHSubscriberEvent("shouldParsePushPayload", GHEvent.PUSH, classpath("payloads/push.json"));
+        new DefaultPushGHEventSubscriber().onEvent(subscriberEvent);
 
-        verify(trigger).onPost(TRIGGERED_BY_USER_FROM_RESOURCE);
+        verify(trigger).onPost(GitHubTriggerEvent.create()
```
lanwen: is it should be `verify(trigger).onPost(same(...))` (with matcher?)

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
KostyaSha: You can't know who is using it.

src/main/java/com/cloudbees/jenkins/GitHubTriggerEvent.java
jglick: IMO way more magic danger than it is worth.

lanwen: ok, should be tested separately

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
lanwen: so we can just continue to implement old interface (marked as deprecated) and do nothing with its methods and just use brand new method

KostyaSha: Or switch to java8?

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -87,6 +102,9 @@ private boolean runPolling() {
                         PrintStream logger = listener.getLogger();
                         long start = System.currentTimeMillis();
                         logger.println("Started on " + DateFormat.getDateTimeInstance().format(new Date()));
+                        if (event.getOrigin() != null) {
+                            logger.format("Started by event from %s on %tc%n", event.getOrigin(), event.getTimestamp());
```
KostyaSha: if you made toString, then simply print event

src/main/java/com/cloudbees/jenkins/GitHubTriggerEvent.java
KostyaSha: +

src/main/java/org/jenkinsci/plugins/github/extension/GHSubscriberEvent.java
```diff
@@ -0,0 +1,39 @@
...
+import javax.servlet.http.HttpServletRequest;
+import jenkins.scm.api.SCMEvent;
+import org.kohsuke.github.GHEvent;
+
+/**
+ * An event for a {@link GHEventsSubscriber}.
+ *
+ * @since 1.25.2
+ */
+public class GHSubscriberEvent extends SCMEvent<String> {
```
KostyaSha: So you are adding hard dependency on plugin that community never discussed to design?

KostyaSha:
> we're probably going to pull all the SCM stuff out of core and into SCM API

Where it and it's design was discussed with community/devs? How that happened that your plugin become hard dependency for everybody and everybody should depend on it?

KostyaSha:
Opening release sources https://github.com/jenkinsci/scm-api-plugin/tree/d03e1d08c8cc800e3edc361b3b313bf69ece67ef no docs, how at all it happenned to be dependency for everybody...

stephenc:
@KostyaSha 
> Where it and it's design was discussed with community/devs? 

Well the current long term goals is to move most stuff out of core and into separate plugins.

While we could pull the existing (crappy) SCM classes into their own plugin, to my mind it makes more sense to move them into SCM API.

When/if it comes time to move them, rest assured that - if I am involved - there will be a discussion with the community.

> https://github.com/jenkinsci/scm-api-plugin/graphs/contributors How that happened that your plugin become hard dependency for everybody and everybody should depend on it?

Well most people see the multi-branch stuff as the way forward, SCM API is the API that these things are built on top of. I have tried to keep separation of concerns (with the added cognitive costs involved) such that the multibranch stuff is all in Branch API while the pure SCM related stuff is in SCM API.

I have provided [guides for people to develop competing alternatives](https://github.com/jenkinsci/scm-api-plugin/blob/master/docs/consumer.adoc) to Branch API. Just because I see Branch API as the best approach does not mean I am arrogant and believe it is the "one true solution".

Ideally the `SCMSource`s should actually be part of the core SCM plugins, so for example Subversion and Git plugins have `SCMSource`s built in. TBH this plugin should really be the plugin with the `SCMSource` for GitHub, but that is not the case and so we have a separate plugin just to add the `SCMSource` implementations - which complicates things for users.

How many people are complaining about the efforts I am putting into the SCM API and Branch APIs? From what I can see (other than bugs that need fixing) almost everyone likes the SCM API and Branch API approach - but perhaps people are afraid to speak up (which saddens me).

I really do not want to go down the hard-fork path that you seem to want to push everyone towards. I hope this is just another case of "Lost in Translation"

stephenc:
I think the test failure is unrelated:

```
Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "Executing shouldTurnOnOverridingWhenThereIsCredentials(com.cloudbees.jenkins.GlobalConfigSubmitTest)"
```

KostyaSha:
> How many people are complaining about the efforts I am putting into the SCM API and Branch APIs? From what I can see (other than bugs that need fixing) almost everyone likes the SCM API and Branch API approach - but perhaps people are afraid to speak up (which saddens me).
> 
> I really do not want to go down the hard-fork path that you seem to want to push everyone towards. I hope this is just another case of "Lost in Translation"

No, you didn't get my point. Direction is good, but it done in the same way as started contributing in jenkins. I can show example, you blocked CloudProvisioningRequest enhancement PR and most of improvements for cloud in core, then you say that you want cloud-api-plugin, ok. People ready to participate (i.e. me), but nothing can't happen without devs from company X. I definitely sure that acegi will be replaced with spring-security only when it will be needed for company X needs. 

I'm also J developer (unfortunately) and reading development mail list. I don't remember any related discussion, i see only that somebody came from company X and says "it's future, use it". Docs? Later! Just merge and use plugin XX. Where is communication with community (if it still exists)?
I wrote new api for GH triggering and was ready to merge this stuff into github-plugin to make great plugin (but was forced to run away from jenkinsci because jenkins and jenkins infra are locked on single vendor), was ready to make multibranch, but... ? No docs, wait for docs. Ok! Now there is github-multibranch that makes 2 gh trigger useless, and full of bad practises of 1 old plugin. There is no any guarantee that you wouldn't decide to merge branch-source into scm/branch plugins or make something that will suddenly deprecate or stop working from upstream dependency.
When scm-api appeared in dependency it had only few simple classes and now it monster (that reminds me one more plugin 😃 ) . 

> almost everyone likes the SCM API and Branch API approach

Most of people even doesn't know that such plugins exists. Like != code and use it. Documentation appeared only in december afair. I got description from @jglick not so long ago. I don't talk even about dependency mess. One of solution that i saw from people - remove workflow plugins from production instance. But it other story.

> really do not want to go down the hard-fork path that you seem to want to push everyone towards.

I push to do better communication/design BEFORE doing something and not AFTER ("deal with it"). Mine rewritten (in some sense forked by functionality) plugins doesn't affect any other plugins, so i have no need to disturb anybody (excluding libraries like github-api). org-folder, branch/scm/etc affects everybody, but not discussed with everybody. You can get release/fix for any plugin you want, community - can't.

KostyaSha:
> I think the test failure is unrelated:

That's java 😄  If it pass locally, then ok.

stephenc:
> No, you didn't get my point. Direction is good, but it done in the same way as started contributing in jenkins. I can show example, you blocked CloudProvisioningRequest enhancement PR and most of improvements for cloud in core, then you say that you want cloud-api-plugin, ok. People ready to participate (i.e. me), but nothing can't happen without devs from company X. 

Rest assured I am very frustrated that I have not been given time to fix the Cloud API problems

> I definitely sure that acegi will be replaced with spring-security only when it will be needed for company X needs.

Actually I have a plan to bypass the need to do that... and I have agreement to let me do some quick experiments, but right now I am trying to tidy up some other bigger messes

KostyaSha:
> Actually I have a plan to bypass the need to do that... and I have agreement to let me do some quick experiments, but right now I am trying to tidy up some other bigger messes

And then it ends with security-plugin? :rofl: 
Replace SecurityFilters.groovy initialisation in Realm with spring-security classes chains + write hudson security classes based on spring-security? Right now frustrating and f... with security chains (hope i didn't end with fork).

stephenc:
> Most of people even doesn't know that such plugins exists. Like != code and use it. Documentation appeared only in december afair. 

Yes, I agree that the documentation should have been written sooner, but there is documentation now and I even documented a clear clean path for implementing the old-style SCM API in a modern way. 

> I got description from @jglick not so long ago. 

> I don't talk even about dependency mess. One of solution that i saw from people - remove workflow plugins from production instance. But it other story.

What is the dependency mess in SCM API? SCM API specifically has [no additional runtime dependencies](https://github.com/jenkinsci/scm-api-plugin/blob/7e3ccd67e44ea1492183c0ce9b8adeafd67f982c/pom.xml#L86-L105)

Now if you are talking about the maze of pipeline dependencies, I hear you, but how is that relevant to this plugin?

KostyaSha:
> Now if you are talking about the maze of pipeline dependencies, I hear you, but how is that relevant to this plugin?

I don't remember details but some scm related stuff was moved from workflow plugins into git and then everything bomb. Maybe not related directly to scm, but related to use/do modern apis.

src/main/java/com/cloudbees/jenkins/GitHubTriggerEvent.java
KostyaSha: I usually placing reflection tostring builder because people always forget to add new fields.

stephenc:
@lanwen are we good to go?

lanwen:
released as `1.26.0`

KostyaSha:
@lanwen please squash/fixup commits in one without mentioning user next time. Now i have useless highlights from commits.

PR #167, opened 2017-03-11
Use display url api to generate run backlink

i386:
When Blue Ocean is installed we can generate the backlink to its web UI instead of Jenkins classic. Same code has been adopted by mailer, github branch source and other plugins that generate links to the UI.

@reviewbybees 

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/167)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -132,6 +132,12 @@
             <version>1.11</version>
         </dependency>
 
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>display-url-api</artifactId>
+            <version>1.0</version>
+        </dependency>
+
         <dependency><!-- exists in the core -->
             <groupId>org.jenkins-ci.modules</groupId>
             <artifactId>instance-identity</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -135,7 +135,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>display-url-api</artifactId>
-            <version>1.0</version>
+            <version>2.0</version>
         </dependency>
 
         <dependency><!-- exists in the core -->
--- a/src/main/java/org/jenkinsci/plugins/github/status/sources/BuildRefBackrefSource.java
+++ b/src/main/java/org/jenkinsci/plugins/github/status/sources/BuildRefBackrefSource.java
@@ -4,6 +4,7 @@ import hudson.Extension;
 import hudson.model.Descriptor;
 import hudson.model.Run;
 import hudson.model.TaskListener;
+import org.jenkinsci.plugins.displayurlapi.DisplayURLProvider;
 import org.jenkinsci.plugins.github.extension.status.GitHubStatusBackrefSource;
 import org.kohsuke.stapler.DataBoundConstructor;
 
@@ -25,7 +26,7 @@ public class BuildRefBackrefSource extends GitHubStatusBackrefSource {
     @SuppressWarnings("deprecation")
     @Override
     public String get(Run<?, ?> run, TaskListener listener) {
-        return run.getAbsoluteUrl();
+        return DisplayURLProvider.get().getRunURL(run);
     }
 
     @Extension
--- a/src/test/java/org/jenkinsci/plugins/github/status/sources/BuildRefBackrefSourceTest.java
+++ b/src/test/java/org/jenkinsci/plugins/github/status/sources/BuildRefBackrefSourceTest.java
@@ -3,6 +3,7 @@ package org.jenkinsci.plugins.github.status.sources;
 import hudson.model.FreeStyleProject;
 import hudson.model.Run;
 import hudson.model.TaskListener;
+import org.jenkinsci.plugins.displayurlapi.DisplayURLProvider;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.runner.RunWith;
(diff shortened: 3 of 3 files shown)


arashcuzi:
Please approve this PR, we are using the GitHub plugin at work with Blue Ocean UI and would love to have the Blue build link show up in our required checks rather than the vanilla UI.

KostyaSha:
> Please approve this PR, we are using the GitHub plugin at work with Blue Ocean UI and would love to have the Blue build link show up in our required checks rather than the vanilla UI.

My questions about this APIs in jenkinsci/jenkins repo was ignored, now the same stuff is pulled via plugins. Questions didn't change.

oleg-nenashev:
@KostyaSha could you please add a link to your questions at least.

Also @reviewbybees and 😿 to @i386 for not using it in hist PRs

i386:
@oleg-nenashev there are other plugins that generate notifications for test failures where we would want users to be directed to Blue Ocean if that was their preferred UI. It's worth noting here that the JUnit plugin isn't just a plugin for JUnit test reporting/parsing. For reasons that are a bad accident of history, the JUnit plugin data model is used by almost every other test reporting plugin. 

Now, I concede that theoretically this coupling to JUnit is the bad kind of coupling but practically [JUnit is available on 92.44% of all Jenkins installations](http://stats.jenkins.io/plugin-installation-trend/junit.stats.json). 

To @KostyaSha's point about `display-url-api` being "more than an API plugin" - we have to provide a implementation of this plugin for classic so that, well, the software works for users. (We also have some code in there so that links won't break when people remove or disable implementations).

There is no way today to get a theoretical `classic-display-url` plugin installed on users systems except for having a plugin like `mailer` depend on it so that it should get installed (but then mailer would have an 'unnecessary' dependency... this discussion starts again). Unfortunately that ship has sailed long ago and I do not want it to derail the discussion around coupling with JUnit. If anyone is interested in fixing this particular problem, create a discussion for it on the developers mailing list with a proposal and we can take it from there (I won't be responding to questions about it here).

Now, there are a few options here to decouple from JUnit:

*Move the data model from JUnit into its own plugin `test-reporting-api` and have `display-url-api` plugin depend on it*
The work to split API and the JUnit implementation has already been done - most of the API code lives in `hudson.tasks.test` and the JUnit implementation is in `hudson.tasks.junit`. `hudson.tasks.test` would be broken out into its own plugin as `hudson.tasks.test`.

This is personally my preferred option.

Pros
* Technically easy change to make
* Would decouple a lot of other plugins from a JUnit implementation
* Display URL API only depends on models from `test-reporting-api`

Cons
* Explaining to the community why we are doing this with the ensuing 3 month battle to make the change (yay, open source!)
* Pragmatic PoV: Lots of people complaining that Github plugin doesn't work with Blue Ocean (which is coming out of beta very soon!)
* I am sure the dependency on `test-reporting-api` is one dependency too many for some purists.

*Delete org.jenkinsci.plugins.displayurlapi.DisplayURLProvider#getTestURL*

All references to the JUnit plugin are deleted from `display-url-api`.

My least favourite option, because the user experience will broken for users (they will only get links to classic for tests).

Pros
* Decoupled from JUnit plugin

Cons
* Plugins will have to be updated to remove references to this API. 
* Some systems might break while these plugins are all updated (throwing method not found and the like) which would upset users.
* Users relying on this functionality won't get Blue Ocean links (lots of tickets, complaining).

lanwen:
👎 if u want another backref, just add another class, no need to change existing

i386:
@lanwen wouldn't you then end up with two links? One that always points to classic then another that could point to Blue Ocean but also classic? What's the point of that?

lanwen:
@i386 to allow to be choosen exactly what user wants. For example i like to test BO and view it rarely, but i don't think it can be used now for everyday work. So i don't want to get the link that pointing to BO.

stephenc:
Isn't there a global config that lets the system admin select where the display-url plugin points to?

I thought the whole point of the display-url plugin was that the admin could then direct everything to their desired UI via a single switch rather than having to go and reconfigure everything in multiple places...

(Granted I have not looked at display-url or the BO provider for that, but if it doesn't provide a global config that lets you direct, then it should - and pronto - OTOH if it does provide a global config then I fail to see the objection)

lanwen:
global configuration should be used only for setup on how jenkins operates itself. Backref link is too small thing to move it to the global config

stephenc:
@lanwen as I understand it, the display-url plugin provides a global config to determine the UI the admin is configuring as the primary UI of Jenkins. All backlinks should be to the Admin-chosen primary UI. We use a global config to define the URL of Jenkins that is used in the backlinks, hence it makes sense to use a global config to define the UI of the URL of Jenkins for all external backlinks.

Anyway, this is @i386's baby I'll let him discuss it with you

lanwen:
That changes all the things. If main UI is managed by global config, this should be reflected by default backref link provider.

i386:
Yes, there is a user profile option and system prop to force things globally.

i386:
Also changing the preference won't break any links. The plugin does a redirect based on your preference.

lanwen:
So ok, approved. 

I'll merge this, but before please write some explanation for the dependency in pom (as comment) why it now requires junit to not forget to refactor it when things with dependencies go better.

Also for me **display-url-api** should contain full model and other plugins should be able to use this model as consumers or as providers. So by default - no any test urls is provided. With Junit plugin - junit links, with any other xUnit - own provider and so on.

oleg-nenashev:
Just for the record, I am not fully convinced that it worth getting the transient dependency on JUnit plugin and potentially all plugin dependencies introduced there in newer releases (e.g. if somebody does a Pipeline build step).

But it is up to @lanwen to merge if he agrees with that

lanwen:
@oleg-nenashev agree - its not the best solution, but 92% is big enough number to make changes after merge.

oleg-nenashev:
Created https://issues.jenkins-ci.org/browse/JENKINS-43538 for the issue.
Removing my review since it is up to maintainer to decide the way forward

i386:
I have removed the dependency on JUnit

i386:
ping @KostyaSha

KostyaSha:
Sorry, losing unread emails on pc after reading on mobile.

lanwen:
released as 1.27.0

PR #169, opened 2017-04-07
Add push event information as arguments to triggered jobs.

matble:
This change adds the git reference and head commit as arguments to triggered jobs.  This allows Jenkins jobs to implement specialized logic.  For example, a job can now have different behavior if the develop branch is pushed to vs. master.

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
+++ b/src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
@@ -8,9 +8,16 @@ import hudson.XmlFile;
 import hudson.console.AnnotatedLargeText;
 import hudson.model.AbstractProject;
 import hudson.model.Action;
+import hudson.model.CauseAction;
 import hudson.model.Item;
 import hudson.model.Job;
+import hudson.model.ParameterDefinition;
+import hudson.model.ParameterValue;
+import hudson.model.ParametersAction;
+import hudson.model.ParametersDefinitionProperty;
 import hudson.model.Project;
+import hudson.model.StringParameterValue;
+import hudson.model.queue.QueueTaskFuture;
 import hudson.triggers.SCMTrigger;
 import hudson.triggers.Trigger;
 import hudson.triggers.TriggerDescriptor;
@@ -44,9 +51,11 @@ import java.io.PrintStream;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.text.DateFormat;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Date;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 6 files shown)


lanwen:
Hello! Thanks for your contribution, but this plugin only triggers polling, so you have already such information in your git. You can take a look to env parameters provided by git plugin.

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -140,7 +153,23 @@ public void run() {
...
+                        if (parameterIsOverridable(pv.getName())) {
+                            it.remove();
+                        }
+                    }
+                    values.add(new StringParameterValue("ref", ref));
+                    values.add(new StringParameterValue("head", head));
+
+                    // Schedule the job with the parameters and cause
+                    QueueTaskFuture<?> build = asParameterizedJobMixIn(job)
+                            .scheduleBuild2(0, new ParametersAction(values), new CauseAction(cause));
```
lanwen: in any case, it should be EnvironmentContributingAction

matble:
That is a good point. However, the polling logic is only going to give you things latest commit and last branch pushed to, if two nearly simultaneous pushes occur then information is lost.  It is also less intuitive to have to write that logic then simply have the head and reference passed directly into the job as environment variables or arguments. Each job trigger would come from exactly the correct place without the need for polling.

That does bring up the point that getting that information does supersede the need for polling.  I didn't want to diverge from what the plugin is currently doing too much.  Perhaps the injection of the parameters should be a separate option in the plugin entirely?

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
matble: Not entirely surely what you mean, do you mean add the parameters as environment variables only?  Or add as environment variables as well as job parameters similar to what the Jenkins Github Pull Request Builder plugin does?

lanwen:
This trigger should be killed sometime :) It runs polling without any logic under the hood. So you can't change triggering behavior in an agile way.

Please look at new `scm-api`: https://github.com/jenkinsci/scm-api-plugin/blob/master/docs/implementation.adoc

GitHub should be a scm-source which provides events with complete info to something like an event bus. Then consumers subscribe to events, filter uninteresting and do the things you want.

I'm working on the implementation, but i spend only a few hours per week on this and it is in very beginning. So you can think about this kind of implementation too

matble:
I'm happy to take a look but if is a very large change I don't know how much time I would have to devote either.  In either case, do you think this is a meaningful intermediate step?  

At the very least, having the ref/head present as params/env variables after a poll allows the jobs to be more agile than they are now and removes the need for jobs have to logic to figure out last touched branch.

lanwen:
This change breaks binary compatibility and looks like a special case in general. So I don't think it could be merged in this form :(

matble:
I'm happy to add a new configuration parameter to explicitly allow this to maintain backwards compatibility, although I don't know how breaking it really is.  The plugin was triggering jobs without any arguments so the only possible argument values would be whatever was defined as "default" arguments.  

There is really no elegant solution for this today.  No solution to two pushes to different branches at same time, as well as having to use other plugins to get the same behavior.  I think a lot of people run across this issue and have to work around it or accept as-is.

KostyaSha:
> This change adds the git reference and head commit as arguments to triggered jobs. This allows Jenkins jobs to implement specialized logic. For example, a job can now have different behavior if the develop branch is pushed to vs. master.

Why not use normal trigger https://github.com/KostyaSha/github-integration-plugin/ ?

PR #171, opened 2017-05-25
backup for commit web hook

sathiya-mit:
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/171)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/github/webhook/BackupForCommitWebhook.java
+++ b/src/main/java/org/jenkinsci/plugins/github/webhook/BackupForCommitWebhook.java
@@ -1,0 +1,124 @@
+package org.jenkinsci.plugins.github.webhook;
+
+import com.cloudbees.jenkins.GitHubPushTrigger;
+import hudson.EnvVars;
+import hudson.Extension;
+import hudson.model.AbstractProject;
+import hudson.model.AperiodicWork;
+import hudson.model.AsyncAperiodicWork;
+import hudson.model.TaskListener;
+import hudson.slaves.EnvironmentVariablesNodeProperty;
+import hudson.slaves.NodeProperty;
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/github/webhook/BackupForCommitWebhook.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    @Override
+    protected void execute(TaskListener taskListener) throws IOException, InterruptedException {
+        LOGGER.info("Executing BackupForCommitWebhook");
+        if (isBackupForCommitWebhookEnabled()) {
+            List<AbstractProject> jobs = Jenkins.getInstance().getItems(AbstractProject.class);
+            for (AbstractProject job : jobs) {
+                if (isEligible(job)) {
+                    if (job.poll(taskListener).hasChanges()) {
```
KostyaSha: That not functionality for github plugin, that for `git plugin`

sathiya-mit: @KostyaSha thanks for your comment. You mean to say, this change is for git plugin and not for github plugin ?

KostyaSha: Yes, `poll` is abstractproject->gitscm functionality. Obviously you want to run periodic poll on repository that is done via gitscm (git-plugin). I will repeat in 100th time that github-plugin now do only triggering via github hook -> gitscm.poll(). So the right place for periodic run is git-plugin. But job ALREADY has 'build periodically commits" that is generic cron based configurable trigger for project.poll(). So you don't need anything to do at all beside enable checkbox.

PR #172, opened 2017-07-11
[FIXED JENKINS-45448] Add a display name to GitHub Servers configuration

stephenc:
See [JENKINS-45448](https://issues.jenkins-ci.org/browse/JENKINS-45448)

@reviewbybees

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/172)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/github/config/GitHubServerConfig.java
+++ b/src/main/java/org/jenkinsci/plugins/github/config/GitHubServerConfig.java
@@ -17,6 +17,13 @@ import hudson.security.ACL;
 import hudson.util.FormValidation;
 import hudson.util.ListBoxModel;
 import hudson.util.Secret;
+import java.io.IOException;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.util.Collections;
+import java.util.List;
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
 import jenkins.model.Jenkins;
 import jenkins.scm.api.SCMName;
 import org.apache.commons.lang3.StringUtils;
@@ -31,14 +38,6 @@ import org.kohsuke.stapler.DataBoundSetter;
 import org.kohsuke.stapler.QueryParameter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-import javax.annotation.CheckForNull;
-import javax.annotation.Nonnull;
-import java.io.IOException;
-import java.net.MalformedURLException;
-import java.net.URL;
-import java.util.Collections;
-import java.util.List;
 
 import static com.cloudbees.plugins.credentials.CredentialsMatchers.filter;
 import static com.cloudbees.plugins.credentials.CredentialsMatchers.withId;
@@ -64,7 +63,7 @@ public class GitHubServerConfig extends AbstractDescribableImpl<GitHubServerConf
     /**
      * Common prefixes that we should remove when inferring a {@link #name}.
      *
-     * @since 1.27.0
+     * @since 1.28.0
      */
     private static final String[] COMMON_PREFIX_HOSTNAMES = {
         "git.",
@@ -81,7 +80,7 @@ public class GitHubServerConfig extends AbstractDescribableImpl<GitHubServerConf
     /**
      * The name to display for the public GitHub service.
      *
-     * @since 1.27.0
+     * @since 1.28.0
      */
     private static final String GITHUB_NAME = "GitHub";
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/jenkinsci/plugins/github/config/GitHubServerConfig.java
```diff
@@ -127,6 +162,19 @@ public void setManageHooks(boolean manageHooks) {
     public void setCustomApiUrl(boolean customApiUrl) {
     }
 
+    /**
+     * Gets the optional display name of this server.
+     * @return the optional display name of this server, may be empty or {@code null} but best effort is made to ensure
+     * that it has some meaningful text.
+     */
+    public String getName() {
```
KostyaSha: `getDisplayName()` is standard pattern for jenkins where something should be shown in UI

src/main/java/org/jenkinsci/plugins/github/config/GitHubServerConfig.java
```diff
@@ -59,22 +62,45 @@
...
+     *
+     * @since 1.27.0
+     */
+    private static final String[] COMMON_PREFIX_HOSTNAMES = {
+        "git.",
+        "github.",
+        "vcs.",
+        "scm.",
+        "source."
+    };
```
KostyaSha: why it needed?

stephenc: Because if somebody provides an url like `https://github.example.com` we really would rather give that the name `example` rather than `github example` because the `github` is redundant information... on the other hand if it was `https://github1.example.com` then we want `github1 example` as the suggested display name

src/main/java/org/jenkinsci/plugins/github/config/GitHubServerConfig.java
lanwen: I think we can use both - `getDisplayName` and `getName` as `getName` looks like simple getter, but inside contains some logic (which should be avoided for getters).
With displayName we can emphasize that it gets name or some calculated default value

src/main/java/org/jenkinsci/plugins/github/config/GitHubServerConfig.java
```diff
@@ -59,22 +62,45 @@
...
      * Because of {@link GitHub} hide this const from external use we need to store it here
      */
     public static final String GITHUB_URL = "https://api.github.com";
 
     /**
+     * The name to display for the public GitHub service.
+     *
+     * @since 1.27.0
+     */
+    private static final String GITHUB_NAME = "GitHub";
```
lanwen: Maybe "Public GitHub"?

src/main/java/org/jenkinsci/plugins/github/config/GitHubServerConfig.java
stephenc: So to clarify I have added the `getDisplayName()` method that produces the formatted display name

src/main/java/org/jenkinsci/plugins/github/config/GitHubServerConfig.java
```diff
@@ -127,6 +161,34 @@ public void setManageHooks(boolean manageHooks) {
...
 
+    /**
+     * Gets the optional display name of this server.
+     *
+     * @return the optional display name of this server, may be empty or {@code null} but best effort is made to ensure
+     * that it has some meaningful text.
+     * @since 1.28.0
+     */
+    public String getName() {
+        if (StringUtils.isBlank(name)) {
```
lanwen: with new display name method it can be simple getter

src/main/java/org/jenkinsci/plugins/github/config/GitHubServerConfig.java
KostyaSha: exactly what i wanted comment

KostyaSha:
some push hide our comments

src/main/java/org/jenkinsci/plugins/github/config/GitHubServerConfig.java
```diff
@@ -184,8 +180,11 @@ public String getName() {
      */
     public String getDisplayName() {
         String n = getName();
-        String a = StringUtils.isBlank(apiUrl) || GITHUB_URL.equals(apiUrl)
-                ? "https://github.com" : StringUtils.removeEnd(apiUrl, "/api/v3");
+        boolean gitHubOrg = StringUtils.isBlank(apiUrl) || GITHUB_URL.equals(apiUrl);
+        if (StringUtils.isBlank(n)) {
+            n = gitHubOrg ? GITHUB_NAME : SCMName.fromUrl(apiUrl, COMMON_PREFIX_HOSTNAMES);
```
lanwen: please avoid one-letter vars next time. In java its very hard to follow code if it has obfuscated variables

lanwen:
what about `PUBLIC_GITHUB`?

lanwen:
hmm it's strange, but where CI build? @oleg-nenashev

KostyaSha:
@lanwen probably they disabled standard builder and enforcing everybody to use their cloudbees pipeline

lanwen:
meh... is there any info about this change?

KostyaSha:
We can add travis build ;)

lanwen:
Travis hangs on tests :( i've tried

PR #174, opened 2017-07-31
Add CI service from ci.jenkins.io

stephenc:
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/174)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,1 @@
+buildPlugin()


stephenc:
@lanwen @KostyaSha looks like some interesting failures on Windows will be exposed with this one

lanwen:
Seems `GitHubClientCacheCleanupTest` works only on nix

KostyaSha:
What i see on this ugly UI is that windows and linux has red icons, so both failed

lanwen:
will fix that later...

stephenc:
@KostyaSha the linux build was interrupted by the failure in the windows build

KostyaSha:
@stephenc stupid logic, CI should show the real result

stephenc:
@KostyaSha take it up with @rtyler or whoever wrote the shared library such that a failure on one of linux/windows triggers an abort of the other

lanwen:
It looks like linux build time outed

PR #179, opened 2017-10-29
[JENKINS-43786] More compatible adaptation

recena:
[JENKINS-43786](https://issues.jenkins-ci.org/browse/JENKINS-43786)

@KostyaSha, @ikedam, What do you think?

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/github/admin/GitHubHookRegisterProblemMonitor/message.groovy
+++ b/src/main/resources/org/jenkinsci/plugins/github/admin/GitHubHookRegisterProblemMonitor/message.groovy
@@ -2,10 +2,10 @@ package org.jenkinsci.plugins.github.admin.GitHubHookRegisterProblemMonitor
 
 def f = namespace(lib.FormTagLib)
 
-div(class: 'warning') {
+div(class: 'alert alert-warning') {
     form(method: 'post', action: "${rootURL}/${my?.url}/act", name: my?.id) {
-        text(_('hook.registering.problem'))
         f.submit(name: 'yes', value: _('view'))
         f.submit(name: 'no', value: _('dismiss'))
     }
+    text(_('hook.registering.problem'))
 }
--- a/src/main/resources/org/jenkinsci/plugins/github/admin/GitHubHookRegisterProblemMonitor/message.groovy
+++ b/src/main/resources/org/jenkinsci/plugins/github/admin/GitHubHookRegisterProblemMonitor/message.groovy
@@ -1,11 +1,24 @@
 package org.jenkinsci.plugins.github.admin.GitHubHookRegisterProblemMonitor
 
+import hudson.util.VersionNumber
+import jenkins.model.Jenkins
+
 def f = namespace(lib.FormTagLib)
 
-div(class: 'alert alert-warning') {
-    form(method: 'post', action: "${rootURL}/${my?.url}/act", name: my?.id) {
-        f.submit(name: 'yes', value: _('view'))
-        f.submit(name: 'no', value: _('dismiss'))
+if (Jenkins.getVersion().isNewerThan(new VersionNumber("2.88"))) {
+    div(class: 'alert alert-warning') {
+        form(method: 'post', action: "${rootURL}/${my?.url}/act", name: my?.id) {
+            f.submit(name: 'yes', value: _('view'))
+            f.submit(name: 'no', value: _('dismiss'))
+        }
+        text(_('hook.registering.problem'))
     }
-    text(_('hook.registering.problem'))
+} else {
+    div(class: 'warning') {
+        form(method: 'post', action: "${rootURL}/${my?.url}/act", name: my?.id) {
+            text(_('hook.registering.problem'))
+            f.submit(name: 'yes', value: _('view'))
+            f.submit(name: 'no', value: _('dismiss'))
+        }
+    }
 }
--- a/src/main/resources/org/jenkinsci/plugins/github/admin/GitHubHookRegisterProblemMonitor/message.properties
+++ b/src/main/resources/org/jenkinsci/plugins/github/admin/GitHubHookRegisterProblemMonitor/message.properties
@@ -1,3 +1,3 @@
 view=View
 dismiss=Dismiss
-hook.registering.problem=There were some problems while registering or removing one ore more GitHub webhooks. Would you like to view the problems?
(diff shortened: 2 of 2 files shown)


src/main/resources/org/jenkinsci/plugins/github/admin/GitHubHookRegisterProblemMonitor/message.groovy
```diff
@@ -1,11 +1,24 @@
...
+import hudson.util.VersionNumber
+import jenkins.model.Jenkins
+
 def f = namespace(lib.FormTagLib)
 
-div(class: 'alert alert-warning') {
-    form(method: 'post', action: "${rootURL}/${my?.url}/act", name: my?.id) {
-        f.submit(name: 'yes', value: _('view'))
-        f.submit(name: 'no', value: _('dismiss'))
+if (Jenkins.getVersion().isNewerThan(new VersionNumber("2.88"))) {
```
recena: `2.88` is only a tentative version, I'm waiting for getting this https://github.com/jenkinsci/jenkins/pull/2857 already merged

ikedam: Why do you need to wait that?
`alert` and `alert-warning` is available in Jenkins 1.633 (https://github.com/jenkinsci/ssh-slaves-plugin/pull/70#pullrequestreview-63935217)

src/main/resources/org/jenkinsci/plugins/github/admin/GitHubHookRegisterProblemMonitor/message.groovy
```diff
@@ -1,11 +1,24 @@
...
+        }
+        text(_('hook.registering.problem'))
+    }
+} else {
+    div(class: 'warning') {
+        form(method: 'post', action: "${rootURL}/${my?.url}/act", name: my?.id) {
+            text(_('hook.registering.problem'))
+            f.submit(name: 'yes', value: _('view'))
+            f.submit(name: 'no', value: _('dismiss'))
+        }
```
ikedam: Why do you want to put `text` to `form`?
(and not for Jenkins >= 2.88?)

src/main/resources/org/jenkinsci/plugins/github/admin/GitHubHookRegisterProblemMonitor/message.groovy
```diff
@@ -1,11 +1,24 @@
...
     }
-    text(_('hook.registering.problem'))
-}
+}
```
ikedam: Let's add a line break.
Though no line break is almost harmless for text editors, text files must end with line breaks for the POSIX definition.

ikedam:
@recena 
Actually I can't get why you mentioned me as I haven't participate github-plugin nor JENKINS-43786.
You might have mistaken me for someone else to invite.

recena:
@ikedam My apologies, I meant @lanwen.

src/main/resources/org/jenkinsci/plugins/github/admin/GitHubHookRegisterProblemMonitor/message.groovy
oleg-nenashev: Originally `text(_('hook.registering.problem'))` was invoked after the form clause. It should be restored if there is no specific reason to do that

recena: @oleg-nenashev No, there is an important reason for that. The DOM generated should be synced with the proposal on https://github.com/jenkinsci/jenkins/pull/2857

ikedam: This `else` block is for Jenkins <= 2.88, that is, without jenkinsci/jenkins#2857 .
Do you mean this DOM should be fixed regardless of jenkinsci/jenkins#2857 ?

src/main/resources/org/jenkinsci/plugins/github/admin/GitHubHookRegisterProblemMonitor/message.groovy
lanwen: what if it will be lts? Is side effects worse it?

recena: @lanwen I don't think so. With this approach if the plugin is running on Jenkins version `< 2.88` we see the current appearance, but if the plugin is running on Jenkins `> 2.88` we see the new design.

src/main/resources/org/jenkinsci/plugins/github/admin/GitHubHookRegisterProblemMonitor/message.groovy
recena: @ikedam Yes, https://github.com/jenkinsci/jenkins/pull/2857 provides a new re-styling and if a maintainer wants to use it, needs to perform a small change like this. But we want to keep backward compatibility in the UI.

recena: @oleg-nenashev Please, could you update your review?

ikedam: @recena 
I'm afraid you didn't get the point I asked.
I suppose this PR contains 2 changes:

* Fix for misplaced text and buttons. (this means there's a bug in the current version)
* Preparation for jenkinsci/jenkins#2857 .

You haven't mentioned the prior one, and it makes me wonder whether that change is expected one, or the one by mistake. 

Attaching screenshot might clarify the fix by this PR.

oleg-nenashev: > No, there is an important reason for that. The DOM generated should be synced with the proposal on ...

It is for the legacy branch. Why does it need to be synced?

recena: @oleg-nenashev I don't know what you are not understanding. There are two blocks, one with the original implementation and other with the new one. And yes, the order on how the elements are rendered is important.

PR #180, opened 2017-10-31
Update github-api dependency

nwillems:
This is done to accommodate the larger IDs for PRs.

I'm not sure this is actually needed - since this(https://github.com/kohsuke/github-api/issues/387) thing is happening. But just in case.

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/180)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -95,7 +95,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>github-api</artifactId>
-            <version>1.86</version>
+            <version>1.90</version>
         </dependency>
 
         <dependency>


KostyaSha:
You can update github-api-plugin in jenkins, no need to depend on latest.

joppa27:
@KostyaSha I'm not sure how plugin dependency resolution works in Jenkins, but shouldn't we merge this PR to make 1.90 the default github-api version for this plugin?  On my jenkins instances jobs fail because the status ID returned from the github API is a `Long` and not an `Int`. 

I'm a little worried that this fix is not making it to required plugins, and I'm not sure how to help with that.  Also, how does one update the `github-api-plugin` to use the 1.90 version of the github-api in Jenkins?

KostyaSha:
> Also, how does one update the github-api-plugin to use the 1.90 version of the github-api in Jenkins?

github-api-plugin 1.90 was released, so you can just update in update center. github-plugin is binary compatible with github-api-plugin versions, so if issue is in library-plugin then you can install/update it.

PR #182, opened 2017-11-21
[JENKINS-48139] Support ignoring push events from certain github users (ie a Jenkins user)

dbsanfte:
Support for ignoring push events from certain Github usernames. The obvious use case is ignoring push events from a Jenkins user and keeping them from re-triggering builds.

Tests + help files.

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/182)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/github/config/GitHubPluginConfig.java
+++ b/src/main/java/org/jenkinsci/plugins/github/config/GitHubPluginConfig.java
@@ -63,6 +63,13 @@ public class GitHubPluginConfig extends GlobalConfiguration {
     private URL hookUrl;
     private HookSecretConfig hookSecretConfig = new HookSecretConfig(null);
 
+    /**
+     * @see #getCommittersToIgnore()
+     * @see #setCommittersToIgnore(String)
+     */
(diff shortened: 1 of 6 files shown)


KostyaSha:
And that wouldn't work like you expect, triggering decision should be in trigger and not in incoming hooks. Either you will get in any case build for this user from trigger run.

lanwen:
This is not a global story. All decisions should be delegated to job level

lanwen:
This is not a global story. All decisions should be delegated to job level

dbsanfte:
Happy to make changes, but for a feature that should have been in the plugin from the get-go, beggars shouldn't be so choosy. 

Is it possible to delegate this to the trigger without there being a "failed build" in Jenkins? That's what the [ci skip] plugins result in. Events from banned users shouldn't be recognized at all.

KostyaSha:
Triggers has possibility to re-run checks, so any filtering on event processing will be ignored and you will get all builds. Filtering events is broken by design or may help only for pure triggers. But it wouldn't really help even for  github push hooks trigger existing here because git polling will anyway find something different.

PR #183, opened 2017-12-15
[JENKINS-47820] Update github-api dependency

kawazaki0:
Updating org.jenkins-ci.plugins.github-api to the newest version should resolve an issue JENKINS-47820 about parsing large IDs from github.

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/183)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -95,7 +95,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>github-api</artifactId>
-            <version>1.86</version>
+            <version>1.90</version>
         </dependency>
 
         <dependency>


KostyaSha:
duplicates https://github.com/jenkinsci/github-plugin/pull/180

lanwen:
Hm, i think it could simplify update process

KostyaSha:
> Hm, i think it could simplify update process

Update of what? You are adding unneeded upper dependencies. When you are updating jenkins it's updates all plugins, big is in github-api-plugin and not in github-plugin.

lanwen:
Update of github-plugin itself. Or new installation. We already have this dependency, so why not to depend on latest stable version?

KostyaSha:
Why not bump java to 9 because it has better gc?

lanwen:
Updating to java 9 will not bring it to jenkins and other plugins automatically

oleg-nenashev:
Jenkins is not going to support Java 9 in short-term. https://issues.jenkins-ci.org/browse/JENKINS-40689

PR #185, opened 2018-01-09
Always update existing hooks to make sure all options (e.g. shared se…

silbernm:
…cret) are up-to-date

Easy option for solving https://issues.jenkins-ci.org/browse/JENKINS-48610.

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/185)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/github/webhook/WebhookManager.java
+++ b/src/main/java/org/jenkinsci/plugins/github/webhook/WebhookManager.java
@@ -32,6 +32,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Predicates.notNull;
 import static com.google.common.base.Predicates.or;
 import static java.lang.String.format;
+import static org.apache.commons.collections.CollectionUtils.isEqualCollection;
 import static org.jenkinsci.plugins.github.config.GitHubServerConfig.allowedToManageHooks;
 import static org.jenkinsci.plugins.github.extension.GHEventsSubscriber.extractEvents;
 import static org.jenkinsci.plugins.github.extension.GHEventsSubscriber.isApplicableFor;
@@ -188,6 +189,14 @@ public class WebhookManager {
 
                     Set<GHEvent> alreadyRegistered = from(hooks)
                             .transformAndConcat(eventsFromHook()).toSet();
+
+                    if (hooks.size() == 1 && isEqualCollection(alreadyRegistered, events)) {
+                        final Secret secret = GitHubPlugin.configuration().getHookSecretConfig().getHookSecret();
+                        if (secret == null) {
+                            LOGGER.debug("Hook already registered for events {}", events);
+                            return null;
+                        }
+                    }
 
                     Set<GHEvent> merged = from(alreadyRegistered).append(events).toSet();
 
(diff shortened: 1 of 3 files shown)


KostyaSha:
Looks like a hack, is there any more strict way to verify that secret changed?

silbernm:
I don't think there is a reliable way to detect secret changes. Unfortunately, GitHub returns the value "********" for the secret of an existing hook.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/jenkinsci/github-plugin/185#-:-L2RXbRrDVfxt6eLWsV5:b-1cm0hb)*

KostyaSha:
So if secret exists and configured we should update it?

silbernm:
True, we don't have to update the hook if no secret is configured in the github plugin. In that case, we can also leave an already configured secret in the hook, since the github plugin just discards the X-Hub-Signature.

As described in the the accompanying bug https://issues.jenkins-ci.org/browse/JENKINS-48610, we could also just update the secret value of the hook. This however would require quite some changes to the GitHub API plugin, and still result in the same number of HTTP requests.

I was also thinking of just updating a hook if we receive a hook payload with an invalid X-Hub-Signature. However, an attacker could then trick us into flooding GitHub with hook update requests.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/jenkinsci/github-plugin/185#-:-L2R_z50UzpigHyr8Cam:b-eb32ba)*

silbernm:
To answer your questions more directly: Yes, if the secret is configured in the GitHub plugin, we must update the hook, because we can't check whether the secret in the hook is correct.

Since we read the secret from the Jenkins credentials, we can't know when the secret has been changed for the last time. For example, a user could modify the credential files on the disk without us knowing. And even if we were able to detect both the last change of the secret in Jenkins and in GitHub, I think it would be too fragile to just assume that the secret is correct if the change in GitHub was later than the change in Jenkins.

A completely different solution that comes to my mind is the following: We could add an optional parameter to the GitHubPluginConfig.doReRegister method and a "Force" checkbox right next to the "Re-register hooks for all jobs" button in the GitHub plugin configuration. Then we could keep the current behavior by default, but let the user regenerate all hooks unconditionally if desired.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/jenkinsci/github-plugin/185#-:-L2RoAVfZmEaixtZAFSP:b-6ma96g)*

lanwen:
will it really replace the hook? Or just add new?

silbernm:
@lanwen: I have not touched the code that updates an existing hook. But yes, I have tested the patched plugin, and it replaces any existing hooks.

@KostyaSha: I have pushed changes such that the hooks are only replaced unconditionally if the secret is not configured in the GitHub plugin.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/jenkinsci/github-plugin/185#-:-L2S-WvTbEC15kKXbYrB:bfpibsr)*

ljupchokotev:
Is there anything blocking this from being merged? It is really useful if you have a lot of repositories and you need to update the shared secret for the webhook. Currently, it means going and manually changing it in each repository webhook configuration or deleting the webhook from all repositories and let Jenkins re-register it. It is a lot of manual work either way.

KostyaSha:
> However, an attacker could then trick us into flooding GitHub with hook update requests.

Probably he can do it with re-registration also. As soon as job will be reconfigured jenkins will attempt to set token...
We can add global option with default new behavior and delete it after few releases.

silbernm:
@KostyaSha: Reconfiguring a job requires the attacker to be logged into Jenkins. The attack with the invalid X-Hub-Signature could be performed without login. Therefore, I would still prefer the PR as implemented right now: If a job has a secret, then its webhook gets re-registered in GitHub even if it already exists.

This PR is now over a year old - is anybody still interested in this contribution? If so, I am happy to tackle any specific remarks. If not, I'm going to close it end of February.

ljupchokotev:
I am still interested in this. I believe this PR makes more sense than the current logic for the webhooks.

PR #187, opened 2018-01-10
Update Parent POM to 2.37 and use Apache HttpComponents API 4.x plugin 

oleg-nenashev:
I was running PCT against the plugin and hit some issues with that.

- [x] - Update Parent POM and resolve upper bound dependency issues
- [x] - Switch to https://github.com/jenkinsci/apache-httpcomponents-client-4-api-plugin so that GitHubWebHookFullTest does not blow up with `ON: NoClassDefFoundError: Could not initialize class com.jayway.restassured.RestAssured` due to the missing httpmime dependency

@reviewbybees @lanwen @KostyaSha @raul-arabaolaza

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/187)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -154,9 +154,9 @@
         </dependency>
 
         <dependency>
-            <groupId>org.apache.httpcomponents</groupId>
-            <artifactId>httpclient</artifactId>
-            <version>4.5.2</version>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>apache-httpcomponents-client-4-api</artifactId>
+            <version>4.5.3-2.0</version>
             <scope>test</scope>
         </dependency>
 
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>2.6</version>
+        <version>2.37</version>
         <relativePath />
     </parent>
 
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
Merged with master, let's what CI says now

pom.xml
```diff
@@ -48,13 +48,12 @@
 
     <properties>
         <jenkins.version>1.625.3</jenkins.version>
-        <jenkins-test-harness.version>1.625.3</jenkins-test-harness.version>
```
KostyaSha: what minimal version will be now?

oleg-nenashev: Does it matter? it is Jenkins Test Harness (test tool), not Jenkins Core

lanwen: It brings some dependencies too

oleg-nenashev: only on the test scope. The version would become 2.31. For testing of the plugin with new Jenkins core in Plugin Compat Tester it is critical to have 2.23 at least so that it supports cores without war-for-test

KostyaSha:
Hm.. github-api uses okhttp, what is using httpclient? mock?

dwnusbaum:
@KostyaSha Yes it looks like [wiremock:1.57 uses httpclient:4.5](https://github.com/tomakehurst/wiremock/blob/6c178d7b781cca7b80851510459d53ad2bbcd17f/build.gradle#L30).

oleg-nenashev:
@reviewbybees done

oleg-nenashev:
Actually I see one issue there. lemme check it before proceeding

oleg-nenashev:
Yes, due to to whatever reason httpcore is bundled directly... Different versions are bundled in this version and in the master one, so it is not a regression in this pull request

```
  inflating: WEB-INF/licenses.xml    
  inflating: WEB-INF/lib/slf4j-jdk14-1.7.25.jar  
  inflating: WEB-INF/lib/httpclient-4.5.3.jar  
  inflating: WEB-INF/lib/github.jar  
  inflating: WEB-INF/lib/okio-1.6.0.jar  
  inflating: WEB-INF/lib/jackson-databind-2.6.0.jar  
  inflating: WEB-INF/lib/jackson-core-2.6.0.jar  
  inflating: WEB-INF/lib/okhttp-2.7.5.jar  
  inflating: WEB-INF/lib/okhttp-urlconnection-2.7.5.jar  
  inflating: WEB-INF/lib/groovy-sandbox-1.10.jar  
  inflating: WEB-INF/lib/jackson-annotations-2.6.0.jar  
  inflating: WEB-INF/lib/httpcore-4.4.6.jar  
  inflating: WEB-INF/lib/commons-lang3-3.4.jar 
```

mvn dependency:tree does not help much, likely a packaging issue

timja:
this was fixed in https://github.com/jenkinsci/github-plugin/pull/209 and can be closed

PR #188, opened 2018-01-12
[JENKINS-48012] Require a X-Hub-Signature header when receiving a hook payload and if…

silbernm:
… a secret is configured

As reported in JENKINS-48012, an attacker can currently circumvent the signature check of the GitHub plugin simply by omitting the X-Hub-Signature header.

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/188)

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/github/test/HookSecretHelper.java
+++ b/src/test/java/org/jenkinsci/plugins/github/test/HookSecretHelper.java
@@ -62,4 +62,20 @@ public class HookSecretHelper {
     public static void storeSecret(final String secretText) {
         storeSecretIn(Jenkins.getInstance().getDescriptorByType(GitHubPluginConfig.class), secretText);
     }
+
+    /**
+     * Unsets the current hook secret.
+     *
+     * @param config where to remove
+     */
+    public static void removeSecretIn(GitHubPluginConfig config) {
+        config.getHookSecretConfig().setCredentialsId(null);
+    }
+
+    /**
+     * Unsets the current hook secret.
+     */
+    public static void removeSecret() {
+        removeSecretIn(Jenkins.getInstance().getDescriptorByType(GitHubPluginConfig.class));
+    }
 }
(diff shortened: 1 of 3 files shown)


src/test/java/org/jenkinsci/plugins/github/test/HookSecretHelper.java
```diff
@@ -62,4 +62,20 @@ public void run() {
     public static void storeSecret(final String secretText) {
         storeSecretIn(Jenkins.getInstance().getDescriptorByType(GitHubPluginConfig.class), secretText);
     }
+
+    /**
+     * Unsets the current hook secret.
+     *
+     * @param config where to remove
+     */
+    public static void removeSecretIn(GitHubPluginConfig config) {
```
daniel-beck: `@Restricted` perhaps?

oleg-nenashev: It's in the Test scope

daniel-beck:
Seems to be introduced by #134:

> also don't bother signature validation if no header from github with signature

Versions from v1.21.0 appear to be affected.

daniel-beck:
@oleg-nenashev 

> Currently it says "Checks that an incoming request has a valid signature, if there is specified a signature in the config." which is exactly what the method does.

It only checks the signature validity if it's present, but if there's no signature (which is necessary) it won't fail. Or do I misunderstand this?

silbernm:
@daniel-beck 
@oleg-nenashev 
I think we are on the same page, but the documentation is misleading. There is never a "signature" in the config, but a "hook secret". I understood the Javadoc comment as "if there is specified a secret in the config" (because that is the only thing that makes sense IMHO) and coded my fix according to that interpretation. I have tried coming up with a better wording and I have moved the line that reads the signature header further down, such that it is clear that the header is validated if and only if a hook secret is specified in the GitHub plugin config.

Regarding GitHub Enterprise on-premise: That is exactly the setup that we are using at my workplace. When not configuring a shared secret in the GitHub plugin configuration, the fixed method will continue *not* to verify the signature, even if one was provided. So we should be fine on that use-case.

daniel-beck:
> `[ERROR] src/main/java/org/jenkinsci/plugins/github/webhook/RequirePostWithGHHookPayload.java[135] (sizes) LineLength: Line is longer than 120 characters (found 124).`

PR #197, opened 2018-08-28
[JENKINS-53149] Follow same pattern when setting contextSource

mdelapenya:
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/197)


mdelapenya:
Closing, as I think the problem is elsewhere: https://github.com/jenkinsci/github-plugin/blob/master/src/main/java/org/jenkinsci/plugins/github/util/BuildDataHelper.java#L32

The build could contain more than one workspaces: one for the shared library, and one for the real project to build.

Right now, the build data obtained is the first one, which is the shared library.

Working on a fix :)

PR #198, opened 2018-08-30
JENKINS-53149 Fix build data calculation when there are more than one build actions

mdelapenya:


mdelapenya:
Hey @lanwen, I've tested this locally and seems to fix the problem with stage statuses and shared libraries.

Could you please take a look at this PR?

Thanks in advance

lanwen:
@KostyaSha

mdelapenya:
BTW, I was able to fix the broken tests adding a fallback to default implementation  when there is only one build data, hope it helps in the review process

src/main/java/org/jenkinsci/plugins/github/util/BuildDataHelper.java
```diff
@@ -19,6 +22,48 @@
...
+     * the proper build data. This way, the SHA returned in the build data will
+     * relate to the project
+     *
+     * @param parentName name of the parent build
+     * @param parentFullName full name of the parent build
+     * @param buildDataList the list of build datas from a build run
+     * @return the build data related to the project, null if not found
+     */
+    public static BuildData calculateBuildData(
+        String parentName, String parentFullName, List<BuildData> buildDataList) {
```
lanwen: codestyle - move `);` to next line plz

src/main/java/org/jenkinsci/plugins/github/util/BuildDataHelper.java
```diff
@@ -29,7 +74,12 @@ private BuildDataHelper() {
      */
     @Nonnull
     public static ObjectId getCommitSHA1(@Nonnull Run<?, ?> build) throws IOException {
-        BuildData buildData = build.getAction(BuildData.class);
+        List<BuildData> buildDataList = build.getActions(BuildData.class);
+
+        Job<?, ?> parent = build.getParent();
+
+        BuildData buildData = calculateBuildData(
+            parent.getName(), parent.getFullName(), buildDataList);
```
lanwen: codestyle - move `);` to next line plz

src/main/java/org/jenkinsci/plugins/github/util/BuildDataHelper.java
```diff
@@ -19,6 +22,48 @@
...
+            return buildDataList.get(0);
+        }
+
+        String projectName = parentFullName.replace(parentName, "");
+
+        if (projectName.endsWith("/")) {
+            projectName = projectName.substring(0, projectName.lastIndexOf('/'));
+        }
+
+        if (buildDataList == null) {
```
lanwen: plz move it on top

src/main/java/org/jenkinsci/plugins/github/util/BuildDataHelper.java
```diff
@@ -19,6 +22,48 @@
...
+     *
+     * @param parentName name of the parent build
+     * @param parentFullName full name of the parent build
+     * @param buildDataList the list of build datas from a build run
+     * @return the build data related to the project, null if not found
+     */
+    public static BuildData calculateBuildData(
+        String parentName, String parentFullName, List<BuildData> buildDataList) {
+
+        if (buildDataList != null && buildDataList.size() == 1) {
```
lanwen: with null precheck this null check is redundant

mdelapenya:
Thanks for your feedback @lanwen, I added two commits with the requested changes.

The most difficult part was to understand how the build data was collected, so at the end of the day I had to debug (blessed debugger!)

Cheers!

KostyaSha:
probably good, though line width could be 120

KostyaSha:
> The most difficult part was to understand how the build data was collected, so at the end of the day I had to debug (blessed debugger!)

BuildData is a design mistake. It designed like a job storage for all builds, but it attached to every build and instead of getting data from job it iterates on 'latestBuild' that fails for async/parallel builds. Any tricks around BuildData will be broken by design unfortunately.

mdelapenya:
@KostyaSha I can change the line ending to 120 if you prefer 👍

lanwen:
released as 1.29.3
sorry for the delay

mdelapenya:
No worries at all! I understand it could take time to do it

Thanks for your help!

PR #205, opened 2019-02-01
JENKINS-55920: add CasC support to GitHubPluginConfig.

dgarzon:
This PR addresses the issue raised in the `configuration-as-code-plugin`: https://github.com/jenkinsci/configuration-as-code-plugin/issues/672.

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/205)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/github/config/GitHubPluginConfig.java
+++ b/src/main/java/org/jenkinsci/plugins/github/config/GitHubPluginConfig.java
@@ -20,6 +20,8 @@ import org.jenkinsci.plugins.github.migration.Migrator;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.DoNotUse;
 import org.kohsuke.github.GitHub;
+import org.kohsuke.stapler.DataBoundConstructor;
+import org.kohsuke.stapler.DataBoundSetter;
 import org.kohsuke.stapler.QueryParameter;
 import org.kohsuke.stapler.StaplerRequest;
 import org.kohsuke.stapler.interceptor.RequirePOST;
@@ -76,6 +78,7 @@ public class GitHubPluginConfig extends GlobalConfiguration {
     @SuppressWarnings("unused")
     private transient InstanceIdentity identity;
 
+    @DataBoundConstructor
     public GitHubPluginConfig() {
         load();
     }
@@ -85,6 +88,7 @@ public class GitHubPluginConfig extends GlobalConfiguration {
     }
 
     @SuppressWarnings("unused")
+    @DataBoundSetter
     public void setConfigs(List<GitHubServerConfig> configs) {
         this.configs = configs;
     }
@@ -97,14 +101,16 @@ public class GitHubPluginConfig extends GlobalConfiguration {
         return from(getConfigs()).filter(allowedToManageHooks()).first().isPresent();
     }
 
-    public void setHookUrl(URL hookUrl) {
+    @DataBoundSetter
+    public void setHookUrl(String hookUrl) {
         if (overrideHookUrl) {
-            this.hookUrl = hookUrl;
+            this.hookUrl = parseHookUrl(hookUrl);
         } else {
             this.hookUrl = null;
         }
     }
 
+    @DataBoundSetter
     public void setOverrideHookUrl(boolean overrideHookUrl) {
         this.overrideHookUrl = overrideHookUrl;
     }
@@ -258,7 +264,16 @@ public class GitHubPluginConfig extends GlobalConfiguration {
         return hookSecretConfig;
     }
 
+    @DataBoundSetter
     public void setHookSecretConfig(HookSecretConfig hookSecretConfig) {
         this.hookSecretConfig = hookSecretConfig;
+    }
+
+    private URL parseHookUrl(String hookUrl) {
+        try {
+            return new URL(hookUrl);
+        } catch (MalformedURLException e) {
+            return null;
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


oleg-nenashev:
Would be nice to have a JIRA ticket for it so that we add the case to the
compatibility dashboard

dgarzon:
@oleg-nenashev here you go: https://issues.jenkins-ci.org/browse/JENKINS-55920

oleg-nenashev:
Thank you!

src/main/java/org/jenkinsci/plugins/github/config/GitHubPluginConfig.java
```diff
@@ -97,14 +101,16 @@ public boolean isManageHooks() {
         return from(getConfigs()).filter(allowedToManageHooks()).first().isPresent();
     }
 
-    public void setHookUrl(URL hookUrl) {
+    @DataBoundSetter
+    public void setHookUrl(String hookUrl) {
```
oleg-nenashev: It breaks binary compatibility. Please also keep the original method

src/main/java/org/jenkinsci/plugins/github/config/GitHubPluginConfig.java
```diff
@@ -258,7 +264,16 @@ public HookSecretConfig getHookSecretConfig() {
...
     public void setHookSecretConfig(HookSecretConfig hookSecretConfig) {
         this.hookSecretConfig = hookSecretConfig;
     }
+
+    private URL parseHookUrl(String hookUrl) {
+        try {
+            return new URL(hookUrl);
+        } catch (MalformedURLException e) {
+            return null;
+        }
```
oleg-nenashev: Add system logging at least?

dgarzon: 👍

dgarzon:
@oleg-nenashev I tested the hpi and keeping the old setter does not work, as it is picked up by 
CasC even though it is not annotated. Any ideas why?

dgarzon:
@oleg-nenashev actually, it does work. The only thing we needed was annotating `setOverrideHookUrl` with `@DataBoundSetter` and exposing a getter. Now CasC works as expected 👍

src/main/java/org/jenkinsci/plugins/github/config/GitHubPluginConfig.java
```diff
@@ -76,6 +78,7 @@
     @SuppressWarnings("unused")
     private transient InstanceIdentity identity;
 
+    @DataBoundConstructor
     public GitHubPluginConfig() {
         load();
```
KostyaSha: DataBoundConstructor is used for binding new data, `load()` is loading old data. Thats looks wrong

dgarzon: This is the constructor being used by CasC, and it uses the `getter` and `setter` to mutate the instance. The other constructor would make `List<GitHubServerConfig> configs` required, which should not. What do you suggest would be an appropriate fix to address your concerns?

KostyaSha: @jglick could you describe why DBC is set on top of Bean constructor that loads data?

jglick: This is wrong; the annotation should be deleted. `GitHubPluginConfig` is a singleton (`GlobalConfiguration`) so it may have `@DataBoundSetter`s but that is it.

jglick:
BTW I strongly recommend inclusion of a functional test in PRs like this. There are several examples to look at of plugins which have a `test` dep on `configuration-as-test` for this purpose.

er1c:
I just ran into this issue, are there still outstanding issues I could help with to get it merged?

er1c:
I have a unit test written for this, but I'm getting a bunch of unrelated errors doing  `mvn test`:

```
shouldNotAllowSSRFUsingHookUrl(org.jenkinsci.plugins.github.config.GitHubPluginConfigTest)  Time elapsed: 0.159 sec  <<< ERROR!
org.jvnet.hudson.reactor.ReactorException: java.lang.Error: java.lang.reflect.InvocationTargetException
Caused by: java.lang.Error: java.lang.reflect.InvocationTargetException
Caused by: java.lang.reflect.InvocationTargetException
Caused by: java.lang.NullPointerException
```

I made sure I'm running java8, what's the proper procedure here?

Thanks,

Eric

jglick:
I do not get such an error. Hard to guess at the cause without seeing the stack trace of the `NullPointerException`.

er1c:
It looks like these tests are already marked as "existing failures" on the master branch: https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fgithub-plugin/detail/PR-208/1/tests, so the CI is having issues too

er1c:
Here is the full (example) error I get: https://gist.github.com/er1c/9713ccac7e6b6ece4516b9dac9f9e20a . I tried on both my OSX and a linux box, same result.

I just checked out master and it does not have this error, so let me see if I can figure out what in this PR broke the build.

casz:
> BTW I strongly recommend inclusion of a functional test in PRs like this. There are several examples to look at of plugins which have a `test` dep on `configuration-as-code` for this purpose.

This might be as easy as suggested. JCasC requires that the plugin targets Java 8 and higher Jenkins core version :cry:

Unless you are fine with bumping those values.

Some quick number crunching on http://stats.jenkins.io/pluginversions/github.html

92.88% are above 2.60 who have this plugin installed

er1c:
I'm going to submit a separate PR that upgrades `github-plugin` to `jekins.version` `2.60.3` & java8.   I have it compiling, and most of the tests successful, just working through other code changes that would be required for the updated dependencies.  That should unblock this PR to include a JCasC test.

er1c:
I would suggest closing this PR and referencing: https://github.com/jenkinsci/github-plugin/pull/210 for ongoing tracking.

PR #209, opened 2019-03-02
Java8 & jenkins.version 2.60.3

er1c:
* Update to jenkins.version 2.60.3 (allows for better recent plugin compatibility/tests)
    * Update to java8
    * Update tests for new signature/api changes

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/209)


er1c:
Looks like I missed the install tests from the CI, I'll see if I can fix those to get the green check

pom.xml
```diff
@@ -101,19 +110,33 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>git</artifactId>
-            <version>3.4.0</version>
+            <version>3.9.3</version>
+            <exclusions>
```
oleg-nenashev: If you have to do that, it is rather suppressing the issue. Better to declare dependencies directly in pom.xml, especially workflow-step-api which is used in production and excluded below

pom.xml
```diff
@@ -143,21 +166,22 @@
...
-            <scope>test</scope>
+            <version>3.1.0</version>
         </dependency>
 
+        <!--TEST DEPS-->
         <dependency>
             <groupId>org.apache.httpcomponents</groupId>
             <artifactId>httpclient</artifactId>
-            <version>4.5.2</version>
+            <version>4.5.3</version>
```
oleg-nenashev: Use https://github.com/jenkinsci/apache-httpcomponents-client-4-api-plugin instead ?

pom.xml
```diff
@@ -182,20 +206,63 @@
...
-            <version>${workflow.version}</version>
+            <version>${workflow-cps.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.jenkins-ci.plugins.workflow</groupId>
+            <artifactId>workflow-step-api</artifactId>
+            <version>${workflow-step-api.version}</version>
+            <scope>test</scope>
```
oleg-nenashev: I believe it should be in the default scope

src/main/java/org/jenkinsci/plugins/github/config/GitHubTokenCredentialsCreator.java
```diff
@@ -141,13 +140,17 @@ public FormValidation doCreateTokenByCredentials(
...
         try {
             token = createToken(
-                    notNull(creds, "Why selected creds is null?").getUsername(),
-                    creds.getPassword().getPlainText(),
+                    creds.getUsername(),
+                    Secret.toString(creds.getPassword()),
                     defaultIfBlank(apiUrl, GITHUB_URL)
             );
-        } catch (IOException e) {
+        } catch (Exception e) {
```
oleg-nenashev: why?

pom.xml
```diff
@@ -47,15 +47,24 @@
     </issueManagement>
 
     <properties>
-        <jenkins.version>1.625.3</jenkins.version>
-        <jenkins-test-harness.version>1.625.3</jenkins-test-harness.version>
+        <jenkins.version>2.60.3</jenkins.version>
```
timja: This is a very old LTS version,
Up the maintainers here, and I know this plugin is quite conservative.

But looking at the stats:
http://stats.jenkins.io/pluginversions/github.html

The last version on the stats (1.29.3) was released 5 months ago, so only need to look at the last column to see who is upgrading

I would target 2.138.1 as 96% of users who are upgrading are on it

If you go back to 2.121.1 you get 98% coverage

pom.xml
```diff
@@ -47,15 +47,24 @@
...
+        <jenkins-test-harness.version>2.47</jenkins-test-harness.version>
```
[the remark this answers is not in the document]
er1c: can you clarify this change request?  Do you want me to remove the version, and simply specify the version dependency in the pom `<dependency>` directly, or don't specify it at all (e.g. let it be brought in as a transitive dependency?)

src/main/java/org/jenkinsci/plugins/github/config/GitHubTokenCredentialsCreator.java
er1c: Wooops, this was unintended, this block of code was giving me hell for a while, and at one point I was trying to catch `IOException` and the `NPE` from `notNull`, I will revert this.

er1c:
@oleg-nenashev  can you clarify the comment about java files being made executable?

What I see:

```
-rw-r--r--  1 eric  staff  1571 Mar  1 08:30 ./src/main/java/com/coravy/hudson/plugins/github/GithubLinkAction.java`
-rw-r--r--  1 eric  staff  1867 Mar  1 08:30 ./src/main/java/org/jenkinsci/plugins/github/GitHubPlugin.java
```

pom.xml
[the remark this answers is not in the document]
oleg-nenashev: I would recommend a transitive one. Plugin POM keeps it up to date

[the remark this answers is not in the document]
er1c: ok, thanks.  I was attempting to be as conservative in changes as possible, I'll get these dependencies cleaned up and re-verify the tests

pom.xml
er1c: Upon reflection, I think, in the end, I ended up specifying these dependencies directly anyhow.  I'll remove the exclusions and update the pom accordingly.

pom.xml
er1c: All of the workflows?  Or just the step-api?

pom.xml
er1c: They were not needed, I backed them out.  Unfortunately the result of iterative work and not reviewing my PR well enough.

pom.xml
er1c: Changed!

pom.xml
casz: I believe the plugin pom is suggesting 2.138.4 because of security concerns :smile: https://github.com/jenkinsci/plugin-pom/pull/167

casz:
@er1c this is what he means with executables

image

a few of them got modified

er1c:
@casz thanks!  I updated the PR, not sure how that would have happened, I was editing them on a windows box in IDEA at one point...

er1c:
@oleg-nenashev bump, I think I addressed your outstanding concerns

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -155,7 +162,11 @@ public void run() {
      * Returns the file that records the last/current polling activity.
      */
     public File getLogFile() {
-        return new File(job.getRootDir(), "github-polling.log");
+        if (job == null) {
+            return null;
```
oleg-nenashev: not sure this is a good idea, but it does not make the situation worse at least. It would have been NPE anyway.
Would be nice to update Javadoc to document the new behavior

er1c:
I can update this (or create a separate PR), but I believe it would require targeting Jenkins version `2.143` to pickup this fix: https://issues.jenkins-ci.org/browse/JENKINS-52024.

I attempted to target the most conservative jenkins release, since it was previously targeted so low.

Is the java11 a blocker for this to be merged?  (I want to rebase the annotation changes in https://github.com/jenkinsci/github-plugin/pull/210, so that can also be reviewed/merged)

pom.xml
KostyaSha: @timja i'm on 2.107.3 i don't see any reason in forcing dependency update.

pom.xml
```diff
@@ -101,19 +104,19 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>git</artifactId>
-            <version>3.4.0</version>
+            <version>3.9.3</version>
```
KostyaSha: negative, 3.9.3 is the LATEST release, there is no any reason to force using the one latest single available version. What if it have bugs (it always has bugs), users will have no chance to downgrade it.

er1c:
I just switched back to the git `3.4.0` and changed back to the workflow version that it affords.  I re-audited the plugin changes I made, I believe this is about as conservative as dependencies as possible.

pom.xml
er1c: I changed back to `3.4.0` and reverted back to an older workflow version.  Did you have any other suggestions?

KostyaSha: test scopes are not critical

er1c:
Are you recommending I change them?

pom.xml
```diff
@@ -143,21 +148,23 @@
         <dependency><!-- exists in the core -->
             <groupId>org.jenkins-ci.modules</groupId>
             <artifactId>instance-identity</artifactId>
-            <version>1.3</version>
+            <version>2.1</version><!-- matches 2.60.3 -->
```
KostyaSha: So is it provided by core?

er1c: That's a good question, it was a previous annotation.  is there a good way to verify?

er1c: I actually just downloaded the core.jar file and it is not included, so I'm going to remove this comment & provided

KostyaSha: i remind that it's jar and not a plugin...

pom.xml
er1c: Are you recommending I change them?

KostyaSha: No opinion for test scope, but for hard deps it's bad practice to bump minimal requires dependencies.

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -155,7 +162,11 @@ public void run() {
...
+        } else {
```
jsoref: Not sure about coding style, but I wouldn't use an `else` here ("else after return").

er1c: If we want to get fancy, we can also update the signature to be a `RandomAccessFile`: https://gist.github.com/er1c/bcec81835c4eea05a208995e7c37a4da to ensure it gets deleted after creation, and guarantee it isn't null.  That requires a signature change tho.

er1c: I updated this block of code, and added additional null checks/`@CheckForNull` annotation.

pom.xml
```diff
@@ -110,10 +108,17 @@
             <version>2.2.0</version>
         </dependency>
 
+        <!-- From jenkins 2.60.3 -->
```
batmat: Not sure this is the best way. If you've got a conflict somewhere and this is pulled in by the core, a good way is to exclude it from other dependencies. This will allow the plugin to always behave with the version coming from the Core.

Doing this enforced here is going to cause differences when the Core is bumped and you miss aligning it to a possibly updated version in the Core dependencies.

pom.xml
batmat: Same idea as my previous comment. It's pulled by the core indeed.

See https://github.com/jenkinsci/jenkins/blob/cc964ef664250b6737e471ef546a46a37421cb4f/war/pom.xml#L95-L99

pom.xml
```diff
@@ -140,24 +145,25 @@
...
         </dependency>
 
-        <!--TEST DEPS-->
-        <dependency> <!-- from org.eclipse.jetty:jetty-server:9.2.15.v20160210 -->
+        <!-- Required for java8 -->
+        <dependency> <!-- from org.eclipse.jetty:jetty-server:9.4.5.v20170502-->
             <groupId>javax.servlet</groupId>
             <artifactId>javax.servlet-api</artifactId>
-            <scope>test</scope>
+            <version>3.1.0</version>
```
batmat: same, should probably be configured to take from the Core instead of pinned here

src/main/java/com/cloudbees/jenkins/GitHubPushCause.java
```diff
@@ -37,5 +37,17 @@ public GitHubPushCause(File pollingLog, String pusher) throws IOException {
     public String getShortDescription() {
         return format("Started by GitHub push by %s", trimToEmpty(pushedBy));
     }
+
+    @Override
+    public boolean equals(Object o) {
+        return o instanceof GitHubPushCause || super.equals(o);
```
batmat: This is a bug IIUC. This must be an `&&`, or any `GitHubPushCause` instance will be seen as equal?

er1c: Good catch, I also need to compare `pushedBy`

er1c: Does this look correct?

```
    @Override
    public boolean equals(Object o) {
        return o instanceof GitHubPushCause &&
                (null != o && this.pushedBy == ((GitHubPushCause) o).pushedBy) &&
                super.equals(o);
    }
```

pom.xml
er1c: Can you give me an example here?  In this case I'm trying to use the existing conventions to reduce the changes.  It sounds like you are offering an improvement on the existing codebase?  I think another reviewer earlier commented on something as was set to `provided`, and it may not have actually been provided by core.  I think `javax.servlet-api` might have been one example where I had to add the dependency to the POM for it to compile on java8 (vs just the tests compiling).  In that case you are just recommending remove the version tag from the dependency?

er1c: I think I figured it out, this should be resolved

pom.xml
er1c: This should be resolved, please verify

src/main/java/com/cloudbees/jenkins/GitHubPushCause.java
er1c: This should be resolved.

er1c:
@KostyaSha can we get this merged?

KostyaSha:
Waiting review from @jglick

pom.xml
jglick: Surely this could just be removed?

pom.xml
```diff
@@ -182,16 +183,30 @@
             <scope>test</scope>
         </dependency>
 
+        <!-- For GlobalMatrixAuthorizationStrategy -->
```
jglick: Usually you would instead use `MockAuthorizationStrategy`.

pom.xml
```diff
@@ -182,16 +183,30 @@
...
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
-            <artifactId>workflow-job</artifactId>
+            <artifactId>workflow-cps</artifactId>
             <version>${workflow.version}</version>
             <scope>test</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.jenkins-ci.ui</groupId>
+                    <artifactId>jquery-detached</artifactId>
```
jglick: Looks like a workaround for your use of an ancient (1.x) version.

src/main/java/com/cloudbees/jenkins/GitHubPushCause.java
```diff
@@ -37,5 +37,19 @@ public GitHubPushCause(File pollingLog, String pusher) throws IOException {
...
+    public boolean equals(Object o) {
+        return o instanceof GitHubPushCause
+                && this.pushedBy.equals(((GitHubPushCause) o).pushedBy)
+                && super.equals(o);
+    }
+
+    @Override
+    public int hashCode() {
+        int hash = super.hashCode();
+        hash = 89 * hash + (this.pushedBy != null ? this.pushedBy.hashCode() : 0);
```
jglick: Could also use `java.util.Objects` for these two.

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -135,16 +145,23 @@ public void run() {
...
                         cause = new GitHubPushCause(pushBy);
                     }
-                    if (asParameterizedJobMixIn(job).scheduleBuild(cause)) {
+
+                    if (null != job && asParameterizedJobMixIn(job).scheduleBuild(cause)) {
                         LOGGER.info("SCM changes detected in " + job.getFullName()
                                 + ". Triggering #" + job.getNextBuildNumber());
                     } else {
-                        LOGGER.info("SCM changes detected in " + job.getFullName() + ". Job is already in the queue");
+                        LOGGER.info("SCM changes detected in " + (null == job ? "null" : job.getFullName())
```
jglick: Presumably if `job == null` you could just return early from `run`?

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -234,7 +255,12 @@ public String getUrlName() {
         }
 
         public String getLog() throws IOException {
-            return Util.loadFile(getLogFile());
+            File logFile = getLogFile();
+            if (null == logFile) {
+                return "";
```
jglick: Better to throw an `IOException`.

pom.xml
er1c: This was previously suggested by @batmat I think, I originally pinned a version, but yes, this is being added as a work around.

pom.xml
er1c: Good to know, but I'm not sure I want to change that functionality.  I'm trying to keep the changes to a minimum.  @KostyaSha thoughts?

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
er1c: I think the impetus here was the code quality checker complaining about no null check, I'll see if I can add it further up, that is a better way to handle it, and can leave this line unchanged.

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
er1c: Good catch, I think I updated some of the changes to throw, and clearly not others.  I'll update this position.

src/main/java/com/cloudbees/jenkins/GitHubPushCause.java
er1c: 👍

pom.xml
er1c: Unfortunately, the tests fail if it isn't included. 


```
       <a href="https://jenkins.io/redirect/users-mailing-list">The users list</a> might be also useful in understanding what has happened.</p><h2>Stack trace</h2><pre style="margin:2em; clear:both">java.lang.NoSuchMethodError: javax.servlet.http.HttpServletResponse.getStatus()I
	at org.eclipse.jetty.server.handler.ErrorHandler.doError(ErrorHandler.java:142)
	at org.eclipse.jetty.server.handler.ErrorHandler.handle(ErrorHandler.java:78)
	at org.eclipse.jetty.server.Response.sendError(Response.java:655)
	at javax.servlet.http.HttpServletResponseWrapper.sendError(HttpServletResponseWrapper.java:118)
	at javax.servlet.http.HttpServletResponseWrapper.sendError(HttpServl
```

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
er1c: Yeah, spot bugs yells at me:

```

[2019-03-27T20:06:25.805Z] [ERROR] Possible null pointer dereference of hudson.triggers.Trigger.job in com.cloudbees.jenkins.GitHubPushTrigger$1.run() [com.cloudbees.jenkins.GitHubPushTrigger$1, com.cloudbees.jenkins.GitHubPushTrigger$1] Dereferenced at GitHubPushTrigger.java:[line 167]Known null at GitHubPushTrigger.java:[line 163] NP_NULL_ON_SOME_PATH
```
adding back the null check.

er1c:
@KostyaSha I believe all of the outstanding issues have been resolved, can you please review?

pom.xml
jglick: Fine, just FYI.

er1c:
@KostyaSha bump?

er1c:
> I think @jglick mention to throw exception from method, not throw on null on every code place.
> File permission change is still there

Throwing the exception doesn't fix the build failure (unfortunately), it still wants the null check code to be in place.

I can still add the exception in a few minutes if you would like.

The permissions were changed - if you view the diff I actually FIXED one of the permissions that was checked in with 0755:

image

er1c:
@KostyaSha any chance to get this looked at?  Thanks!

er1c:
@KostyaSha my weekly bump :) Thanks!

pom.xml
```diff
@@ -140,24 +138,27 @@
...
         </dependency>
 
-        <dependency><!-- exists in the core -->
+        <dependency>
             <groupId>org.jenkins-ci.modules</groupId>
             <artifactId>instance-identity</artifactId>
-            <version>1.3</version>
-            <scope>provided</scope>
+            <version>2.1</version><!-- Version matches https://mvnrepository.com/artifact/org.jenkins-ci.main/jenkins-war/2.60.3 -->
+            <scope>provided</scope><!-- https://wiki.jenkins.io/display/JENKINS/Instance+Identity "add a provided scope dependency to this module into your plugin" -->
```
jglick: What a mess. On my to-do list: make most if not all bundled modules into plain plugins.

pom.xml
jglick: Then someone did something wrong here. Track down what. It is not normal to depend explicitly on `servlet-api` from a plugin POM.

pom.xml
jglick: Suggest leaving a `TODO` comment requesting this be deleted as soon as the relevant `workflow-*` dependencies are upgraded to something not so ridiculously old.

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -154,7 +162,11 @@ public void run() {
     /**
      * Returns the file that records the last/current polling activity.
      */
-    public File getLogFile() {
+    public File getLogFile() throws IOException {
+        if (null == job) {
+            throw new IOException("Job was null, unable to getLogFile");
```
jglick: Should this not be detected earlier in the cycle, rather than throwing an exception?

pom.xml
er1c: Just to be clear, you don't want any changes here, right? :)

jglick: Right, sorry for not being clear—the way modules currently work, there is no clean way to depend on them.

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
er1c: The entire null checking here was added because the syntax package check start wouldn't pass.  Previously this, would have just thrown an NPE.  I don't have an actual example where this would occur, and I'm not familiar enough with the plugin to hazard any guesses.  Would you want this to be fixed as a precondition for this to get merged?  Or would a TODO here suffice (as it is also no worse than the current master, maybe better with the additional error messages that could help illustrate the condition in the wild)

jglick: Ah I see, back to https://github.com/jenkinsci/github-plugin/pull/209#discussion_r269091269. I think the best fix would be for the polling cycle to start off assigning `job` to a local variable. If null, exit right away—unconfigured, nothing to do. Otherwise you have a non-null reference which can be used without further checks within this method and passed to any helper methods it calls.

jglick:
Again I am _not_ a maintainer of this plugin, I just keep on getting review requested on this PR for whatever reason.

pom.xml
er1c: Just so I am clear too, you think the entire dependency should be able to be removed, not simply removing the scope/test and replacing with the version pin, right?

jglick: Probably yes. Try deleting the explicit dep and then checking `mvn dependency:tree` to see if something else is improperly pulling in this dep. I saw something about Jetty in there? Might suffice to just have another `<exclude>` somewhere.

er1c: I was able to remove the pinning, change the scope from `test` => `provided` and this does seem to be working...I'm pretty confused, as I think I tried this before and was not able to get it to work, but perhaps there was some additional pinning that I fixed that makes this work.

It looks like there are some servlet hooks here: https://github.com/jenkinsci/github-plugin/blob/497f43251e3bfc130f125f75e45f73e06894361d/src/main/java/com/cloudbees/jenkins/GitHubWebHookCrumbExclusion.java that is requiring the servlet extension, I'm still trying to figure out exactly what the purpose of it is.

pom.xml
er1c: TODO Added

pom.xml
jglick: It is normal enough for plugins to be referring to Servlet API classes—they are in the signature of several APIs defined in Jenkins core, and generally by [Stapler](http://stapler.kohsuke.org/reference.html). However the dep is normally defined in the parent POM; it should not need to be mentioned explicitly in the plugin’s POM.

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
er1c: I have updated this, I changed the signature for `getLogFile` to take a `@NonNull Job<?, ?>` (made it private) and updated a few references.  I didn't explicitly make a local `job` variable, since I could just do the single null check in the start of the `onPost`, which then lets me use the class's `job` as a non-null field in the rest of the code block fine, and revert some of the job-null checks that were added.

I added a `getLogFile` wrapper to maintain API compatibility (I have no idea who would use this!).  But I also didn't want to touch modifying the public API and any unintended consequences that would have.

er1c: It looks like spotbugs hates me :).  I'll add the local `job` variable as suggested to avoid this.

```

[2019-04-15T16:50:14.606Z] [ERROR] hudson.triggers.Trigger.job could be null and is guaranteed to be dereferenced in com.cloudbees.jenkins.GitHubPushTrigger$1.run() [com.cloudbees.jenkins.GitHubPushTrigger$1, com.cloudbees.jenkins.GitHubPushTrigger$1, com.cloudbees.jenkins.GitHubPushTrigger$1] Dereferenced at GitHubPushTrigger.java:[line 157]Dereferenced at GitHubPushTrigger.java:[line 160]Known null at GitHubPushTrigger.java:[line 156] NP_GUARANTEED_DEREF

[2019-04-15T16:50:14.606Z] [ERROR] Null passed for non-null parameter of GitHubPushTrigger.getLogFileForJob(Job) in com.cloudbees.jenkins.GitHubPushTrigger$1.run() [com.cloudbees.jenkins.GitHubPushTrigger$1, com.cloudbees.jenkins.GitHubPushTrigger$1] Method invoked at GitHubPushTrigger.java:[line 150]Known null at GitHubPushTrigger.java:[line 150] NP_NULL_PARAM_DEREF

[2019-04-15T16:50:14.606Z] [ERROR] Null passed for non-null parameter of GitHubPushTrigger.getLogFileForJob(Job) in com.cloudbees.jenkins.GitHubPushTrigger$1.runPolling() [com.cloudbees.jenkins.GitHubPushTrigger$1, com.cloudbees.jenkins.GitHubPushTrigger$1] Method invoked at GitHubPushTrigger.java:[line 107]Known null at GitHubPushTrigger.java:[line 107] NP_NULL_PARAM_DEREF

[2019-04-15T16:50:14.606Z] [INFO] 

```

jglick: That sounds easiest. Again these are just my suggestions as an outside observer of the plugin.

er1c: These changes are already pushed/and CI came back green

er1c:
@KostyaSha I think this is ready for merge, you are the current maintainer, right?

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -91,22 +94,33 @@ public void onPost(String triggeredByUser) {
...
                         PrintStream logger = listener.getLogger();
+
                         long start = System.currentTimeMillis();
                         logger.println("Started on " + DateFormat.getDateTimeInstance().format(new Date()));
                         if (event.getOrigin() != null) {
                             logger.format("Started by event from %s on %tc%n", event.getOrigin(), event.getTimestamp());
                         }
-                        boolean result = SCMTriggerItems.asSCMTriggerItem(job).poll(listener).hasChanges();
+                        SCMTriggerItem item = SCMTriggerItems.asSCMTriggerItem(currentJob);
+                        if (null == item) {
```
KostyaSha: As we are updating to newer java, please update null checks to human readable `Objects.isNull/nonNull()` 
(only in modified places)

KostyaSha:
> I just keep on getting review requested on this PR for whatever reason.

Thanks, dependencies were looking weird before. Now looks better.

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
er1c: Updated!  I'm just waiting for the checks to make sure nothing changed with the spotbugs

KostyaSha: I think it can't understand checks, so ignore it.

er1c:
Spotbugs is complainig about the last line:

```
                        SCMTriggerItem item = SCMTriggerItems.asSCMTriggerItem(currentJob);
                        if (Objects.isNull(item)) {
                            throw new RuntimeException("Job is not an SCMTriggerItem: " + currentJob);
                        }
                        boolean result = item.poll(listener).hasChanges();
```

Despite the above null check, I can revert these back to `null == item` and it will fix the spotbugs error, or leave them inplace.

er1c:
> Spotbugs is complainig about the last line:
> 
> ```
>                         SCMTriggerItem item = SCMTriggerItems.asSCMTriggerItem(currentJob);
>                         if (Objects.isNull(item)) {
>                             throw new RuntimeException("Job is not an SCMTriggerItem: " + currentJob);
>                         }
>                         boolean result = item.poll(listener).hasChanges();
> ```
> Despite the above null check, I can revert these back to `null == item` and it will fix the spotbugs error, or leave them inplace.

I updated this line back to `null == item` for spotbugs:

```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 02:06 min
[INFO] Finished at: 2019-04-17T20:00:16-07:00
```

er1c:
PR is green

batmat:
@KostyaSha this looks like quite correct now? Do you think we could consider merging this, or can you clarify what would be missing? Thanks!

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -91,22 +95,33 @@ public void onPost(String triggeredByUser) {
...
+                            throw new RuntimeException("Job is not an SCMTriggerItem: " + currentJob);
```
casz: Could we please avoid Runtime Exception and throw a proper exception?

Like illegal state exception?

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -155,6 +172,17 @@ public void run() {
      * Returns the file that records the last/current polling activity.
      */
     public File getLogFile() {
+        try {
+            return getLogFileForJob(notNull(job, "Job can't be null!"));
+        } catch (IOException ex) {
+            throw new RuntimeException(ex);
```
casz: Why not just add throws to the method and let Jenkins handle it. Like you did with getLogFileForJob.
You're rethrowing the exception just wrapped in another exception. Hence it makes sense to add throws to the method signature.

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
er1c: I threw a runtime exception, because the code already had a catch for it below/didn't want to make a signature change.

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
er1c: I just didnt want to make a signature change for the method if there were any reverse-compatibility API issues.

er1c: I was actually tempted to @Deprecate this method, since I don't know how/why it is exposed outside this method

casz: This is exposed for Jenkins to provide a view of the trigger log

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
casz: Illegal state exception is also a runtime exception just more descriptive

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
er1c: Are you sure? As far as I can tell, it isn't implementing any interface, it compiles/tests fine without it

er1c: @KostyaSha this public `getLogFile` isn't used anywhere other than the `github-plugin` I would like to simply remove the public interface,

https://github.com/jenkinsci/github-plugin/search?q=getlogfile&unscoped_q=getlogfile
https://github.com/search?q=org%3Ajenkinsci+GitHubPushTrigger+getlogfile&type=Code

As far as I can tell this isn't used by any other project, any objections, or should I simply mark it as `@Deprecated`?

KostyaSha: jenkinsci is not the only place where plugins exist, there are a lot of private plugins, if it was public it should be public.

er1c: OK, is the existing code fine, or should I change the signature to throw IOException?

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -92,13 +95,19 @@ public void onPost(String triggeredByUser) {
      * Called when a POST is made.
      */
     public void onPost(final GitHubTriggerEvent event) {
+        if (Objects.isNull(job)) {
+            return; // nothing to do
+        }
+
+        Job<?, ?> currentJob = notNull(job, "Job can't be null");
```
KostyaSha: Why is it needed?

er1c: This was suggested by @jglick the dereferencing allows us to guarantee the job continues to be not-null for the rest of the execution block.

er1c:
@koystasha thank you!

PR #210, opened 2019-03-02
JENKINS-55920: add CasC support to GitHubPluginConfig w/Tests

er1c:
This is based upon both https://github.com/jenkinsci/github-plugin/pull/205 and https://github.com/jenkinsci/github-plugin/pull/209

I can rebase the PR once #209 is addressed to make the PR cleaner.

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/210)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -54,6 +54,7 @@
         <concurrency>1</concurrency>
         <java.level>8</java.level>
         <workflow.version>1.14.2</workflow.version>
+        <configuration-as-code.version>1.12</configuration-as-code.version>
     </properties>
 
     <repositories>
@@ -187,6 +188,21 @@
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>matrix-auth</artifactId>
             <version>1.1</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>io.jenkins</groupId>
+            <artifactId>configuration-as-code</artifactId>
+            <version>${configuration-as-code.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>io.jenkins</groupId>
+            <artifactId>configuration-as-code</artifactId>
+            <version>${configuration-as-code.version}</version>
+            <classifier>tests</classifier>
             <scope>test</scope>
         </dependency>
 
--- a/src/main/resources/org/jenkinsci/plugins/github/config/GitHubPluginConfig/config.groovy
+++ b/src/main/resources/org/jenkinsci/plugins/github/config/GitHubPluginConfig/config.groovy
@@ -26,8 +26,7 @@ f.section(title: descriptor.displayName) {
                 table(width: "100%", style: "margin-left: 7px;") {
                     f.optionalBlock(title: _("Specify another hook URL for GitHub configuration"),
                             inline: true,
-                            field: "overrideHookUrl",
-                            checked: instance.overrideHookURL) {
+                            checked: instance.isOverrideHookUrl) {
                         f.entry(field: "hookUrl") {
                             f.textbox(checkMethod: "post")
                         }
--- a/src/test/java/org/jenkinsci/plugins/github/config/ConfigAsCodeTest.java
+++ b/src/test/java/org/jenkinsci/plugins/github/config/ConfigAsCodeTest.java
@@ -1,0 +1,73 @@
+package org.jenkinsci.plugins.github.config;
+
+import io.jenkins.plugins.casc.ConfigurationAsCode;
+import io.jenkins.plugins.casc.misc.ConfiguredWithCode;
+import io.jenkins.plugins.casc.misc.JenkinsConfiguredWithCodeRule;
+import org.junit.Rule;
+import org.junit.Test;
+import org.jvnet.hudson.test.JenkinsRule;
+
+import java.util.List;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.both;
+import static org.hamcrest.Matchers.hasItems;
+import static org.hamcrest.Matchers.hasSize;
+import static org.hamcrest.Matchers.is;
+import static org.jenkinsci.plugins.github.test.GitHubServerConfigMatcher.withClientCacheSize;
+import static org.jenkinsci.plugins.github.test.GitHubServerConfigMatcher.withCredsId;
+import static org.jenkinsci.plugins.github.test.GitHubServerConfigMatcher.withApiUrl;
+import static org.jenkinsci.plugins.github.test.GitHubServerConfigMatcher.withIsManageHooks;
+import static org.jenkinsci.plugins.github.test.GitHubServerConfigMatcher.withName;
+
+public class ConfigAsCodeTest {
+
+    @Rule
+    public JenkinsConfiguredWithCodeRule r = new JenkinsConfiguredWithCodeRule();
+
+    @Test
+    @ConfiguredWithCode("configuration-as-code.yml")
+    public void shouldSupportConfigurationAsCode() throws Exception {
+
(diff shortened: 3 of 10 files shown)


pom.xml
```diff
@@ -182,20 +206,70 @@
...
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>matrix-auth</artifactId>
+            <version>2.3</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>io.jenkins</groupId>
+            <artifactId>configuration-as-code</artifactId>
+            <version>1.0</version>
```
timja: use the latest released version please (1.7) https://github.com/jenkinsci/configuration-as-code-plugin/releases

casz:
@er1c would be great if you could rebase:
This diff is no good: https://github.com/dgarzon/github-plugin/compare/HOTFIX-add-casc-support...er1c:HOTFIX-add-casc-support

er1c:
@casz  https://github.com/er1c/github-plugin/pull/1

pom.xml
```diff
@@ -207,6 +207,13 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>io.jenkins</groupId>
+            <artifactId>configuration-as-code</artifactId>
+            <version>1.7</version>
```
casz: Please update to a newer version and consider adding the `tests` classifier.

src/test/java/org/jenkinsci/plugins/github/config/ConfigAsCodeTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+public class ConfigAsCodeTest {
+
+    @Rule
+    public JenkinsRule r = new JenkinsRule();
+
+    @Test
+    public void shouldSupportConfigurationAsCode() throws Exception {
+        ConfigurationAsCode.get().configure(
+                ConfigAsCodeTest.class.getResource("configuration-as-code.yml").toString());
```
casz: once you add the `tests` classifier, please use the annotation `@ConfiguredWithCode("configuration-as-code.yml")`

er1c:
This has been rebased, and the outstanding review comments addressed.

src/main/resources/org/jenkinsci/plugins/github/config/GitHubPluginConfig/config.groovy
```diff
@@ -26,8 +26,7 @@ f.section(title: descriptor.displayName) {
                 table(width: "100%", style: "margin-left: 7px;") {
                     f.optionalBlock(title: _("Specify another hook URL for GitHub configuration"),
                             inline: true,
-                            field: "overrideHookUrl",
-                            checked: instance.overrideHookURL) {
+                            checked: instance.isOverrideHookUrl) {
```
KostyaSha: either method call either method call via field name, i think it wouldn't work

casz: JCasC does say no to optional blocks, use optional property https://github.com/jenkinsci/configuration-as-code-plugin/blob/master/docs/PLUGINS.md#step-4

casz: Or move it out of the optional block, it smells :cry:

src/test/java/org/jenkinsci/plugins/github/config/ConfigAsCodeTest.java
```diff
@@ -0,0 +1,73 @@
...
+                        .and(withCredsId(is("private_cred_id")))
+                        .and(withClientCacheSize(is(40)))
+                        .and(withIsManageHooks(is(false)))
+                ));
+    }
+
+    @Test
+    @ConfiguredWithCode("configuration-as-code.yml")
+    public void export_configuration() throws Exception {
+        /* TODO (From JCASC): Need to provide some YAML assertion library so that the resulting exported yaml
```
timja: We need to update the plugin docs but it is being tested in this way in a few plugins and in the configuration-as-code repo:
https://github.com/jenkinsci/azure-keyvault-plugin/blob/master/src/test/java/org/jenkinsci/plugins/azurekeyvaultplugin/ConfigAsCodeTest.java#L33

src/test/resources/org/jenkinsci/plugins/github/config/configuration-as-code.yml
```diff
@@ -0,0 +1,17 @@
+unclassified:
+
+  githubpluginconfig:
```
timja: maybe another more minimal test which shows what you need if you set manage hooks false and no hook url?

src/main/resources/org/jenkinsci/plugins/github/config/GitHubPluginConfig/config.groovy
er1c: This isn't part of my wheel house, I think the way I have it setup, conforms to the example here: https://wiki.jenkins.io/display/JENKINS/Jelly+form+controls

I'm removing the "field" from the toggling of the checkbox to just toggle it based upon the value,  I'll see if I can add a test to verify that the checkbox is "checked" based upon the current code or correct if that doesn't work.

src/test/java/org/jenkinsci/plugins/github/config/ConfigAsCodeTest.java
timja: I've updated the docs now:
https://github.com/jenkinsci/configuration-as-code-plugin/blob/master/docs/PLUGINS.md#model-export-test

startnow65:
@er1c Thanks for the good work!
Any idea when this would be merged ? I need this so I can set my hook url with JCasC.

Thanks

src/main/resources/org/jenkinsci/plugins/github/config/GitHubPluginConfig/config.groovy
KostyaSha: `isOverrideHookUrl()` or `instance.overrideHookURL`  code had correct call

src/test/java/org/jenkinsci/plugins/github/config/ConfigAsCodeTest.java
```diff
@@ -0,0 +1,73 @@
...
+            stream can be checked for expected content. */
+        ConfigurationAsCode.get().export(System.out);
```
KostyaSha: never user System.out, user loggers if you want to print something.

m-barthelemy:
Hi, we're having this issue where `hookUrl` is ignored in configuration as code.
This PR seems to fix it. Will it get merged?

oleg-nenashev:
I'd guess https://github.com/jenkinsci/github-plugin/pull/210/files#r277181851 is still a blocker, no?

thirstydeveloper:
Any movement on this? My team is blocked by not being able to set the `overrideHookUrl` property.

Misteur-Z:
Basically the only thing to set manually at startup or after a Jenkins upgrade. I would be happy to help with this issue, I can do some testing but I am no Java/Groovy dev.

thefirstofthe300:
Based on @er1c's Github contribution history, it looks like he probably doesn't have bandwidth to work on this anymore. Does anyone here have the Java experience to take over this PR?

I am also hitting this particular issue.

er1c:
> Based on @er1c's Github contribution history, it looks like he probably doesn't have bandwidth to work on this anymore. Does anyone here have the Java experience to take over this PR?
> 
> I am also hitting this particular issue.

I'm currently slammed and don't have time for this.  For anyone who is blocked, you can work around it via: https://gist.github.com/er1c/c057640d43c0dce149b0a0c23d1ce153#file-jenkins-github-yaml-L80

casz:
Thanks @er1c for picking this up 👍

PR #212, opened 2019-04-01
[JENKINS-56830] alternative use of 'html_url' in payload of event

McFoggy:
As documented in github API docs (https://developer.github.com/v3/repos/#response) the url field of the event should/could point to the API endpoint.
In the github-plugin it is expected to work on the public html url which is handled by the 'html_url' field of the event.
This commit thus try as before to build a GitHubRepositoryName from the 'url' field ; if that fails, as a fallback it also tries the 'html_url' field.
See also https://github.community/t5/GitHub-API-Development-and/consistency-of-repository-url-between-event-types/td-p/21209 for some explanations.

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/212)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/github/webhook/subscriber/DefaultPushGHEventSubscriber.java
+++ b/src/main/java/org/jenkinsci/plugins/github/webhook/subscriber/DefaultPushGHEventSubscriber.java
@@ -76,7 +76,22 @@ public class DefaultPushGHEventSubscriber extends GHEventsSubscriber {
         URL repoUrl = push.getRepository().getUrl();
         final String pusherName = push.getPusher().getName();
         LOGGER.info("Received PushEvent for {} from {}", repoUrl, event.getOrigin());
-        final GitHubRepositoryName changedRepository = GitHubRepositoryName.create(repoUrl.toExternalForm());
+        GitHubRepositoryName fromEventRepository = GitHubRepositoryName.create(repoUrl.toExternalForm());
+
+        if (fromEventRepository == null) {
+            // On push event on github.com url === html_url
+            // this is not consistent with the API docs and with hosted repositories
+            // see https://goo.gl/c1qmY7
+            // let's retry with 'html_url'
+            URL htmlUrl = push.getRepository().getHtmlUrl();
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/github/webhook/subscriber/DefaultPushGHEventSubscriber.java
```diff
@@ -76,7 +76,22 @@ protected void onEvent(final GHSubscriberEvent event) {
...
+
+        if (fromEventRepository == null) {
+            // On push event on github.com url === html_url
+            // this is not consistent with the API docs and with hosted repositories
+            // see https://goo.gl/c1qmY7
+            // let's retry with 'html_url'
+            URL htmlUrl = push.getRepository().getHtmlUrl();
+            fromEventRepository = GitHubRepositoryName.create(htmlUrl.toExternalForm());
+            if (fromEventRepository != null) {
+                LOGGER.info("PushEvent handling: 'html_url' field "
```
KostyaSha: info is printed by default in jenkins logger, this will lead to redundant messages
Please change to trace or debug

McFoggy: done in c0c83fb commit.

McFoggy:
@KostyaSha modification done as as requested ; please look at latest push. Thx

solvingj:
This should have printed it's message with `Logger.warning()` because currently, if this alternate path is taken, the last message in the default system log is: 

```
WARNING c.c.jenkins.GitHubRepositoryName#create: Could not match URL
```
Which would seem like it's the end of the road for this PUSH hook being processed.  But, adding a custom logger and turning up the debug level, we can see this happening after: 

```
PushEvent handling: 'html_url' field has been used to retrieve project information (instead of default 'url' field)
```
Which indicates that in-fact, we did in-fact successfully create a `GitHubRepositoryName` and assign it to `fromEventRepository`. 

Unfortunately, builds still do not trigger on this path.  But, this comment is only to suggest changing the log level here.

PR #214, opened 2019-05-06
Also use GitHub Project property for webhooks

br0ch0n:
It is possible for one to be in a situation where the SCM address used by Jenkins for cloning doesn't exactly match the address advertised by github in its webhooks.  E.g. your github enterprise installation thinks its name is github.corporate.internal, an address perhaps reachable from a user's desktop but not from your jenkins slave, which must use github.proxy.com.

This PR tells the jenkins webhook receiver to consider the Github Project property field **in addition to** the SCM field when considering candidates to "poke", thus it's safe and shouldn't break existing usage.  Note that the [Github Pull Request Builder](https://github.com/jenkinsci/ghprb-plugin) plugin seems to use the Github Project property field exclusively rather than the SCM field.

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/214)

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
+++ b/src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
@@ -1,5 +1,6 @@
 package com.cloudbees.jenkins;
 
+import com.coravy.hudson.plugins.github.GithubProjectProperty;
 import hudson.EnvVars;
 import hudson.Extension;
 import hudson.ExtensionList;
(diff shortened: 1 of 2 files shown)


src/main/java/com/cloudbees/jenkins/GitHubRepositoryNameContributor.java
```diff
@@ -131,6 +132,17 @@ public void parseAssociatedNames(Item item, Collection<GitHubRepositoryName> res
...
+                // under General job properties to the result collection
+                // as it may differ from the SCM URL due to proxy usage
+                Job job = (Job) item;
+                GithubProjectProperty ghpp = (GithubProjectProperty) job.getProperty(GithubProjectProperty.class);
+                if (ghpp != null) {
+                    GitHubRepositoryName githubrepo = GitHubRepositoryName.create(ghpp.getProjectUrl().toString());
+                    if (githubrepo != null) {
+                        result.add(githubrepo);
+                    }
+                }
```
emmaroberts-nbs, line 145: Should this be within the if statement for `triggerItem != null`?

Also where are the tests to test this functionality?

KostyaSha: This contributor is only about SCM and trigger existing in this plugin is only about SCM. 

https://github.com/KostyaSha/github-integration-plugin/ is based on GHPP and works fine without scm.

PR #215, opened 2019-05-20
Update github-api version. Excludes for jackson. Bump findbugs version.

kshultzCB:
* Dependency update for github-api
* Excludes added for `jackson-annotations`, `jackson-core`, and `jackson-databind`, which are here via (test-scoped) wiremock libraries. 
* This brought with it a version bump for findbugs. I'm running Maven 3.6.0 locally, so following [this article](https://stackoverflow.com/questions/53676071/maven-clean-command-java-util-collections-unmodifiablerandomaccesslist-to-prope), I bumped to 3.0.4.

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/215)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -50,7 +50,7 @@
         <jenkins.version>2.60.3</jenkins.version>
         <release.skipTests>false</release.skipTests>
         <maven.javadoc.skip>true</maven.javadoc.skip>
-        <findbugs-maven-plugin.version>3.0.2</findbugs-maven-plugin.version>
+        <findbugs-maven-plugin.version>3.0.4</findbugs-maven-plugin.version>
         <concurrency>1</concurrency>
         <java.level>8</java.level>
         <workflow.version>1.14.2</workflow.version>
@@ -93,7 +93,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>github-api</artifactId>
-            <version>1.90</version>
+            <version>1.95</version>
         </dependency>
 
         <dependency>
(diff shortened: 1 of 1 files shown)


abayer:
So this is to deal with a security scanner warning about the jackson version - it's goofy, yeah, but it is what it is. @KostyaSha - would you be willing to help me out and merge/release this? =) Thanks!

pom.xml
```diff
@@ -93,7 +93,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>github-api</artifactId>
-            <version>1.90</version>
+            <version>1.95</version>
```
KostyaSha: Why this dep needs go be updated? Is there any binary compatibility change?

kshultzCB: It has to do with the security scanner warning about jackson versions. github-api 1.90 has a dependency on [jackson-api 2.7.3](https://github.com/jenkinsci/github-api-plugin/blob/github-api-1.90/pom.xml#L31-L35), whereas 1.95 has its dependency on [2.9.7.1](https://github.com/jenkinsci/github-api-plugin/blob/github-api-1.95/pom.xml#L31-L35).

KostyaSha: this is not the problem of this plugin, this plugin is only defining minimal required binary compatible version. This is how jenkins dependencies works, hpi is based on maven versioning.

oleg-nenashev: FTR I agree with @KostyaSha here. Unless there is a strong need to update (API, integration testing with newer version, and so on), there is no practical need to update

madchap:
Hey @KostyaSha 

I would have a vested interest in seeing this PR in a release soon. Do you think it would be possible to plan for one at your convenience?

Thank you.

PR #218, opened 2019-08-01
Enable Release Drafter for the repository

oleg-nenashev:
@lanwen was using GitHub releases for last releases anyway. Why don't we automate it?

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/218)

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -1,0 +1,5 @@
+# https://github.com/jenkinsci/.github/blob/master/.github/release-drafter.adoc
+_extends: .github
+version-template: $MAJOR.$MINOR.$PATCH
+tag-template: github-$NEXT_PATCH_VERSION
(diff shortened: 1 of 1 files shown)


.github/release-drafter.yml
```diff
@@ -0,0 +1,5 @@
+# https://github.com/jenkinsci/.github/blob/master/.github/release-drafter.adoc
+_extends: .github
+version-template: $MAJOR.$MINOR.$PATCH
+tag-template: github-$NEXT_PATCH_VERSION
```
KostyaSha: is it git tags?

oleg-nenashev:
Yes

.github/release-drafter.yml
KostyaSha: Should they match existing git tags?

oleg-nenashev: Right

PR #221, opened 2019-10-07
End Pr210

KostyaSha:
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/221)


KostyaSha:
@timja ping

pom.xml
```diff
@@ -54,6 +54,7 @@
         <concurrency>1</concurrency>
         <java.level>8</java.level>
         <workflow.version>1.14.2</workflow.version>
+        <configuration-as-code.version>1.12</configuration-as-code.version>
```
timja:
```suggestion
        <configuration-as-code.version>1.32</configuration-as-code.version>
```

src/test/java/org/jenkinsci/plugins/github/test/GitHubServerConfigMatcher.java
```diff
@@ -31,4 +84,50 @@ protected String featureValueOf(GitHubServerConfig actual) {
...
+                return valueOrNull(actual, "name");
+            }
+        };
+    }
+
+    private static String valueOrNull(Mapping mapping, String key) {
+        try {
+            return mapping.get(key).asScalar().getValue();
+        } catch (NullPointerException | ConfiguratorException e) {
+//            LOG.error("", e);
```
timja: ?

src/test/java/org/jenkinsci/plugins/github/test/GitHubServerConfigMatcher.java
```diff
@@ -31,4 +84,50 @@ protected String featureValueOf(GitHubServerConfig actual) {
...
+            throw new AssertionError(key);
+//            return "UNKNOWN VALUE: " + key;
```
timja: ?

src/test/java/org/jenkinsci/plugins/github/config/ConfigAsCodeTest.java
```diff
@@ -0,0 +1,96 @@
...
+                        .and(withApiUrl(is("https://api.some.com")))
+                        .and(withCredsId(is("private_cred_id")))
+                        .and(withClientCacheSize(is(40)))
+                        .and(withIsManageHooks(is(false)))
+        ));
+    }
+
+    @Test
+    @ConfiguredWithCode("configuration-as-code.yml")
+    public void exportConfiguration() throws Exception {
```
timja: you could probably write a simpler test with a fixture, there's nothing dynamic here so should be straightforward:
https://github.com/jenkinsci/slack-plugin/blob/master/src/test/java/jenkins/plugins/slack/jcasc/ConfigurationAsCodeTest.java#L34

KostyaSha: What is the difference with the source yaml?

src/test/java/org/jenkinsci/plugins/github/test/GitHubServerConfigMatcher.java
KostyaSha: There is no methods to check existance of the key. Either get once with removal either get NPE. Exception is expected against object that has no key, no need to log, so commented.

src/test/java/org/jenkinsci/plugins/github/test/GitHubServerConfigMatcher.java
KostyaSha: the same

src/test/java/org/jenkinsci/plugins/github/config/ConfigAsCodeTest.java
timja: the source yaml is the same, the expected one doesn't have the attribute that you do a .get on, in the case of slack it's missing `slackNotifier` but has all the fields under it

src/test/java/org/jenkinsci/plugins/github/test/GitHubServerConfigMatcher.java
timja: more around is this commented out code intended to stay.

src/test/java/org/jenkinsci/plugins/github/config/ConfigAsCodeTest.java
KostyaSha: i not sure also btw, final yaml is missing default return settings, is it ok?

KostyaSha: @timja i can merge&release finally this caasc support and i can accept PR for test update if you want to change it

src/test/java/org/jenkinsci/plugins/github/test/GitHubServerConfigMatcher.java
KostyaSha: cleaned

src/test/java/org/jenkinsci/plugins/github/config/ConfigAsCodeTest.java
timja: probably not worth it, it works and is unlikely to ever need to change.

KostyaSha:
Thanks for CasC :) Waiting for read only global options.

casz:
Thanks @KostyaSha for finishing this up 😊

JackHowa:
This seems like it's a problem. https://issues.jenkins-ci.org/browse/JENKINS-60738

PR #222, opened 2019-11-16
Added option to reuse GitSCM's excluded users list

pbalint:
This change enables the github plugin to respect GitSCM's excluded user list (which is used when polling) for github webhooks:

image


pbalint:
@KostyaSha , @lanwen May I get a review, please?

src/main/java/com/cloudbees/jenkins/GitHubPushTrigger.java
```diff
@@ -98,6 +104,22 @@ public void onPost(final GitHubTriggerEvent event) {
...
+                SCM scm = ((AbstractProject<?, ?>) job).getScm();
+                if (scm instanceof GitSCM) {
+                    UserExclusion exclusions = ((GitSCM) scm).getExtensions().get(UserExclusion.class);
+                    if (exclusions != null) {
+                        excludedUsers = exclusions.getExcludedUsersNormalized();
+                    }
+                }
+            }
+
+            if (excludedUsers != null && excludedUsers.contains(event.getTriggeredByUser())) {
```
KostyaSha: Should be ignorecase

pbalint: It appears to me the polling is case sensitive as well:
https://github.com/jenkinsci/git-plugin/blob/a7ad6eb38a00397c4a656b53d99bfbad8ea3b2cc/src/main/java/hudson/plugins/git/extensions/impl/UserExclusion.java#L63
So if the webhooks checks the exclusion list in a case insensitive manner, there will be a counter-intuitive difference...

KostyaSha: gitscm extension is for git users, github users case insensitive you do comparison in terms of GH functionality it should be ignorecase

pbalint: Done

pbalint:
Ok, I've updated the PR and changed it to be case insensitive...

PR #225, opened 2020-02-10
[JENKINS-49332] Better error messages

carlossg:
when github repo webhook can't be created

Don't throw exceptions as we are handling the error, just print to the log

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/225)


src/main/java/org/jenkinsci/plugins/github/webhook/WebhookManager.java
```diff
@@ -2,6 +2,7 @@
 
 import com.cloudbees.jenkins.GitHubRepositoryName;
 import com.google.common.base.Function;
+import com.google.common.base.Optional;
```
jglick, line 5: Seems to be a compilation error.

KostyaSha:
> This would really be useful, as I get these stack traces all the time. Would a cleaned-up version of this PR be welcome @KostyaSha?

I were waiting user to fixup https://github.com/jenkinsci/github-plugin/pull/226  but any working variant will be ok

carlossg:
I think I forgot to come back to finish this.
It should work now, with the fixed tests

carlossg:
CI failing with not enough space on disk

jglick:
> I were waiting user to fixup #226

Ah, even better to not have to show an error at all because we never try to do something we are not permitted to do.

carlossg:
I've changed it to not show the problem if there are no permissions, lowered the log level to debug and added more testing

jimklimov:
About #226 - sorry I fell out of the loop, hope to get back to that soon. Or feel free to follow up with PRing wanted cleanups on top of that :)

psalaberria002:
Can this one be merged? We have hundreds of repos adding the stacktrace to the log. It would be very nice to get that fixed.

KostyaSha:
@carlossg @jglick is it done?

carlossg:
yes it is done

PR #228, opened 2020-04-03
Upgrade to github-api 1.114 and okhttp3

bitwiseman:
@KostyaSha 
This change updates github-plugi to use okhttp3. Okhttp 2.7.5 is seriously out of date and no longer supported. We should move to using okhttp3.

@KostyaSha @lanwen  

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/228)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.57</version>
+        <version>4.2</version>
         <relativePath />
     </parent>
 
@@ -47,7 +47,7 @@
     </issueManagement>
 
     <properties>
-        <jenkins.version>2.60.3</jenkins.version>
+        <jenkins.version>2.164.3</jenkins.version>
         <release.skipTests>false</release.skipTests>
         <maven.javadoc.skip>true</maven.javadoc.skip>
         <findbugs-maven-plugin.version>3.0.4</findbugs-maven-plugin.version>
@@ -80,22 +80,9 @@
         </dependency>
 
         <dependency>
-            <groupId>org.slf4j</groupId>
-            <artifactId>slf4j-jdk14</artifactId>
-            <version>${slf4jVersion}</version>
-        </dependency>
-
-        <dependency>
-            <groupId>com.squareup.okhttp</groupId>
-            <artifactId>okhttp-urlconnection</artifactId>
-            <version>2.7.5</version>
-            <optional>false</optional>
-        </dependency>
-
-        <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>github-api</artifactId>
-            <version>1.90</version>
+            <version>1.114.2</version>
         </dependency>
 
         <dependency>
@@ -143,14 +130,14 @@
         <dependency>
             <groupId>org.jenkins-ci.modules</groupId>
             <artifactId>instance-identity</artifactId>
-            <version>2.1</version><!-- Version matches https://mvnrepository.com/artifact/org.jenkins-ci.main/jenkins-war/2.60.3 -->
+            <version>2.2</version>
             <scope>provided</scope><!-- https://wiki.jenkins.io/display/JENKINS/Instance+Identity "add a provided scope dependency to this module into your plugin" -->
         </dependency>
 
         <dependency>
             <groupId>javax.servlet</groupId>
             <artifactId>javax.servlet-api</artifactId>
-            <scope>provided</scope><!-- Provided by core: https://mvnrepository.com/artifact/org.jenkins-ci.main/jenkins-core/2.60.3 -->
+            <scope>provided</scope>
         </dependency>
 
         <!--TEST DEPS-->
@@ -311,6 +298,16 @@
         </dependency>
 
     </dependencies>
+
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>org.jenkins-ci</groupId>
+                <artifactId>annotation-indexer</artifactId>
+                <version>1.12</version>
+            </dependency>    
+        </dependencies>
+    </dependencyManagement>
 
     <build>
         <plugins>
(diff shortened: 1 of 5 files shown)


src/main/java/org/jenkinsci/plugins/github/internal/GitHubClientCacheOps.java
```diff
@@ -94,6 +95,7 @@ public static Path getBaseCacheDir() {
      *
      * @param configs active server configs to exclude caches from cleanup
      */
+    @SuppressFBWarnings(value = "RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE")
```
lanwen, line 98: hey, can you give more insights of why is this necessary?

src/main/java/org/jenkinsci/plugins/github/internal/GitHubLoginFunction.java
```diff
@@ -107,31 +106,15 @@ private Proxy getProxy(String apiUrl) {
      * @return connector to be used as backend for client
      */
     private OkHttpConnector connector(GitHubServerConfig config) {
-        OkHttpClient client = new OkHttpClient().setProxy(getProxy(defaultIfBlank(config.getApiUrl(), GITHUB_URL)));
+        OkHttpClient.Builder builder = BASECLIENT.newBuilder()
```
lanwen, line 109: you can skip instance creation with builder creation directly via new OkHttpClient.Builder()

src/main/java/org/jenkinsci/plugins/github/util/JobInfoHelpers.java
```diff
@@ -111,7 +113,8 @@ public boolean apply(ITEM item) {
         if (item instanceof ParameterizedJobMixIn.ParameterizedJob) {
             ParameterizedJobMixIn.ParameterizedJob pJob = (ParameterizedJobMixIn.ParameterizedJob) item;
 
-            for (Trigger candidate : pJob.getTriggers().values()) {
+            Map<TriggerDescriptor, Trigger<?>> triggerMap = pJob.getTriggers();
```
lanwen, line 116: can you give more details regarding this change?

pom.xml
```diff
@@ -5,7 +5,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.54</version>
+        <version>4.2</version>
```
lanwen, line 8: 4 uses kotlin runtime, so I would rather stick with 3 to reduce this dependency which could be dangerous in a jenkins world

src/main/java/org/jenkinsci/plugins/github/util/JobInfoHelpers.java
KostyaSha: return type (generics) is changed on new core

pom.xml
bitwiseman: @lanwen 
This is the plugin pom version number not the okhttp3 version number.

src/main/java/org/jenkinsci/plugins/github/internal/GitHubLoginFunction.java
bitwiseman: @lanwen 
The guidance is to have a single client and then create new clients from that client using `newBuilder()`.

src/main/java/org/jenkinsci/plugins/github/internal/GitHubClientCacheOps.java
bitwiseman: @lanwen 
Spotbugs continues to have issues with properly identifying the the null check automatically generated when using [try with resources](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) as something that cannot be fixed by the person writing the code.  It reports a problem with [line 109](https://github.com/jenkinsci/github-plugin/pull/228/files#diff-92e6819e3ec96cf3b77dec0a1aece0deR109) which isn't even changed in this PR.

pom.xml
lanwen: lol, sorry, wrong look :)

pom.xml
```diff
@@ -47,10 +47,11 @@
...
     <properties>
-        <jenkins.version>2.60.3</jenkins.version>
+        <jenkins.version>2.164.3</jenkins.version>
         <release.skipTests>false</release.skipTests>
         <maven.javadoc.skip>true</maven.javadoc.skip>
         <findbugs-maven-plugin.version>3.0.4</findbugs-maven-plugin.version>
+        <hamcrest.version>2.2</hamcrest.version>
         <concurrency>1</concurrency>
         <java.level>8</java.level>
         <workflow.version>1.14.2</workflow.version>
```
oleg-nenashev, line 57: Would be great to switch to https://github.com/jenkinsci/bom in a later pull request. This version of Pipeline is ancient, and I here we are not testing with versions being really used with the target Jenkins core

pom.xml
```diff
@@ -162,26 +150,31 @@
...
+            <version>${hamcrest.version}</version>
             <scope>test</scope>
         </dependency>
-
+      
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
-            <version>4.12</version>
+            <version>4.13</version>
```
oleg-nenashev, line 177: Parent POM provides it

```suggestion
```

pom.xml
```diff
@@ -306,6 +299,16 @@
 
     </dependencies>
 
+    <dependencyManagement>
```
oleg-nenashev, line 302: Why is it needed? BOM supplies it
https://github.com/jenkinsci/jenkins/blob/master/bom/pom.xml#L170-L174

bitwiseman: Not using bom.

pom.xml
bitwiseman: Agree bom is a good move, but that is a separate change.

KostyaSha:
probably, ok. Is it done?

bitwiseman:
@KostyaSha 
Yes this is done.

KostyaSha:
let's try

PR #229, opened 2020-04-29
[JENKINS-62097] GitHubPluginConfig.hookSecretConfig → hookSecretConfigs

jglick:
Amends #134. [JENKINS-62097](https://issues.jenkins-ci.org/browse/JENKINS-62097)

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/229)

JENKINS-62097

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/github/config/GitHubPluginConfig.java
+++ b/src/main/java/org/jenkinsci/plugins/github/config/GitHubPluginConfig.java
@@ -5,6 +5,7 @@ import com.google.common.base.Function;
 import com.google.common.base.Predicate;
 import com.google.common.base.Predicates;
 import hudson.Extension;
+import hudson.Util;
 import hudson.XmlFile;
 import hudson.model.Descriptor;
 import hudson.model.Item;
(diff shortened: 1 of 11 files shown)


src/main/java/org/jenkinsci/plugins/github/config/GitHubPluginConfig.java
```diff
@@ -270,13 +284,29 @@ private static void validateConfig(boolean state, String message) {
...
 
+    @Deprecated
     public HookSecretConfig getHookSecretConfig() {
-        return hookSecretConfig;
+        return hookSecretConfigs != null && !hookSecretConfigs.isEmpty()
+                ? hookSecretConfigs.get(0)
+                : new HookSecretConfig(null);
     }
 
-    @DataBoundSetter
```
KostyaSha, line 277: will it break existing caas configs?

jglick: Not sure specifically how CasC works here. In general it does not guarantee backward compatibility even when settings XML is compatible. Out of scope for the design.

KostyaSha: it would be bad to break casc users, such change would require some version and compat hpi marker...

jglick: `compatibleSinceVersion` shows up in the GUI, not useful for CasC users.

In general when updating core/plugins with CasC you may need to make adjustments to YAML to match. AFAIK you get a straightforward error message otherwise so it is pretty obvious.

PR #232, opened 2020-05-11
Fix typo on docs url

timja:
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/232)


oleg-nenashev:
Would be great to get a release, sorry for the mistake

PR #236, opened 2020-07-21
Add Webhook register problem to Troubleshooting

bitwiseman:
Jenkins is adding categories to MagagementLink to group item.
This change adds the webhook problem item to the "Troubleshooting" category.

see https://github.com/jenkinsci/jenkins/pull/4546
see https://github.com/jenkinsci/script-security-plugin/pull/302

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/github/admin/GitHubHookRegisterProblemMonitor.java
+++ b/src/main/java/org/jenkinsci/plugins/github/admin/GitHubHookRegisterProblemMonitor.java
@@ -253,5 +253,11 @@ public class GitHubHookRegisterProblemMonitor extends AdministrativeMonitor impl
         public String getDisplayName() {
             return Messages.hooks_problem_administrative_monitor_displayname();
         }
+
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/github/admin/GitHubHookRegisterProblemMonitor.java
```diff
@@ -253,5 +253,11 @@ public String getDescription() {
         public String getDisplayName() {
             return Messages.hooks_problem_administrative_monitor_displayname();
         }
+
+        // TODO: Override `getCategory` instead using `Category.TROUBLESHOOTING` when minimum core version is 2.226+,
```
KostyaSha, line 257: 2.164.3 is the current core

bitwiseman: Yes, that is the point.  This makes this work and build with earlier versions like 2.164.3 but also return a category when run on later versions that support categories.

bitwiseman:
@KostyaSha Can we go ahead and merge this?

PR #237, opened 2020-07-28
Update JCasC version

alecharp:
This is just to update JCasC version to the latest version compatible with the Jenkins Core requirement of GitHub Plugin.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -55,7 +55,7 @@
         <concurrency>1</concurrency>
         <java.level>8</java.level>
         <workflow.version>1.14.2</workflow.version>
-        <configuration-as-code.version>1.32</configuration-as-code.version>
+        <configuration-as-code.version>1.36</configuration-as-code.version>
     </properties>
 
     <repositories>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -201,10 +201,9 @@
         </dependency>
 
         <dependency>
-            <groupId>io.jenkins</groupId>
-            <artifactId>configuration-as-code</artifactId>
+            <groupId>io.jenkins.configuration-as-code</groupId>
+            <artifactId>test-harness</artifactId>
             <version>${configuration-as-code.version}</version>
-            <classifier>tests</classifier>
```
KostyaSha, line 207: why should it be hard dep?

alecharp: I just removed the classifier, this is still a test dependency.

KostyaSha:
Any idea what exactly fails in PR checks?

alecharp:
This is to make sure the plugin is using the latest possible version of CasC to validate that it's still working. Newer version of CasC require a newer version of Jenkins Core.

timja:
this is no longer required, https://github.com/jenkinsci/github-plugin/pull/243 did the update via bom

PR #241, opened 2020-11-06
feat(updating reregisterallhooks method): only register parent, not all children

dgutzmann:
Instead of trying to register a webhook for each child job in a multibranch-pipeline project, only
register the base project.

With many multibranch pipelines and many prs/branches per project, the total number of calls to github (in my case github enterprise) can become very large and cause rate limiting/abuse rate limiting to be reached

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/github/util/JobInfoHelpers.java
+++ b/src/main/java/org/jenkinsci/plugins/github/util/JobInfoHelpers.java
@@ -11,6 +11,7 @@ import hudson.model.Job;
 import hudson.triggers.Trigger;
 import hudson.triggers.TriggerDescriptor;
 import jenkins.model.ParameterizedJobMixIn;
+import jenkins.scm.api.SCMSourceOwner;
 import org.jenkinsci.plugins.github.extension.GHEventsSubscriber;
 
 import javax.annotation.CheckForNull;
@@ -147,11 +148,10 @@ public final class JobInfoHelpers {
      *
      * @return predicate with true on apply if item is not a child of WorkflowMultiBranchProject
      */
-    public static <ITEM extends Item> Predicate<ITEM> isNotChild() {
+    public static <ITEM extends Item> Predicate<ITEM> isNotSCMSourceOwner() {
         return new Predicate<ITEM>() {
             public boolean apply(ITEM item) {
-                return !(item.getParent().getClass().getName().equals("org.jenkinsci.plugins.workflow.multibranch"
-                                                                 + ".WorkflowMultiBranchProject"));
+                return !(item.getParent() instanceof SCMSourceOwner);
             }
         };
     }
--- a/src/main/java/org/jenkinsci/plugins/github/util/JobInfoHelpers.java
+++ b/src/main/java/org/jenkinsci/plugins/github/util/JobInfoHelpers.java
@@ -141,5 +141,19 @@ public final class JobInfoHelpers {
             }
         };
     }
+
+    /**
+     * Can be useful to ignore child jobs on reregistering hooks
+     *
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/github/util/JobInfoHelpers.java
```diff
@@ -141,5 +141,19 @@ protected Job asJob() {
...
+
+    /**
+     * Can be useful to ignore disabled jobs on reregistering hooks
+     *
+     * @return predicate with true on apply if item is buildable
+     */
+    public static <ITEM extends Item> Predicate<ITEM> isParentMultiBranch() {
+        return new Predicate<ITEM>() {
+            public boolean apply(ITEM item) {
+                return (item.getParent().getClass().getName() != "org.jenkinsci.plugins.workflow.multibranch"
```
atsharp, line 153: I think you'll need to use `.equals` for string comparison:
```
!"org.jenkinsci.plugins.workflow.multibranch.WorkflowMultiBranchProject".equals(item.getParent().getClass().getName());
```

atsharp (approved):
LGTM

src/main/java/org/jenkinsci/plugins/github/util/JobInfoHelpers.java
```diff
@@ -141,5 +141,19 @@ protected Job asJob() {
...
+
+    /**
+     * Can be useful to ignore child jobs on reregistering hooks
+     *
+     * @return predicate with true on apply if item is not a child of WorkflowMultiBranchProject
+     */
+    public static <ITEM extends Item> Predicate<ITEM> isNotChild() {
+        return new Predicate<ITEM>() {
+            public boolean apply(ITEM item) {
+                return !(item.getParent().getClass().getName().equals("org.jenkinsci.plugins.workflow.multibranch"
```
jglick, line 153: `SCMSourceOwner`?

jglick (commented):
No particular opinion; did not realize this plugin even registered hooks to begin with. `github-branch-source` also does so, though this style should probably be considered deprecated: better to use an App which automatically receives webhooks for anything it is installed on.

dgutzmann:
> No particular opinion; did not realize this plugin even registered hooks to begin with. `github-branch-source` also does so, though this style should probably be considered deprecated: better to use an App which automatically receives webhooks for anything it is installed on.

I haven't really used the github-branch-source approach via the github app yet, possibly a better solution in general for auto scanning repos and such.

I was having other issues and I followed this [guide](https://support.cloudbees.com/hc/en-us/articles/224621648-GitHub-Webhook-Troubleshooting) and that is when I came across the 

> Re-register hooks for all jobs

 button and saw that it had an issue with the child jobs...

src/main/java/org/jenkinsci/plugins/github/util/JobInfoHelpers.java
dgutzmann: Sorry, new to all the jenkins code.  Can you please elaborate on your statement?

src/main/java/org/jenkinsci/plugins/github/util/JobInfoHelpers.java
```diff
@@ -141,5 +141,19 @@ protected Job asJob() {
...
+                                                                 + ".WorkflowMultiBranchProject"));
```
jglick, line 154: maybe

```suggestion
                return !(item.getParent() instanceof SCMSourceOwner);
```

KostyaSha: stock multibranch may use jobs for triggering and registration

jglick:
FYI https://github.com/jenkinsci/github-branch-source-plugin/pull/289, still on my back burner.

PR #242, opened 2020-12-02
[JENKINS-50154] Fix webhook payload signature generation when Unicode symbols are present in it

ababushk:
This fixes a bug when plugin rejects webhook payload came from GitHub due to incorrect signature although secret token is configured correctly in Jenkins and GitHub repo/org settings.
It turned out that signature was generated incorrectly when Unicode characters were used in repo's description (in our case it was the ™ character). 

I have a concern of using already deprecated method in new code, but the alternative is to add new dependency to pom.xml (apache-commons.text)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -81,7 +81,11 @@
             <artifactId>commons-lang3</artifactId>
             <version>3.9</version>    
         </dependency>
-
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-text</artifactId>
+            <version>1.7</version>
+        </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>github-api</artifactId>
--- a/src/main/java/org/jenkinsci/plugins/github/webhook/GHWebhookSignature.java
+++ b/src/main/java/org/jenkinsci/plugins/github/webhook/GHWebhookSignature.java
@@ -2,6 +2,7 @@ package org.jenkinsci.plugins.github.webhook;
 
 import hudson.util.Secret;
 import org.apache.commons.codec.binary.Hex;
+import org.apache.commons.lang3.StringEscapeUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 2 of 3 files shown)


ababushk:
Hmm, all those CheckStyle errors are TODO comments, should I fix that?

limitedAtonement:
I don't know why I was notified about this pull request. I don't know the Jenkins code base and I'm no expert in Java unicode handling (though `unescapedPayload.getBytes("latin1")` makes me a wee bit nervous in an ignorant, fear, uncertainty sort of way). I think the only thing I can add is: seven commits is unnecessary. One commit should be fine, or if you really want to break it up, two: one for the test and one for the feature. If you want to chronicle your development saga, do so in commit messages.

Thanks for the fix and for making Jenkins better!

limitedAtonement:
Oh yes, I also meant to say, as an outsider, I don't see adding a dependency to apache-commons.text as a liability, especially if done for the sake of avoiding deprecated function calls.

ababushk:
@limitedAtonement I'm sorry you've received unwanted notification, but thank you for the review! I'm completely agree about number of commits, I've just got used to the way it is configured in my team's repo - all commits in PR are squashed before merge. I also agree about new dependency - I think it is better on the long run.

ababushk:
@KostyaSha could you please take a look?

jonathonbattista:
Any ETA on this? 

Webhooks are completely unusable for us.

ababushk:
@jonathonbattista as a workaround you can build a plugin from my fork https://github.com/ababushk/github-plugin/tree/JENKINS_50154_unicode_payload . You need Java and Maven installed on your system. After you clone the repo, execute `mvn hpi:hpi` - this will build a *.hpi file which you can install in Jenkins' Update Center's advanced options.

timja:
Checkstyle errors will be fixed in https://github.com/jenkinsci/github-plugin/pull/243

timja (approved):

oleg-nenashev (approved):
Looks good!

KostyaSha:
i think we can try it

eyalzek:
Jobs stopped triggering for us on merges/new PR after updating to v1.33.0. Can this be the reason? We had to downgrade to 1.32.0

enys:
It looks like I have the same problem: 
```
<title>Error 400 Provided signature [c0180c723bbe3ddb266be493f5c9c1977896ea15] did not match to calculated</title>
</head>

```
I have commented  the Jira issue.

eagletmt:
Same here. I also had to downgrade to 1.32.0 due to the signature mismatch errors.

This PR looks completely broken to me.

1. `new String(unescapedPayload.getBytes("latin1"), UTF_8)` replaces all non-latin1 characters to `?`
    - So a valid JSON payload (e.g., `{"description": "説明"}` or `{"description": "\u8aac\u660e"}`) is converted to broken one
2.  `"\u00e2\u0084\u00a2"` is not a JSON string for ™ character
    - Its codepoint is U+2122. Thus the correct JSON string should be `"\u2122"`. [e2, 84, a2] is a UTF-8 byte sequence of ™.

ababushk:
@KostyaSha could we revert this while I'm searching for the fix?

KostyaSha:
@ababushk could you PR revert? i can merge release then

ababushk:
I've created a PR to revert this https://github.com/jenkinsci/github-plugin/pull/246
I apologize to everyone who suffered from this change - looks like the bug is not in the plugin.
I've tried to setup webhooks on Jenkins instance behind corporate firewall using this instruction https://www.jenkins.io/blog/2019/01/07/webhook-firewalls/, but I've taken pysmee (https://github.com/Akrog/pysmee) instead of official smee.io client written in JavaScript. If the bug exists, it is in pysmee - official client passes webhooks correctly.

KostyaSha:
merged revert and released 1.33.1

PR #243, opened 2020-12-31
Cleanup pom, add bom

timja:
@KostyaSha

if this is merged then https://github.com/jenkinsci/github-plugin/pull/237 can be closed

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>4.10</version>
+        <version>4.15</version>
         <relativePath />
     </parent>
 
@@ -58,7 +58,10 @@
         <concurrency>1</concurrency>
         <java.level>8</java.level>
         <workflow.version>1.14.2</workflow.version>
-        <configuration-as-code.version>1.32</configuration-as-code.version>
+        <tagNameFormat>v@{project.version}</tagNameFormat>
+        <spotbugs.threshold>Low</spotbugs.threshold>
+        <spotbugs.effort>Max</spotbugs.effort>
+        <spotbugs.failOnError>false</spotbugs.failOnError>
     </properties>
 
     <repositories>
@@ -79,7 +82,7 @@
         <dependency>
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-lang3</artifactId>
-            <version>3.9</version>    
+            <version>3.11</version>
         </dependency>
 
         <dependency>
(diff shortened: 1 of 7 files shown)


pom.xml
```diff
@@ -50,15 +50,18 @@
         <revision>1.33.0</revision>
         <changelist>-SNAPSHOT</changelist>
         <gitHubRepo>jenkinsci/github-plugin</gitHubRepo>
-        <jenkins.version>2.164.3</jenkins.version>
+        <jenkins.version>2.222.4</jenkins.version>
```
timja, line 53: 97% of users are on this version (https://stats.jenkins.io/pluginversions/github.html), and it's the minimum supported for new versions of bom currently.

also see https://www.jenkins.io/doc/developer/plugin-development/choosing-jenkins-baseline/

src/test/java/org/jenkinsci/plugins/github/status/sources/ManuallyEnteredRepositorySourceTest.java
```diff
@@ -33,12 +32,11 @@
 
     @Test
     public void nullName() {
-        ManuallyEnteredRepositorySource instance = Mockito.spy(new ManuallyEnteredRepositorySource("https://github.com/jenkinsci/jenkins"));
-        doReturn(null).when(instance).createName(Matchers.anyString());
+        ManuallyEnteredRepositorySource instance = spy(new ManuallyEnteredRepositorySource("a"));
```
timja, line 35: some weird mockito issue that only occurred when running all tests via CLI, couldn't reproduce in IDE or `mvn test`.

the test is checking nullName, so faking an impossible case didn't make sense anymore so I removed the mocking and the test works

src/test/resources/checkstyle/checkstyle-config.xml
```diff
@@ -185,10 +185,6 @@
         <!-- Miscellaneous other checks.                   -->
         <!-- See http://checkstyle.sf.net/config_misc.html -->
         <module name="ArrayTypeStyle"/>
-        <module name="TodoComment">
```
timja, line 188: this is what is causing other PRs to fail CI currently

src/test/resources/checkstyle/checkstyle-config.xml
```diff
@@ -32,6 +32,7 @@
 
 
 <module name="Checker">
+    <property name="cacheFile" value="${checkstyle.cache.file}"/>
```
timja, line 35: most changes in this file are just moving them around due to checkstyle update

oleg-nenashev (approved):

PR #244, opened 2020-12-31
JENKINS-64536 Remove tabular form markup on system configuration page

timja:
Pretty similar fix to https://github.com/jenkinsci/matrix-auth-plugin/commit/31b3bbc14062974f240a26cf88ae04e6c9415571

See https://www.jenkins.io/doc/developer/views/table-to-div-migration/ and https://www.jenkins.io/blog/2020/11/10/major-changes-in-weekly-releases/

cc @KostyaSha 

requires https://github.com/jenkinsci/github-plugin/pull/243

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/github/config/GitHubPluginConfig/config.groovy
+++ b/src/main/resources/org/jenkinsci/plugins/github/config/GitHubPluginConfig/config.groovy
@@ -4,6 +4,7 @@ import com.cloudbees.jenkins.GitHubPushTrigger
 import lib.FormTagLib
 
 def f = namespace(FormTagLib);
+def g = namespace("/lib/matrixauth")
 
 f.section(title: descriptor.displayName) {
     f.entry(title: _("GitHub Servers"),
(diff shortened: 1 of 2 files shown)


src/main/resources/org/jenkinsci/plugins/github/config/GitHubPluginConfig/config.groovy
```diff
@@ -4,6 +4,7 @@ import com.cloudbees.jenkins.GitHubPushTrigger
 import lib.FormTagLib
 
 def f = namespace(FormTagLib);
+def g = namespace("/lib/matrixauth")
```
KostyaSha, line 7: why matrixauth?

oleg-nenashev:
```suggestion
def g = namespace("/lib/github")
```

oleg-nenashev (commented):
copy-paste issue, but the rest is good

src/main/resources/org/jenkinsci/plugins/github/config/GitHubPluginConfig/config.groovy
timja: oops, i assume it worked because matrix auth is normally installed

PR #247, opened 2021-03-15
Minor document  update with clear images

srbala:
Document contains images not visible to read so, added new images. On step 3, check label change over the time, fix the label

- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your master branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [x] Link to relevant issues in GitHub or Jira - Plugin document site https://plugins.jenkins.io/github/
- [x] Link to relevant pull requests, esp. upstream and downstream changes - N/A
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue - N/A

- - -
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/247)


README.md
```diff
@@ -85,7 +85,8 @@ Login and Password to token**
 > credentials.
```
jimklimov, line 86: By trial and error I found that the "Secret text" token, at least for a public github.com account, should contain just the token value ("ghp_Vari0usChars"); the "Test Connection" button helps see if that works. It seems that github knows the user name from that token alone.

I did not uncover what "in corresponding domain" here means in practice, and "Credentials are filtered on changing custom GitHub URL" in help text. My credential ID and description deliberately did not mention neither the account name nor github.com, and it still worked for the default https://api.github.com URL.

This would be nice to clarify in the doc and help, while we are at it.

jimklimov (approved):
Useful update of the docs, especially now that login/pass based stuff does not work :)

README.md
srbala: I adding some some text and screen shorts would be nice, I think.

oleg-nenashev (approved):
Would be nice to retain the original image file names. Anyway, it is a nice improvement, thanks!

srbala:
> Would be nice to retain the original image file names. Anyway, it is a nice improvement, thanks!

@oleg-nenashev Can I replace the old files? OR old name with suffix?

PR #249, opened 2021-05-09
Remove usages of `com.google.common.base.Objects#firstNonNull`

basil:
`com.google.common.base.Objects#firstNonNull` has been removed in recent versions of Guava, so consuming this API is a liability. This PR replaces usages of `com.google.common.base.Objects#firstNonNull` with native Java functionality to allow this plugin to be used seamlessly with any version of Guava in the future.


KostyaSha:
could you make it in Utils as static helper?

basil:
This patch, as-is, is all I can offer at the moment. Feel free to modify it if you would like or to close it otherwise.

PR #258, opened 2021-10-13
'git@' is not static, it can be e.g., 'org-12345@'

ppaul-square:
I had previously thought that 'git@' was a hard-coded part of git, but that is not the case.  This is a valid repo url:
image

After putting FINEST log levels on this class, it was clear that it was failing on the 'git@' requirement.  This change was built with hpi:hpi, given a fake version of 1.999 and uploaded manually into Jenkins.  Once in place, everything worked as expected.
Perhaps a better regex would be required, or perhaps this assumption exists in other places, but I did not hit the problem elsewhere so did not look.

- [x ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [ x] Ensure that the pull request title represents the desired changelog entry
- [ x] Please describe what you did
- [ x] Link to relevant issues in GitHub or Jira
- [ x] Link to relevant pull requests, esp. upstream and downstream changes
- [ x] Ensure you have provided tests - that demonstrates feature works or fixes the issue


KostyaSha:
Could you add a test case https://github.com/jenkinsci/github-plugin/blob/master/src/test/java/com/coravy/hudson/plugins/github/GitHubRepositoryNameTest.java ?

ppaul-square:
Awaiting approval.

KostyaSha:
re-running ci

PR #262, opened 2022-02-09
Apply routine updates to fix PCT with more recent Jenkins versions and JDK11

Vlatombe:
* Update jenkins.version to 2.277.1
* Update matching bom
* Update maven wrapper to a current version
* Update wiremock as the previous version had guava incompatibilities with new core

Overlaps with #261 so maintainer will need to pick one of these (this PR being more conservative using an older Jenkins baseline, but doesn't cover the guava upgrade)

- [X] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [X] Ensure that the pull request title represents the desired changelog entry
- [X] Please describe what you did

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/github/test/GHMockRule.java
+++ b/src/test/java/org/jenkinsci/plugins/github/test/GHMockRule.java
@@ -12,7 +12,6 @@ import org.junit.runners.model.Statement;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
-import wiremock.org.eclipse.jetty.http.HttpStatus;
 
 import static com.cloudbees.jenkins.GitHubWebHookFullTest.classpath;
 import static com.github.tomakehurst.wiremock.client.WireMock.aResponse;
@@ -21,6 +20,7 @@ import static com.github.tomakehurst.wiremock.client.WireMock.post;
 import static com.github.tomakehurst.wiremock.client.WireMock.urlPathEqualTo;
 import static com.github.tomakehurst.wiremock.client.WireMock.urlPathMatching;
 import static java.lang.String.format;
+import static java.net.HttpURLConnection.HTTP_CREATED;
 
 /**
  * Mocks GitHub on localhost with some predefined methods
@@ -133,7 +133,7 @@ public class GHMockRule implements TestRule {
                 service().stubFor(
                         post(urlPathMatching(
                                 format("/repos/%s/%s/statuses/.*", REPO.getUserName(), REPO.getRepositoryName()))
-                        ).willReturn(aResponse().withStatus(HttpStatus.CREATED_201)));
+                        ).willReturn(aResponse().withStatus(HTTP_CREATED)));
             }
         });
     }


aHenryJard (approved):

kerogers-cloudbees (approved):

src/test/java/org/jenkinsci/plugins/github/config/GitHubServerConfigTest.java
```diff
@@ -74,7 +75,7 @@ public void shouldGuessNameIfNotProvided() throws Exception {
         GitHubServerConfig input = new GitHubServerConfig("");
         input.setApiUrl(CUSTOM_GH_SERVER);
         assertThat(input.getName(), is(nullValue()));
-        assertThat(input.getDisplayName(), is("some (http://some.com)"));
```
KostyaSha, line 77: what's the reason for this change?

twasyl: I can't tell as it's working as well. @Vlatombe any reason behind this change?

Vlatombe: Looks like the test is failing with `mvn test -Dtest=GitHubServerConfigTest -Djenkins.version=2.332`. Looking up why but it looks related to guava upgrade.

Vlatombe: There is a swallowed error when running against jenkins 2.332

```
java.lang.NoSuchFieldError: EXACT
	at com.google.common.net.InternetDomainName.findSuffixOfType(InternetDomainName.java:171)
	at com.google.common.net.InternetDomainName.<init>(InternetDomainName.java:152)
	at com.google.common.net.InternetDomainName.from(InternetDomainName.java:210)
	at jenkins.scm.api.SCMName.resolveInternetDomainName(SCMName.java:143)
	at jenkins.scm.api.SCMName.innerFromUrl(SCMName.java:113)
	at jenkins.scm.api.SCMName.fromUrl(SCMName.java:67)
	at org.jenkinsci.plugins.github.config.GitHubServerConfigTest.given__url_with(GitHubServerConfigTest.java:83)
[...]
```

This is caused by the `wiremock:standalone` dependency which is vendoring `com.google.thirdparty.publicsuffix.PublicSuffixPatterns`

Vlatombe: The proper fix would be to remove the wiremock `standalone` classifier once the Jenkins baseline is raised to 2.332 or later. Until then, this test change seems to be the only way to run tests with both baselines (for PCT runs purpose against recent Jenkins baselines)

jtnord: You should always use the wiremock standalone due to the mass of library conflicts- however the version used is very old and may not be correctly packaged.

should use 
```
<dependency>
  <groupId>com.github.tomakehurst</groupId>
  <artifactId>wiremock-jre8-standalone</artifactId>
  <version>xxx</version>
</dependency>
```

Vlatombe: I upgraded wiremock, looks good.

src/test/java/org/jenkinsci/plugins/github/test/GHMockRule.java
```diff
@@ -13,6 +12,7 @@
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
+import wiremock.org.eclipse.jetty.http.HttpStatus;
```
jtnord, line 15: NIT: generally I would not rely on on internal implementations, but use `java.net.HttpURLConnection.HTTP_CREATED`, but this was existing anyway.

jtnord (approved):

KostyaSha:
hope squash merge is not a problem for everyone

kerogers-cloudbees:
@KostyaSha Can you release this plugin? Or do you know who can so I can ask them?

KostyaSha:
> @KostyaSha Can you release this plugin?

done

PR #266, opened 2022-03-09
Update plugin parent POM and BOM

basil:
Picking up where #261 left off.


basil:
@KostyaSha Can you please take a look?

PR #299, opened 2022-09-05
Log missing admin access on INFO level

kutzi:
I don't know what the motivation was to log this on DEBUG only (less verbose logs?), but IMO you're missing quite important info this way in the default log level INFO.
I.e. when a new hook should be created, you only see in the logs:
`INFO	o.j.p.g.webhook.WebhookManager$1#run: GitHub webhooks activated for job X with [GitHubRepositoryName[...]] (events: [PUSH])`
which *looks* like it's successful already ('webhooks activated')
but you miss the DEBUG message which says that you have a correct repo configured, but missing the admin rights
`


kutzi:
In fact - just ran into the same issue again - the logs on INFO level for successful and unsuccessful hook generation look exactly the same!
Only the `GitHub webhooks activated for job` log

KostyaSha:
The problem of info level is that on setups with thousands of repos it will spam. I usually use info only for not really repeating errors. For debug purpose admin can add extra logger (afair this is described in docs).

kutzi:
Is one line of log more in the case of an error really spam?
It would only log if the other line is logged as well, right?

PR #304, opened 2022-10-18
Register XStream alias even earlier

rsandell:
There has been a change in the git plugin that somehow makes `GitHubPluginConfig` load earlier than `GitHubPlugin.start()`
So when running the `MigratorTest` with git plugin 4.12.1 installed (for example during PCT) it fails because the old configuration could not be loaded.

The fix is simply to register the XStream alias before `InitMilestone.SYSTEM_CONFIG_LOADED`  instead of `GitHubPlugin.start()`.

I've verified that it works by temporarily adding the git plugin as a test dependency in the pom of this plugin, and the `MigratorTest` fails before this change and succeeds after the fix got applied.

- [X] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [X] Ensure that the pull request title represents the desired changelog entry
- [X] Please describe what you did
- [X] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/github/GitHubPlugin.java
+++ b/src/main/java/org/jenkinsci/plugins/github/GitHubPlugin.java
@@ -43,10 +43,6 @@ public class GitHubPlugin extends Plugin {
         new Migrator().migrate();
     }
 
-    @Override
-    public void start() throws Exception {
-    }
-
     /**
      * Shortcut method for getting instance of {@link GitHubPluginConfig}.
      *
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/github/GitHubPlugin.java
```diff
@@ -41,7 +45,6 @@ public static void runMigrator() throws Exception {
 
     @Override
     public void start() throws Exception {
-        addXStreamAliases();
```
jtnord, line 44: shouldn't you remove this method completely now?

rsandell: I could

car-roll (approved):

Dohbedoh:
Apparently this causes some initialization issue https://issues.jenkins.io/browse/JENKINS-70351.

PR #328, opened 2023-02-14
[JENKINS-69353] github-plugin test failure on Java 17

basil:
Serializing mock triggers is no longer supported in Java 17, so instead switch to a pure mock-based approach that does not require serialization. Tested by successfully running `mvn clean verify -Dtest=org.jenkinsci.plugins.github.webhook.subscriber.DefaultPushGHEventListenerTest` on both Java 11 and 17.

CC @KostyaSha

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,4 +1,4 @@
 buildPlugin(useContainerAgent: true, configurations: [
-                [platform: 'linux', jdk: '11'],
-                [platform: 'windows', jdk: '11'],
-            ])
+    [platform: 'linux', jdk: 17],
+    [platform: 'windows', jdk: 11],
+])
--- a/pom.xml
+++ b/pom.xml
@@ -129,7 +129,7 @@
 
         <dependency>
             <groupId>org.mockito</groupId>
-            <artifactId>mockito-core</artifactId>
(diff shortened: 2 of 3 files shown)


basil:
Used my admin powers to replay the build with the `Jenkinsfile` changes here: https://ci.jenkins.io/job/Plugins/job/github-plugin/job/PR-328/4/

KostyaSha:
> Serializing mock triggers is no longer supported in Java 17,

why so?

basil:
See [JENKINS-69353](https://issues.jenkins.io/browse/JENKINS-69353). Mockito-generated classes contain fields from the `java.util.concurrent` package, which is not open in https://github.com/jenkinsci/jenkins/blob/62c2fbcc7af34a2bc4edd67e46939fbe90f3b899/war/pom.xml#L228 (and therefore unavailable to XStream when running on Java 17). These tests were trying to serialize a job with a mock-based trigger with XStream. This PR solves the problem by avoiding XStream and using mocks consistently.

basil:
Thanks! A release of this would allow us to get rid of [this hack](https://github.com/jenkinsci/bom/blob/c72362731c98313189c04025c3b7bc181da4c121/Jenkinsfile#L53-L55) in `jenkinsci/bom`.

PR #347, opened 2023-05-02
[JENKINS-60738] Fix global configuration submission from UI

Dohbedoh:
Fixes [JENKINS-60738](https://issues.jenkins.io/browse/JENKINS-60738) caused by changes introduced in https://github.com/jenkinsci/github-plugin/pull/221.

Databinding of `hookUrl` is not working properly, per my understanding because the `DataboundSetter` [setHookUrl(String hookUrl)](https://github.com/jenkinsci/github-plugin/blob/v1.37.0/src/main/java/org/jenkinsci/plugins/github/config/GitHubPluginConfig.java#L117-L118) accepts a `String` but the field is a `URL`. While adding a `public void setHookUrl(URL hookUrl)` might solve some of the problem, then there is yet another issue that disabling the [Specify another hook URL for GitHub configuration](https://github.com/jenkinsci/github-plugin/blob/v1.37.0/src/main/resources/org/jenkinsci/plugins/github/config/GitHubPluginConfig/config.groovy#L29) checkbox does not remove the hook URL at all... that seems wrong.

So the propose fix:

* handle the binding and the optional property state in the `configure(StaplerRequest req, JSONObject json)`
* reactivate and adjusted the global configuration UI tests

- [ ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [ ] Ensure that the pull request title represents the desired changelog entry
- [ ] Please describe what you did


julieheard (approved):
LGTM, manually tested and is working 🙂

jglick:
While this sounds fine as a hotfix, the proper fix would be to delete all the overrides and special cases handling and fix the form to use normal databinding by convention.

src/main/resources/org/jenkinsci/plugins/github/config/GitHubPluginConfig/config.groovy
```diff
@@ -27,10 +27,11 @@ f.section(title: descriptor.displayName) {
...
                     f.optionalBlock(title: _("Specify another hook URL for GitHub configuration"),
+                            name: "isOverrideHookUrl",
                             inline: true,
                             checked: instance.isOverrideHookUrl()) {
                         f.entry(field: "hookUrl") {
-                            f.textbox(checkMethod: "post")
+                            f.textbox(checkMethod: "post", name: "hookUrl")
                         }
                     }
                 }
```
jglick, line 37: For example all of this should just be like

```xml
<f:entry field="hookUrl">
  <f:textbox label="Alternate hook URL for GitHub configuration"/>
</f:entry>
```

which may be blank.

src/main/java/org/jenkinsci/plugins/github/config/GitHubPluginConfig.java
```diff
@@ -190,17 +191,35 @@ protected XmlFile getConfigFile() {
...
+                } else {
+                    setHookSecretConfigs(Collections.emptyList());
+                }
+                if (json.optBoolean("isOverrideHookUrl", false) && (json.has("hookUrl"))) {
+                    setHookUrl(json.optString("hookUrl"));
+                } else {
+                    setHookUrl(null);
+                }
+                req.bindJSON(this, json);
+                clearRedundantCaches(configs);
```
jglick, line 213: Other than this method call, the whole `configure` method should not even need to be overridden.

src/main/resources/org/jenkinsci/plugins/github/config/GitHubPluginConfig/config.groovy
Dohbedoh: Oh I see. Remove all the optional block that is not necessary.. Yeah sounds good to me I can test this.

jglick: Would be more work for sure. The usual problem with this sort of fix is retaining compatibility for both XML settings and JCasC. Sometimes hacking up the JSON processing in the GUI logic is the only option since refactoring to a simpler structure would break something.

Dohbedoh: the thing that bother me is the type `URL` of the `hookUrl`... Because of that type, the value cannot actually be blank :( :

```
java.net.MalformedURLException: no protocol: 
	at java.base/java.net.URL.<init>(URL.java:645)
	at java.base/java.net.URL.<init>(URL.java:541)
	at java.base/java.net.URL.<init>(URL.java:488)
	at org.kohsuke.stapler.Stapler$3.convert(Stapler.java:1141)
Caused: org.apache.commons.beanutils.ConversionException: no protocol: 
```

So it is not that simple to "reset" the hook URL because of the binding.. I thought about changing the type from `URL` to `String` but the existing public getter might be a problem: https://github.com/jenkinsci/github-plugin/blob/master/src/main/java/org/jenkinsci/plugins/github/config/GitHubPluginConfig.java#L131-L141

Changing its signature could be a breaking change. Though it does not seem to be used outside the scope of that plugin repo: https://github.com/search?ref=simplesearch&type=code&q=user%3Ajenkinsci++getHookUrl.

jglick: BTW use permalinks for better display: https://github.com/jenkinsci/github-plugin/blob/47b623aad58baffe8c2d20c0ada1339124229804/src/main/java/org/jenkinsci/plugins/github/config/GitHubPluginConfig.java#L131-L141

jtnord:
@KostyaSha are you ok with this as the active maintainer of this plugin?

src/main/java/org/jenkinsci/plugins/github/config/GitHubPluginConfig.java
Dohbedoh: The only way I found to simplify the data-binding and just do a `req.bindJson` is to change the signature of  `public URL getHookUrl()` to `public String getHookUrl()`. That could be a breaking change. So not sure if that would be acceptable ?  cc @KostyaSha

KostyaSha: better ask @jglick i forgot databinding nuances

jglick: > That could be a breaking change. … So not sure if that would be acceptable ?

https://github.com/search?q=org%3Ajenkinsci%20getHookUrl&type=code does not look like it is used from other plugins.

yahooguntu:
I have the same issue; as a workaround I've been adding `<hookUrl>` directly to `github-plugin-configuration.xml`, but unfortunately it's cleared out when I save/apply System settings in the web UI.

franknarf8:
Can this be merged soon? I've just noticed this open PR after implementing the same fix and opening another PR : https://github.com/jenkinsci/github-plugin/pull/375

Feel free to close the other PR

@KostyaSha @jglick

jglick:
(I am not a maintainer)

Dohbedoh:
@jenkinsci/github-plugin-developers anybody able to merge this ?

jglick:
@KostyaSha perhaps

Dohbedoh:
@KostyaSha @lanwen @oleg-nenashev anybody able to help merge this ?

sebastianreloaded:
> > I have the same issue; as a workaround I've been adding `<hookUrl>` directly to `github-plugin-configuration.xml`, but unfortunately it's cleared out when I save/apply System settings in the web UI.
> 
> If you edit the config file and restart so it has the updated value in memory then the UI should show the correct setting and not clobber it on save.

But even when adding the hookUrl to the xml and restarting, the endpoint still only works for "/github-webhook/". everything else gives a "HTTP ERROR 403 No valid crumb was included in the request"

jgarciacloudbees:
Hi @jenkinsci/github-plugin-developers, is there any chance to merge this PR?

olamy:
@KostyaSha thanks for merging. Could you please make a release now? Thanks

olamy:
> > Hi @jenkinsci/github-plugin-developers, is there any chance to merge this PR?
> 
> the group contains ~4~ 3 users (and a bot) who are not so active, I would suggest you start the process to adopt this plugin.

looks like a good idea.

@KostyaSha let us know if you are too busy to maintain the plugin actively, I'm pretty sure some people will be happy to help to maintain such critical plugin.

PR #368, opened 2023-10-10
Test with Java 21

MarkEWaite:
## Test with Java 21

Java 21 was released Sep 19, 2023. We want to announce full support for Java 21 in early October and would like the most used plugins to be compiled and tested with Java 21.

The acceptance test harness and plugin bill of materials tests are already passing with Java 21. This is a further step to improve plugin readiness for use with Java 21 and for development with Java 21.

The change intentionally tests only two Java configurations, Java 17 and Java 21 because we believe that the risk of a regression that only affects Java 11 is low. We generate Java 11 byte code with the Java 17 and the Java 21 builds, so we're already testing Java 11 byte code.

### Testing done

Confirmed that tests pass with Java 21.

### Submitter checklist
* [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
* [x] Ensure that the pull request title represents the desired changelog entry
* [x] Please describe what you did

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,5 +1,4 @@
 buildPlugin(useContainerAgent: true, configurations: [
     [platform: 'linux', jdk: 21],
     [platform: 'windows', jdk: 17],
-    [platform: 'linux', jdk: 11],
 ])
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,4 +1,5 @@
 buildPlugin(useContainerAgent: true, configurations: [
     [platform: 'linux', jdk: 21],
     [platform: 'windows', jdk: 17],
+    [platform: 'linux', jdk: 11],
 ])
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,4 +1,4 @@
 buildPlugin(useContainerAgent: true, configurations: [
-    [platform: 'linux', jdk: 17],
-    [platform: 'windows', jdk: 11],
+    [platform: 'linux', jdk: 21],
+    [platform: 'windows', jdk: 17],
 ])


MarkEWaite:
@integer , @lanwen , or @oleg-nenashev would you be willing to merge this so that the plugin is regularly tested with Java 21.  No need for a release, merging the pull request is enough.

oleg-nenashev (changes requested):
I am +1 for adding Java 21, but not sure why Java 11 is removed while we keep support for it in the plugin. Testing Java 11 bytecode on a newer JDK version is generally not enough

MarkEWaite:
I've pushed 6d8ef2ff8f3e37f3540c4274d320e624fc03d2d2 to address your concern @oleg-nenashev .  I don't agree with the concern, but it is much more valuable to test Java 21 regularly than to exclude testing of this plugin with Java 11.

We've detected no regressions with the transition to testing Java 11 byte code as generated by Java 17 and Java 21 compilers in the 200+ plugins that have made the transition.

We test with Java 11 on older lines (Jenkins 2.401.x and Jenkins 2.414.x) in the plugin bill of materials while testing with Java 21 on the most recent weekly release.  We've detected no issues with any of the plugins that are included in the plugin BOM.

We want to reduce Jenkins infrastructure costs by not testing configurations that are unlikely to detect issues.  Testing with Java 11 is unlikely to detect issues and adds infrastructure cost.  However, the added infrastructure cost of a few plugins that want to continue testing with Java 11 is much less than the risk of not testing with Java 21.

I've added Java 11 on Linux as a test configuration. I've made the same addition in a few other repositories when the maintainer requested that it be added.  I do not intend to make that same addition in general.  When a plugin maintainer asks to continue testing Java 11 for a specific pull request, I'm willing to adapt that pull request as I've done with this pull request.

oleg-nenashev:
@MarkEWaite I didn't have a chance to review the previous comment but I will try to get to it ASAP. I understand the infrastructure cost implications and I definitely don't want to grow them. If we have proper coverage in BOM builds, it should be okay for integration testing

MarkEWaite:
This plugin is part of the plugin bill of materials, so it is being evaluated in the plugin bill of materials with Java 11.

oleg-nenashev (approved):
Let's align with the default practices for other plugins, at least for now. If there is a need in running more configurations, one could do it on GitHub Actions without putting pressure on the infra

369tux:
https://github.com/jenkinsci/github-plugin/pull/368#pullrequestreview-1726308114

PR #388, opened 2025-02-03
Warn about duplicated events received from GitHub via Admin Monitor

gbhat618:
**Problem**
If duplicated GitHub events get delivered to Jenkins controller, Jenkins need to act on events both times (they are not automatically de-duplicated). These events may be a branch scan, build trigger etc.

<details><summary>further discussion of problem:</summary>

A branch scan event running twice may not be a problem, however build triggering twice could be a problem. Even the builds doesn't get in triggered twice due to `quietPeriod: 5s` (unless the quietPeriod was set to 0, or duplicates come after quietPeriod: 5s, which case two builds do get triggered).

Note: usually GH by itself won't automatically redeliver an event if failed, GH simply marks the event as failed (an event is marked failed if GH didn't receive a 200 OK response within 10s; [reference](https://docs.github.com/en/webhooks/using-webhooks/handling-failed-webhook-deliveries#about-webhook-delivery-failures))

The reason for such duplicated events is,
* mostly, a wrong configuration in GitHub side, such as registering two webhooks (maybe at multiple places such as org, and repo levels) which are targeting the same controller.
* if a custom setup is done to redeliver failed events if the redeliver is going wrong (this should be rare to never)

Although it is typically rare for this duplication issue to happen, if it happens, it could require quite some time to figure out what is going on. And in some case it may simply go unnoticed.

<details><summary>Logs showing build can trigger twice in case of `quietPeriod: 0`</summary>

Note: in the below logs, when the two events came
*  at exactly the same second, Jenkins didn't execute 2 builds, rather both the events corresponded to a single build; except in the build log we saw `Started by GitHub push by gbhat618` twice ([ref](https://github.com/jenkinsci/github-plugin/blob/d2e9e60422ae6f72b6ead117b7ea7a5f8a34470d/src/main/java/com/cloudbees/jenkins/GitHubPushCause.java#L39)); however rest of logs showed build steps were executed only once; I verified by this by having a single executor.
* within 1 second apart, Jenkins executed two builds one after other.

```
2025-02-06 03:20:19.393+0000 [id=87]	INFO	o.j.p.g.w.s.DefaultPushGHEventSubscriber#onEvent: Received PushEvent for https://github.com/gbhat618/dummy from 140.82.115.146:35792 → 140.82.115.146 ⇒ https://localhost:8080/github-webhook/
[...]

PR #389, opened 2025-02-25
Do not use static state variables in Extension class

gbhat618:
Flakiness in the test was identified (possibly) due to the use of static stateful fields in the `@Extension` class. See [this comment](https://github.com/jenkinsci/github-plugin/pull/388#issuecomment-2672968464) for details. Issue was likely parallel test execution mutating the state of static field.

This PR addresses the issue following the [comment here](https://github.com/jenkinsci/github-plugin/pull/388#issuecomment-2679138087).

Note: This problem only affects tests and does not impact Jenkins functionality. Whether the data is stored in a static cache or within a field inside a singleton instance, it effectively centralizes event data in one place.

### Testing done
* Automated tests still executing correctly.
* Manual testing
   * verify the duplicate events monitor functionality is correct as before (tested by sending duplicate events - admin monitor comes, restart jenkins - monitor goes away)

### Submitter checklist
- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [x] Link to relevant issues in GitHub or Jira
- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/github/admin/GitHubDuplicateEventsMonitor.java
+++ b/src/main/java/org/jenkinsci/plugins/github/admin/GitHubDuplicateEventsMonitor.java
@@ -207,7 +207,7 @@ public class GitHubDuplicateEventsMonitor extends AdministrativeMonitor {
          */
         @VisibleForTesting
         @Restricted(NoExternalUse.class)
-        Set<String> getEventCountsTracker() {
+        Set<String> getPresentEventKeys() {
             return eventTracker.asMap().keySet().stream()
                                .filter(key -> eventTracker.getIfPresent(key) != null)
                                .collect(Collectors.toSet());
--- a/src/main/java/org/jenkinsci/plugins/github/admin/GitHubDuplicateEventsMonitor.java
+++ b/src/main/java/org/jenkinsci/plugins/github/admin/GitHubDuplicateEventsMonitor.java
@@ -12,6 +12,7 @@ import com.github.benmanes.caffeine.cache.Ticker;
 import com.google.common.annotations.VisibleForTesting;
 
 import hudson.Extension;
+import hudson.ExtensionList;
 import hudson.model.AdministrativeMonitor;
 import hudson.model.Item;
 import jenkins.model.Jenkins;
@@ -57,7 +58,7 @@ public class GitHubDuplicateEventsMonitor extends AdministrativeMonitor {
 
     @Override
     public boolean isActivated() {
-        return DuplicateEventsSubscriber.isDuplicateEventSeen();
+        return ExtensionList.lookupSingleton(DuplicateEventsSubscriber.class).isDuplicateEventSeen();
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/github/admin/GitHubDuplicateEventsMonitor.java
```diff
@@ -75,7 +76,7 @@ public void checkRequiredPermission() {
     public HttpResponse doGetLastDuplicatePayload() {
         Jenkins.get().checkPermission(Jenkins.SYSTEM_READ);
         JSONObject data;
-        var lastDuplicate = DuplicateEventsSubscriber.getLastDuplicate();
+        var lastDuplicate = ExtensionList.lookupSingleton(DuplicateEventsSubscriber.class).getLastDuplicate();
```
gbhat618, line 79: I'm using ExtensionList.lookupSingleton in two places in this class.

I considered keeping a class-level subscriber field, but that seem to require extra code to manage the reference correctly.
* Since both this monitor class and the subscriber have @Extension, I think they can't store direct references to each other—perhaps because their initialization order isn't guaranteed?
* We could store the reference during execution (`if (subscriber != null) { subscriber = ExtensionList.lookupSingleton(...); }`), but extensions are already looked up in more hot code path - 

https://github.com/jenkinsci/github-plugin/blob/master/src/main/java/org/jenkinsci/plugins/github/extension/GHEventsSubscriber.java#L148-L150

So I think this approach is fine - (`ExtensionList.lookupSingleton` probably doesn't introduce any overhead)

src/main/java/org/jenkinsci/plugins/github/admin/GitHubDuplicateEventsMonitor.java
```diff
@@ -206,10 +207,10 @@ public static TrackedDuplicateEvent getLastDuplicate() {
          */
         @VisibleForTesting
         @Restricted(NoExternalUse.class)
-        static Set<String> getEventCountsTracker() {
-            return EVENT_TRACKER.asMap().keySet().stream()
-                                .filter(key -> EVENT_TRACKER.getIfPresent(key) != null)
-                                .collect(Collectors.toSet());
+        Set<String> getEventCountsTracker() {
+            return eventTracker.asMap().keySet().stream()
+                               .filter(key -> eventTracker.getIfPresent(key) != null)
```
jglick, line 212: BTW I think this would be better written using a stream on `entrySet`.

jglick (approved):

src/main/java/org/jenkinsci/plugins/github/admin/GitHubDuplicateEventsMonitor.java
gbhat618: `entrySet` because it is more performance optimal?

would be like,
```
return eventTracker.asMap().entrySet().stream()
                   .filter(entry -> entry.getValue() != null)
                   .map(Map.Entry::getKey)
                   .collect(Collectors.toSet());
```

But since this is a helper method for tests only, and we only need the keys, ig probably keeping on `keySet()` is easy read.. 🤔 

(I now renamed the method to `getPresentEventKeys` which makes more sense)

jglick: Yeah it is just a bad pattern to iterate a map’s key set and then look up the value each time. Does not really matter of course.

PR #390, opened 2025-03-07
Fix issue with github-plugin-configuration unmarshalling

vwagh-dev:
### Issue:
On restart of OC, github configuration doesn't load due to unmarshalling exception

The error `com.thoughtworks.xstream.mapper.CannotResolveClassException: github-plugin-configuration` indicates that `XStream` is unable to resolve the class name `github-plugin-configuration` when trying to unmarshal the XML file

### Fix:
Ensures that XStream can correctly map the `github-plugin-configuration` alias to the `GitHubPluginConfig`

### Submitter checklist
- [ ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [ ] Ensure that the pull request title represents the desired changelog entry
- [ ] Please describe what you did

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/github/config/GitHubPluginConfig.java
+++ b/src/main/java/org/jenkinsci/plugins/github/config/GitHubPluginConfig.java
@@ -85,6 +85,7 @@ public class GitHubPluginConfig extends GlobalConfiguration {
 
     public GitHubPluginConfig() {
         getConfigFile().getXStream().alias("github-server-config", GitHubServerConfig.class);
+        getConfigFile().getXStream().alias(GITHUB_PLUGIN_CONFIGURATION_ID, GitHubPluginConfig.class);
         load();
     }
 


rsandell:
This looks like it will break old configurations with the old alias.

src/main/java/org/jenkinsci/plugins/github/config/GitHubPluginConfig.java
```diff
@@ -85,6 +85,7 @@ public class GitHubPluginConfig extends GlobalConfiguration {
 
     public GitHubPluginConfig() {
         getConfigFile().getXStream().alias("github-server-config", GitHubServerConfig.class);
+        getConfigFile().getXStream().alias(GITHUB_PLUGIN_CONFIGURATION_ID, GitHubPluginConfig.class);
```
rsandell, line 88: Though the actual problem is that `load()` is called in the constructor it looks like a too involved fix to make that right. Instead all aliases already have a central point where they should be made.
```suggestion
        GitHubPlugin.addXStreamAliases();
```
And add `alias("github-server-config", GitHubServerConfig.class)` to `org.jenkinsci.plugins.github.migration.Migrator#enableAliases` instead.

rsandell:
So the actual root cause of the problem we have seen internally is that there is a mechanism that calls `Jenkins.getRootUrl()` very early in the Jenkins startup sequence. Even before `InitMilestone.SYSTEM_CONFIG_LOADED`. And that is unraveling to load all extensions including `GitHubPluginConfig` for cing the configuration to be loaded too early. So `org.jenkinsci.plugins.github.GitHubPlugin#addXStreamAliases` won't have been called yet before the config is trying to be loaded.
But fixing that might have other consequences that we can't see now.

vwagh-dev:
Found the actual root cause & fixed it there only.

azakharchenko-msol:
@vwagh-dev @rsandell Trying to fix the issue on jenkins server. Is there any suggested workaround to use?

rsandell:
Well in our case we made sure that the offender didn't call getRootUrl until after InitMileStone.SYSTEM_CONFIG_LOADED. For us it was doing it in Plugin.postInitialized , so we just changed that method to an @Initializer method instead.

azakharchenko-msol:
Thank you, see the change https://github.com/jenkinsci/github-plugin/commit/f5f220f271f42bf69f34f61fedd9196f23d6a85b
It seems it does not always  work as expected, because I still get an error like in PR title

PR #793, opened 2026-06-01
Replace rest-assured with Java HttpClient to fix Java 25 PCT failures

### LINKED ISSUE #783: Support Java 25
Currently, this plugin doesn't support Java 25, causing a failure in the PCT, making this plugin untestable on Java 25.
When supporting Java 25, https://github.com/jenkinsci/github-plugin/pull/782#discussion_r2657851536 should be taken into account.

Given Jenkins supports Java 25, this is a considerable blocker.

nevingeorgesunny:
## What are we fixing?

The github-plugin PCT tests fail when run against a Java 25 JVM. Fixes #783

`rest-assured:5.3.2` transitively pulls in `org.apache.groovy:groovy:4.0.11`. Groovy 4.x introduced `asmResolving=true` as the default class resolution strategy, which calls `AsmDecompiler` before the classloader. When the test runs on Java 25, `GroovyClassLoader` compiles `WorkflowScript` at major version 69 at runtime. ASM 9.5 (bundled in `groovy:4.0.11`) only supports up to version 64 — version 69 exceeds that ceiling, causing:

```
IllegalArgumentException: Unsupported class file major version 69
```

See [CloudBees-internal reference](https://cloudbees.atlassian.net/browse/BEE-71990) for full root cause analysis.

## RCA
1. The github-plugin is compiled with --release 17, producing class files at major version 61
2. rest-assured:5.3.2 (test-only dependency) transitively pulls in groovy:4.0.11, which bundles ASM 9.5 (supports class files up to Java 21)
3. When running PCT on Java 25, GitHubPushTriggerTest#shouldStartWorkflowByTrigger calls [jRule.assertBuildStatusSuccess(job.scheduleBuild2(0))](https://github.com/jenkinsci/github-plugin/blob/8e3b0160fdaee58c6d09fe6743567bae88819aa0/src/test/java/com/cloudbees/jenkins/GitHubPushTriggerTest.java#L67)
4. AFAICT, This triggers CpsGroovyShell.reparse() → GroovyClassLoader.parseClass("node { git '...' }") — compiled at runtime on the Java 25 JVM — producing WorkflowScript.class at major version 69
5. Groovy 4.0.11 introduced asmResolving=true as the default resolution strategy, meaning it calls AsmDecompiler before the classloader: [ClassNodeResolver.java#L193](https://github.com/apache/groovy/blob/698c68e35c8825d96360075b4e5f6f92ca0e0338/src/main/java/org/codehaus/groovy/control/ClassNodeResolver.java#L193)
6. ASM 9.5 ceiling is version 64. Version 69 exceeds it → IllegalArgumentException: Unsupported class file major version 69

**Why other plugins are not affected?**
Other plugins don't have rest-assured as a test dependency, so groovy:4.0.11 is never on their test classpath. They have  groovy-all:2.4.21 from Jenkins core , which uses a classloader-first resolution strategy and never calls AsmDecompiler in this path: [ClassNodeResolver.java#L178](https://github.com/apache/groovy/blob/41b990d0a20e442f29247f0e04cbed900f3dcad4/src/main/org/codehaus/groovy/control/ClassNodeResolver.java#L178)

8. ## Changes

1. Replace `rest-assured` with Java's built-in `java.net.http.HttpClient` in `GitHubWebHookFullTest`. `rest-assured` was only used in that one test class for plain HTTP request/response assertions — no Groovy-dependent features (xml-path/json-path) were used. Removing it eliminates the transitive `groovy:4.0.11` dependency entirely.

2. Test with Java 25 and Java 21, Java 25 released September 16, 2025. The Jenkins project wants to support Java 25 soon. Compile and test on ci.jenkins.io with Java 25 and Java 21.
Intentionally continues to generate Java 17 byte code as configured by the plugin parent pom.
Does not compile or test with Java 17 any longer because we have found no issues in the past that were specific to the Java 17 compiler. The plan is to drop support for Java 17 in the not too distant future so that the Jenkins project is only supporting two major Java versions at a time, Java 21 and Java 25.

## Test
- Ran PCT locally tests in java 21 and java 25
-   Since I don't have write access to this repo, the updated Jenkinsfile (Java 25/21) won't take effect in CI until merged. If someone with write access could trigger a build replay with the updated Jenkinsfile, that would confirm Java 25 compatibility in CI as well.

- - -
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/github-plugin/793)

### DIFF
--- a/src/test/java/com/cloudbees/jenkins/GitHubWebHookFullTest.java
+++ b/src/test/java/com/cloudbees/jenkins/GitHubWebHookFullTest.java
@@ -61,7 +61,7 @@ public class GitHubWebHookFullTest {
 
     @AfterEach
     void after() throws Exception {
-        if (httpClient instanceof AutoCloseable closeable) {
+        if ((Object) httpClient instanceof AutoCloseable closeable) { // TODO: replace with httpClient.close() once jenkins.baseline is Java 21+
             closeable.close();
         }
     }
--- a/src/test/java/com/cloudbees/jenkins/GitHubWebHookFullTest.java
+++ b/src/test/java/com/cloudbees/jenkins/GitHubWebHookFullTest.java
@@ -52,18 +52,18 @@ public class GitHubWebHookFullTest {
     private JenkinsRule jenkins;
     private HttpClient httpClient;
 
-    @AfterEach
-    void after() throws Exception {
-        if (httpClient instanceof AutoCloseable closeable) {
-            closeable.close();
-        }
-    }
-
     @BeforeEach
     void before(JenkinsRule rule) throws Throwable {
         jenkins = rule;
         jenkins.getInstance().getInjector().injectMembers(this);
         httpClient = HttpClient.newHttpClient();
+    }
+
+    @AfterEach
+    void after() throws Exception {
+        if (httpClient instanceof AutoCloseable closeable) {
+            closeable.close();
+        }
     }
 
     @Test
--- a/src/test/java/com/cloudbees/jenkins/GitHubWebHookFullTest.java
+++ b/src/test/java/com/cloudbees/jenkins/GitHubWebHookFullTest.java
@@ -6,6 +6,7 @@ import org.apache.commons.io.IOUtils;
 import org.jenkinsci.plugins.github.config.GitHubPluginConfig;
 import org.jenkinsci.plugins.github.webhook.GHEventHeader;
 import org.jenkinsci.plugins.github.webhook.GHEventPayload;
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.jvnet.hudson.test.JenkinsRule;
@@ -50,6 +51,13 @@ public class GitHubWebHookFullTest {
 
     private JenkinsRule jenkins;
     private HttpClient httpClient;
+
+    @AfterEach
+    void after() throws Exception {
+        if (httpClient instanceof AutoCloseable closeable) {
+            closeable.close();
+        }
+    }
 
     @BeforeEach
     void before(JenkinsRule rule) throws Throwable {
--- a/src/test/java/com/cloudbees/jenkins/GitHubWebHookFullTest.java
+++ b/src/test/java/com/cloudbees/jenkins/GitHubWebHookFullTest.java
@@ -15,9 +15,11 @@ import org.kohsuke.github.GHEvent;
 import java.io.File;
 import java.io.IOException;
 import java.net.URI;
+import java.net.URLEncoder;
 import java.net.http.HttpClient;
 import java.net.http.HttpRequest;
 import java.net.http.HttpResponse;
+import java.nio.charset.StandardCharsets;
 
 import static jakarta.servlet.http.HttpServletResponse.SC_BAD_REQUEST;
 import static jakarta.servlet.http.HttpServletResponse.SC_METHOD_NOT_ALLOWED;
@@ -91,7 +93,7 @@ public class GitHubWebHookFullTest {
 
     @Test
     void shouldParseFormWebHookOrServiceHookFromGH() throws Exception {
-        String encoded = "payload=" + java.net.URLEncoder.encode(classpath("payloads/push.json"), "UTF-8");
+        String encoded = "payload=" + URLEncoder.encode(classpath("payloads/push.json"), StandardCharsets.UTF_8);
         HttpResponse<String> response = httpClient.send(
                 HttpRequest.newBuilder(URI.create(getPath()))
                         .POST(HttpRequest.BodyPublishers.ofString(encoded))
--- a/src/test/java/com/cloudbees/jenkins/GitHubWebHookFullTest.java
+++ b/src/test/java/com/cloudbees/jenkins/GitHubWebHookFullTest.java
@@ -16,6 +16,7 @@ import org.kohsuke.github.GHEvent;
 import java.io.File;
 import java.io.IOException;
 import java.net.URI;
+import java.util.Locale;
 import java.net.URLEncoder;
 import java.net.http.HttpClient;
 import java.net.http.HttpRequest;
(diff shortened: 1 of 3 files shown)


src/test/java/com/cloudbees/jenkins/GitHubWebHookFullTest.java
```diff
@@ -18,158 +14,155 @@
...
         jenkins.getInstance().getInjector().injectMembers(this);
-
-        spec = new RequestSpecBuilder()
-                .setConfig(newConfig()
-                        .encoderConfig(encoderConfig()
-                                .defaultContentCharset(Charsets.UTF_8.name())
-                                // GitHub doesn't add charsets, so don't test with them
-                                .appendDefaultContentCharsetToContentTypeIfUndefined(false)))
-                .build();
+        httpClient = HttpClient.newHttpClient();
```
jtnord, line 61: Should this be `closed` in an `@AfterEach`  when on Java21? 

e.g. (assuming the compiler and Spotbugs are not smart enough to flag the following as dead code)

```java
@AfterEach
void after() throws Throwable {
  if(httpClient  instanceof AutoCloseable closable) {
    closable.close();
  }
}
```

jglick:
could be added to https://github.com/orgs/jenkinsci/projects/41 I guess (not sure if there is some more appropriate tracking board).

jglick:
Also should be marked as fixing https://github.com/jenkinsci/github-plugin/issues/783, which I found via https://github.com/jenkinsci/bom/pull/6142.

src/test/java/com/cloudbees/jenkins/GitHubWebHookFullTest.java
jglick: In some cases while waiting for `jenkins.baseline` to be new enough I have had to use the idiom

```java
if ((Object) httpClient instanceof AutoCloseable closeable) { // TODO Java 21+
```

src/test/java/com/cloudbees/jenkins/GitHubWebHookFullTest.java
```diff
@@ -18,158 +14,155 @@
...
+import static org.jenkinsci.plugins.github.webhook.RequirePostWithGHHookPayload.Processor.SIGNATURE_HEADER;
+import static org.jenkinsci.plugins.github.webhook.RequirePostWithGHHookPayload.Processor.SIGNATURE_HEADER_SHA256;
 
 /**
  * @author lanwen (Merkushev Kirill)
  */
 @WithJenkins
 public class GitHubWebHookFullTest {
 
-    // GitHub doesn't send the charset per docs, so re-use the exact content-type from the handler
```
jglick, line 42: Is there a reason this comment is being deleted? Is it somehow related to the changes here?

src/test/java/com/cloudbees/jenkins/GitHubWebHookFullTest.java
```diff
@@ -18,158 +14,155 @@
...
-                .log().all()
-                .expect().log().all()
-                .statusCode(SC_OK)
-                .request()
-                .post(getPath());
+        HttpResponse<String> response = httpClient.send(
+                HttpRequest.newBuilder(URI.create(getPath()))
+                        .POST(HttpRequest.BodyPublishers.ofString(classpath("payloads/ping.json")))
+                        .header("Content-Type", APPLICATION_JSON)
+                        .header(GHEventHeader.PayloadHandler.EVENT_HEADER, GHEvent.PING.name().toLowerCase())
```
jglick: Not new in this PR, but while you are here:
```suggestion
                        .header(GHEventHeader.PayloadHandler.EVENT_HEADER, GHEvent.PING.name().toLowerCase(Locale.ROOT))
```
(and similarly elsewhere) lest a Turkish / Azeri developer get a test failure due to a header value being `pıng`.

src/test/java/com/cloudbees/jenkins/GitHubWebHookFullTest.java
```diff
@@ -18,158 +15,162 @@
...
 
-    public Header eventHeader(GHEvent event) {
-        return eventHeader(event.name().toLowerCase());
-    }
-
-    public Header eventHeader(String event) {
-        return new Header(GHEventHeader.PayloadHandler.EVENT_HEADER, event);
+    private String getPath() {
+        return jenkins.getInstance().getRootUrl() + GitHubWebHook.URLNAME.concat("/");
     }
```
jglick, line 176: Gratuitous diff: moved from previous location.

jglick (approved):

src/test/java/com/cloudbees/jenkins/GitHubWebHookFullTest.java
nevingeorgesunny: tried with 
```
    @AfterEach
    void after() throws Exception {
        if ((Object) httpClient instanceof AutoCloseable closeable) {
            closeable.close();
        }
    }

```
tested locally

src/test/java/com/cloudbees/jenkins/GitHubWebHookFullTest.java
nevingeorgesunny: Sorry , got removed by accident , while I was re writing the test , added it back. Thank you

src/test/java/com/cloudbees/jenkins/GitHubWebHookFullTest.java
nevingeorgesunny: done add Locale.ROOT to all the `.toLowerCase()`

jglick (approved):

nevingeorgesunny:
@KostyaSha can you please review/merge this PR

jtnord (approved):

NotMyFault (approved):
Thanks a bunch!

nevingeorgesunny:
Thanks you all for the approvals 
@NotMyFault / @KostyaSha the Pr is approved  can you please merge and release.

nevingeorgesunny:
@integer @lanwen @oleg-nenashev, can you please merge and release -- this PR has been approved

jglick:
I see https://github.com/jenkinsci/github-plugin/releases/tag/v1.47.0, thanks!

